PR #26, opened 2015-01-09
Improved saucelabs integration

johanneswuerbach:
Hey I just extracted some saucelabs browser testing improvements I made for htmlbars into an ember-cli addon: ember-cli-sauce

This has some advantages over your current implementation:
Use only a single tunnel (faster)
Test results available on SauceLabs
And we can collect further improvements in a central place: https://github.com/johanneswuerbach/ember-cli-sauce

What do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,11 +20,11 @@ before_install:
 
 install:
   - npm install -g bower
-  - npm install -g saucie
   - npm install
   - bower install
 
 before_script:
+  - curl -L https://gist.githubusercontent.com/johanneswuerbach/e4950820c0d685fc545e/raw/sauce_connect_setup.sh | bash # Create a sauce tunnel
   - ember s --live-reload=false & # Start a server so we can hit the fake API from integration tests
   - sleep 10 # wait for the server to be started
 
--- a/testemci.json
+++ b/testemci.json
@@ -3,31 +3,31 @@
   "test_page": "tests/index.html",
   "launchers": {
     "SauceLabs_Chrome": {
-      "command": "saucie --platformSL='Windows 7' --host='http://localhost' --port=8080",
+      "command": "./node_modules/.bin/ember-cli-sauce --platformSL='Windows 7' --no-ct -u <url>",
       "protocol": "tap"
     },
     "SauceLabs_Firefox": {
-      "command": "saucie --platformSL='Windows 7' --browserNameSL='firefox' --host='http://localhost' --port=8080",
+      "command": "./node_modules/.bin/ember-cli-sauce --platformSL='Windows 7' --browserNameSL='firefox' --no-ct -u <url>",
       "protocol": "tap"
     },
     "SauceLabs_IE_9": {
-      "command": "saucie --browserNameSL='internet explorer' --versionSL='9' --platformSL='Windows 7' --host='http://localhost' --port=8080",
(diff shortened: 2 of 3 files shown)


johanneswuerbach:
Looks like the new sauce connect is not proxying port 4200 https://docs.saucelabs.com/reference/sauce-connect/#can-i-access-applications-on-localhost. I'll fix that

jrjohnson:
Great idea to consolidate this into an addon.

.travis.yml
```diff
@@ -20,11 +20,11 @@ before_install:
 
 install:
   - npm install -g bower
-  - npm install -g saucie
   - npm install
   - bower install
 
 before_script:
+  - curl -L https://gist.githubusercontent.com/johanneswuerbach/ce15304c6d33538ecc6e/raw/sauce-connect.sh | bash # Create a sauce tunnel
```
jrjohnson: Can we use the travis addon for opining the tunnel?
http://docs.travis-ci.com/user/sauce-connect/

```yml
addons:
  sauce_connect: true
```
That saves a download trip.

testemci.json
```diff
@@ -3,31 +3,31 @@
   "test_page": "tests/index.html",
   "launchers": {
     "SauceLabs_Chrome": {
-      "command": "saucie --platformSL='Windows 7' --host='http://localhost' --port=8080",
+      "command": "./node_modules/.bin/ember-cli-sauce --platformSL='Windows 7' --no-ct -u <url>",
```
jrjohnson: What is the <url> token here?

jrjohnson:
Sauce support told me last month that localhost was now open on every port except for the andriod emulator and a bug in safari.  Apparently that list is no longer valid.

.travis.yml
johanneswuerbach: Sadly the addon doesn't seem to be installed when PRs are tested.

testemci.json
johanneswuerbach: This token gets replaced by testem with "http://localhost:8080/someId", so you don't need to configure the host & port manually.

.travis.yml
jrjohnson: Thats right, I remember that now.  I was thinking I had ditched it because saucie was opening all its own tunnels.

johanneswuerbach:
Looks like the new sauceconnect screws up the port mapping :cry: :
https://www.dropbox.com/s/1h36h549d53mf7t/Screenshot%202015-01-09%2023.48.21.png?dl=0

Both ports are mapped to the test runner and 4200 does not map to the api... o.O

johanneswuerbach:
I submitted a request to the sauce support. This is definitely broken:
https://saucelabs.com/tests/8cad76d985324bdb9f895da8d969ce9f

```js
var http = require('http');
var fs = require('fs');

http.createServer(function (req, res) {
  res.writeHead(200, {'Content-Type': 'text/plain'});
  res.end('Port 4200');
}).listen(4200);

http.createServer(function (req, res) {
  res.writeHead(200, {'Content-Type': 'text/plain'});
  res.end('Port 7357');
}).listen(7357);

```

johanneswuerbach:
@jrjohnson this is finished now. Some browsers are failing now on travis, but also on my local env. Does Firefox pass the suite for you?

Before the test result was published before all tests finished in some cases, which might have made all browsers look like passing.

jrjohnson:
@johanneswuerbach can you try rebasing this PR and see if the tests now pass?

johanneswuerbach:
@jrjohnson thanks, this passed now.

jrjohnson:
Awesome.  Thanks for this really nice improvement.

PR #31, opened 2015-02-04
Course List Page

jenzweben:
Changed section id =courses to section class=courses on courses.hbs

created courses class on _global.scss


jrjohnson:
Couple of things.
* please leave the id=courses, we need that to pull the right content for testing.
* Can we make it section class='main' so we can re-use that css on the learnerGroup page and not need to duplicate it.
* You left in the collapsed=false code - that shouldn't get merged.

PR #34, opened 2015-02-05
Feature/coursedetail

jenzweben:
added right, left & bottom padding of .8em after Jason made his killer
div structural changes


jrjohnson:
already merged as #33

PR #63, opened 2015-02-09
Basic learner group listing

jrjohnson:
This is the basic content for a learner group list.  Should be functionally complete, needs some styling improvement.


saschaben:
Basic list view seems to be working, still some wonkiness with the back buttons. Assuming this does NOT include individual group displays, just the main list page.

jrjohnson:
Correct. Just the list. Not the groups.

PR #68, opened 2015-02-10
Add info to the course detail session list.

### LINKED ISSUE #53: Fix content on course detail session list
The list rows do not currently display most of the required count content.

jrjohnson:
fixes #53


saschaben:
Appears to be intermittently failing to load session type -- about 40-50% of sessions. no rhyme nor reason to it from the UI side.

jrjohnson:
@saschaben check this now, problems with session type should be resolved.  
@jenzweben - I changed the first offering from the mock to include the time and not just the date since I figured thats what people would want to know.  Is that correct?  Or should it just be mm/dd/yyyy as in the mock?

jenzweben:
Where should I be viewing the changes you made?

saschaben:
@jrjohnson @jenzweben I think the timestamp addition to the date will be welcome by the users. It makes it a little more busy / messy than the mock, but I think the value will outweigh the issue

jenzweben:
OK

Issue #46, opened 2015-02-09
New action dropdown


jrjohnson:
@jenzweben can you take a look at these two options and let me know which you woudl prefer:
1. http://rsschermer.github.io/ember-rl-dropdown/
2. http://dropkickjs.com/

Number is slightly easier to implement, so unless you want the search and group functionality in the dropkick one lets go with that.

jenzweben:
Definitely #1

jrjohnson:
@jenzweben please take a look at improving style on the new dropdown action menu.  I think we can use the same basic structure for other drop down things like the publish button.  There is for sure a z-index issue when the menu is at the bottom of a list.
.

PR #73, opened 2015-02-10
Use an addon for the action menu

jrjohnson:
Replaces action menu across the board with one that actually works.  Once this is merged #46 should be assigned to @jenzweben to take a look at style specifics.

### DIFF
--- a/app/styles/components/_action-menu.scss
+++ b/app/styles/components/_action-menu.scss
@@ -1,40 +1,25 @@
-$dropdown-color: $base-font-color;
-$dropdown-description-color: lighten($dropdown-color, 40);
-$dropdown-border: 1px solid silver;
 $dropdown-background: $base-background-color;
 $dropdown-background-hover: darken($dropdown-background, 3);
-$dropdown-inner-border: 1px solid silver;
-$dropdown-height: 26px;
 $dropdown-padding: 1em;
-$dropdown-distance-from-menu: 50px;
-$dropdown-arrow-top-distance: 0;
+$dropdown-distance-from-menu: $dropdown-padding * 2;
 
 .action-menu {
-  .container {
-    display: inline-block;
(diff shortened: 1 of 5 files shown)


saschaben:
seems to be just fine in groups, but is acting oddly in sessions (perhaps due to only one selection option being available?)

saschaben:
Hmm. this widget definitely wonky as implemented in course list course and session list. NRFPT.

jrjohnson:
Can you specify wonky.  Works for me, so I need some more specifics.

saschaben:
click the widget, make selection from dropdown: lable from dropdown prints to the widget form, and no action occurs. I'd give you a screenshot, but the branch is now giving me a handlebars error and won't render the page...

saschaben:
Uncaught Error: <ilios@component:action-menu::ember4504> Handlebars error: Could not find property 'rl-dropdown-container' on object <ilios@component:action-menu::ember4504>.

jrjohnson:
Thanks.  Use nomless to fix that error.  I will look into the click action.  Not sure if I tested that there.

saschaben:
ilios 2015-02-10 13-01-11
ilios 2015-02-10 13-01-59

jrjohnson:
ok @saschaben give this another look.

saschaben:
much better for course list, group list, and course overview. still has a z-index issue where the dropdown gets hidden when at the bottom of a content area:

ilios 2015-02-10 15-08-32

...and still is wonky per before on sessions list. FYI this is acceptable for demo, but I don't want any code inconsistencies to slip thru the cracks.

jrjohnson:
Session list is broken because it doesn't actually do anything.  I'm going to pass off the z-index problem to @jenzweben for after this is merged and she takes a styling pass.

PR #80, opened 2015-02-11
added label for schoolsfilter

thecoolestguy:
label.class needs to be set to 'display:inline'


jrjohnson:
@jenzweben is going to re-mock the filter interface so I'm going to close this for now, un-merged.

Issue #72, opened 2015-02-10
Create simpler, shorter loading animation


jenzweben:

jenzweben:
duplicate of #245

PR #91, opened 2015-02-11
Animation Improvements

jrjohnson:
I pulled out the delay at the start of the wave which might remove the perceived loading issue.

I also switched to a cross fade animation, see if that looks better.

### DIFF
--- a/app/styles/components/_waveloader.scss
+++ b/app/styles/components/_waveloader.scss
@@ -28,10 +28,10 @@
           @include animation(wave 2.0s infinite ease-in-out);
 
       }
-      @for $i from 1 through 8 {
-        .widget span:nth-child(#{$i}){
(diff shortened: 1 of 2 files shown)


jrjohnson:
The problem is that while the loading animation is playing the browser is EXTREMELY busy parsing the course data coming in.  We can add a bunch of arbitrary delays in here, but in the end we will just be making the user wait to see their content.  Maybe @jenzweben can convert the css animation to an animated gif and we can put that instead?  That wouldn't require the browser to handle the animation sequence.

saschaben:
let's merge as is -- the transitions are fine for tomorrow -- but let's review with @jenzweben per #72.

jenzweben:
updated loading animation http://monsoon-design.com/UCSF/pulse8.html

Issue #64, opened 2015-02-10
Courses List - Add New Button - move to filter-tools


jenzweben:
Please move the Add New button to either the filter-tools div or above the table so that the table can utilize the full width detail-view

saschaben:
This should probably be standardized across all page templates.

jenzweben:
I've mocked it up for Learner Group Landing - working on Courses List mock now

jenzweben:
Here's the new Courses List mock - it shows the scenario where a user has selected All Courses, School of Med, School of Nursing & School of Pharm for 2015

![course-list](https://cloud.githubusercontent.com/assets/5023894/6139859/a870a8c0-b145-11e4-9819-0d0c03df4a7a.png)

jenzweben:
Here is the mock showing the scenario where a user has selected My Courses and only has access to one school - School of Medicine

![course-list-oneschool](https://cloud.githubusercontent.com/assets/5023894/6139960/d0dba264-b146-11e4-8a54-df49bbb6079c.png)

saschaben:
comments/thoughts:  
1) If mulitple schools are selected in this case, should we show the dropdown with a single school ("medecine") rather than doing something else -- but, uselessly I have no suggestion at the moment....
2) Also, let's keep the color of the publish state labels in sync with course pages: not published = amber, scheduled = brown, published = green.

Otherwise, perfect.

jenzweben:
Scenario:  All Courses, 3 Schools Selected (Medicine, Nursing & Pharma)
![course-list](https://cloud.githubusercontent.com/assets/5023894/6140068/36ff161a-b148-11e4-8d50-d00c7482cf54.png)

Scenario:  My Courses - only have access to 1 School (Med)
![course-list-oneschool](https://cloud.githubusercontent.com/assets/5023894/6140075/492d9df2-b148-11e4-9fd9-62e2308e491c.png)

jrjohnson:
Can we go ahead and mock the school selector as a set of checkboxes?  Thats how we're going to need to code it.  If we come up with a better widget we can just replace.  We might even want to go with monkey buttons for lists of less than three (most common) and checkboxes if the count is above that.

saschaben:
[...]

PR #118, opened 2015-02-12
revamped grouplist structure in courses and learnergroups, restyled grouplists

thecoolestguy:
revamped grouplist structure in courses and learnergroups, restyled grouplists, refs #64


jrjohnson:
In general I think @thecoolestguy is correct, this entire component should be called resultlist.  or results-list.  Specific comments in line messages.

app/styles/components/_grouplist.scss
```diff
@@ -1,21 +1,25 @@
...
-    font-size: $base-font-size;
-    text-align: left;
-    background-color: $background-grey;
-    padding: $base-font-size;
+    .grouplist-header {
+      background-color: $header-grey; 
+      overflow: auto;
+      padding: .2em;
+      .grouplist-title {
+        float: left;
```
jrjohnson: Instead of floating these left and right they should be placed on the grid.  See https://github.com/ilios/frontend/blob/master/app/styles/components/_detailview.scss#L46 for an example.  Basically the grouplist already spans 12 of 12 so the title should maybe be 3 of 12 and the h2 9 of 12.

app/templates/learnergroups.hbs
```diff
@@ -34,15 +34,20 @@
       {{input type='text' value=titleFilter placeholder=placeholderValue}}
     </div>
   </div>
-  <section class='grouplist'>
-    <h2>{{title}}</h2>
+
+  <section class='grouplist clearfix'>
```
jrjohnson: Instead of using the section class clearfix use the sass @include clearfix; on the element.

saschaben:
looks good. I think that the word "results" is probably not what we want in the final iteration, but will do for now as a placeholder. we should get user feedback on nomenclature for this one.

Merging.

saschaben:
or rather --- merge as soon as ready!

thecoolestguy:
I renamed the grouplist div's classes to 'resultslist' and removed the floats from the div's in favor of using Neat, but I'm still trying to get a hang of using Neat, so it may not be correct.

Also, I noticed this morning that there was a border around the filter tools that I'd never seen before, so I removed that.

jrjohnson:
Code changes look good.  Nice work.  Needs a rebase to merge.

PR #124, opened 2015-02-13
Limited persistence for development fixtures

jrjohnson:
In order to make the development API more stable I've added some change management functions.  Basically once you run ember serve everything you do in the local api will get stored.  When you kill the server and restart it those changes will go away.  This should make for a more stable development experience.

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -44,7 +44,7 @@ module.exports = function(environment) {
     ENV.adapterHost = 'http://ilios.dev';
     ENV.adapterNamespace = 'app_dev.php/api/v1';
     ENV.contentSecurityPolicy['script-src'] += " 'unsafe-eval'";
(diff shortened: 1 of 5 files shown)


jrjohnson:
@thecoolestguy can you take a look at this and make sure I didn't mess anything up?  Other than an eyeball review if @saschaben can check it out and click around to make sure nothing broke that would be great.  It should be a transparent change to the UI, except for the fact that if you reload the window your changes will remain.

saschaben:
NRFPT: edits to course name cause course sessions to fail with exception error.

saschaben:
the sessions link issue is fixed, but note: course start/end dates and level changes  are _not_ persistent even within a session. A page refresh or navigation away wipes them back to starting place.

`course title`, `external ID` and `directors`: all persistent as expected. Merging and will review for issues after merge.

Issue #76, opened 2015-02-10
Learner Groups page - Add New btn location and need to add "School:" before "Medecine"


jenzweben:
Add "Add New" button to Learner Groups page
I have it to the right of the table in the mockup but it might be better if we move it to the header so that the table can be full width...

![image](https://cloud.githubusercontent.com/assets/5023894/6138299/bdf9b274-b134-11e4-9675-c726e44b129e.png)

jrjohnson:
@jenzweben please do a new mockup for this first so @saschaben can make a choice.

jenzweben:
![learner-groups-landing-v7](https://cloud.githubusercontent.com/assets/5023894/6138632/a10a4c8e-b137-11e4-9b3f-735a5f1c1784.png)

thecoolestguy:
fixed in merge #127

Issue #51, opened 2015-02-09
Need to add class called "schoolsfilter" to the div #schoolsfilter state where a user only has access to one school so that I can style that element


jenzweben:

jenzweben:
learnergroups - I will also be creating a mock in like 5 mins

jenzweben:
![learner-groups-landing-v7](https://cloud.githubusercontent.com/assets/5023894/6138640/a9f1525c-b137-11e4-91fe-5da6b173570c.png)

thecoolestguy:
So, the current 'schoolsfilter' element currently reads "<div id="schoolsfilter" class="filter">Medicine</div>" (for a single-school or multi-school user)

Are you saying that you want it to remain the same for multi-school users but, for single school users it should read: "<div id="schoolsfilter" class="filter schoolsfilter">Medicine</div>"?

jenzweben:
So this case is for when a user only has access to a single school.  If they have multiple schools, they'll see a dropdown.  

Somewhere, somehow we need to add a label - "School:" and conditionally show the name of it (Medecine, Nursing, Pharmacy etc.)

jenzweben:
The end result would be School: Medecine or School:  Pharmacy

thecoolestguy:
fixed in #127

PR #127, opened 2015-02-14
added 'school: ' labelling and inline styling for filter data, added two inline classes

thecoolestguy:
added 'school: ' labelling and inline styling for filter data, added inline-label and inline-data classes to global stylesheet, refs #76, #51

### DIFF
--- a/app/templates/learnergroups.hbs
+++ b/app/templates/learnergroups.hbs
@@ -46,11 +46,11 @@
       <div class="resultslist-title">
         <h2>{{title}}</h2>
       </div>
-    {{!-- {{#if currentCohort}} --}}
+    {{#if selectedCohort}}
         <div class='resultslist-actions'>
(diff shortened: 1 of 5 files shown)


app/templates/learnergroups.hbs
```diff
@@ -40,11 +46,11 @@
       <div class="resultslist-title">
         <h2>{{title}}</h2>
       </div>
-      {{#if currenCohort}}
+    {{!-- {{#if currentCohort}} --}}
```
thecoolestguy: I commented-out the 'currentCohort' check, so the button would appear for checking... Need to discuss need for this check with @jrjohnson

jrjohnson: It is possible that there are NO cohorts in a program.  Or no programs in a school.  Or no schools in the system.  So while if there is at least one cohort we do force the selection of it, we still have to be wary of situations where there isn't.

PR #128, opened 2015-02-14
Fixed CSS for publication status

### LINKED ISSUE #119: "Scheduled" items showing "Publish" label
Color is right, label is wrong. fixit, Jeeves!
ilios 2015-02-12 11-04-53
ilios 2015-02-12 11-05-30

jrjohnson:
The scheduled class was being applied to any item with a tbd status, even if it had also been published.

Fixes #119


saschaben:
OK, but your comment does not make sense to me: if an item has TBD status, it is TBD. the end. If it has "also" been published, what does that mean?

the workflow states available are:

1: Not published: has not been pushed past any state. equivalent to current "draft". available to educators authorized to view, not available to students.
2: Published: has been approved for view by students, regardless of internal content. Visible to students and educators alike.
3: Scheduled: has been partially "published": fully available to educators, but students only see title and time block on calendar, without content ("TBD").

Currently, none of the courses in the list show as scheduled. uncertain if this is correct or not after reading your comment. Please clarify.

jrjohnson:
Course 632 Life Cycle / Prepilogue 2014-15 in the database has publishedAsTbd = true, but it also has publish_event_id=60439.  For that course I would assume that the canonical status would be published.  Is that incorrect?  

There weren't actually any scheduled courses in our data set so I just pushed one it is  347 Transitional Clerkship 2013-14 in SOM.

saschaben:
I believe that scheduled events receive publish_event_id values, since TBD is a version of published - effectively published with a special flag.

jrjohnson:
AH!  Ok - then I have to redo this.  I assume those were just DB artifacts of bad past updates.  I will look at this whole area again and see what is on stage.

saschaben:
Okey dokey.

jrjohnson:
Give it another look.

saschaben:
looks good!

PR #129, opened 2015-02-14
Working course objective manager

### LINKED ISSUE #57: Add course objective inspector
Allow user to link course objective to parent objective.  Change the parent objective title on the course detail page to be a link to this route.

jrjohnson:
Allows a user to associate a course object with a single parent objective or remove that association.

Fixes #57

### DIFF
--- a/app/components/course-header.js
+++ b/app/components/course-header.js
@@ -2,4 +2,5 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
   course: null
+  editable: true,
 });
--- a/app/components/course-header.js
+++ b/app/components/course-header.js
@@ -1,6 +1,6 @@
 import Ember from 'ember';
 
 export default Ember.Component.extend({
-  course: null
+  course: null,
   editable: true,
 });
--- a/app/components/detail-competencies.js
+++ b/app/components/detail-competencies.js
@@ -5,24 +5,26 @@ export default Ember.Component.extend({
   domains: function(){
     var domainContainer = {};
     var domainIds = [];
-    this.get('competencies').forEach(function(competency){
-      var domain = competency.get('domain');
-      if(!domainContainer.hasOwnProperty(domain.get('id'))){
-        domainIds.pushObject(domain.get('id'));
-        domainContainer[domain.get('id')] = Ember.ObjectProxy.create({
-          content: domain,
-          subCompetencies: []
-        });
-      }
-      if(competency.get('id') !== domain.get('id')){
-        var subCompetencies = domainContainer[domain.get('id')].get('subCompetencies');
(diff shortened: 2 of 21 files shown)


saschaben:
Labels:

- "Objective" should be labeled "Course Objective"
- Listing should be labeled "Select Parent Objective"
- If there is only one cohort associated to course, the cohort should be displayed as a read only label, not a 1-item drop down. We may need a label for this as well when there are multiple cohorts: "Select Parent for..."

- page currently just jumps back to top on select/deselect of item: on selection / deselection the page should flip back to the course objective listings.
- is it possible to active highlight the parent competency along with the selected objective on select?

saschaben:
Also: what is our pattern here for controlling when no cohort has yet been selected? No "add new"  buttons on course page? Or a "please select a cohort" message on the objective selector page or when the button is pushed? thoughts?

jrjohnson:
@saschaben fixed a lot of the stuff above with style and title improvements.  I think a more refined style will have to go to @jenzweben but I added classes to make that possible.

I didn't fix the on click return to course list.  I can only go so far right now as on click return to course page.  I can add functionality to make the course page expanded when you return, but actually scrolling to the objective area is more difficult.  Do you want to go ahead and return to the page anyway?  Or stay here where the results of your action are visible?

saschaben:
cool. will review shortly. return to the course expanded page without specific scrolling focus is probably the best choice. at least for now.

saschaben:
better. @jenzweben will need to review, and we will also need to review label nomenclature. currently though, throwing an error (in chrome): go to objective parent, add or reselect parent, return to course, select another parent to review/add --- select list fails to load with following error:
ilios 2015-02-15 12-32-24

saschaben:
...and page reload clears the error and loads the select list correctly.

jrjohnson:
The issues with loading are fixed in #134 
That exposed a different issues there where returning to the manager after going back to a course doesn't load the cohorts.  I'm looking into that now.

jrjohnson:
This requires #134 and #136 to work well.  Without those it works, you just have to refresh the manager after you get there.

jrjohnson:
@saschaben This is now unblocked and I merged in the changes from #134 and #136.  It can now be tested and should be 100% working (or, you know, not as the case may be)

saschaben:
looking good (but sure needs @jenzweben!). A couple things: there is a 3 second lag between clicking a parent selection and the page flipping over to the loading timer / course page. Is there anything we can do in asynchronous API call-land to make this action more instant? It doesn't need to do anything more than show the loader. 
also: let's try altering the label "Select Parent For: " to simply "Selecting for:". Less redundant.

PR #130, opened 2015-02-15
Improve Live Search

### LINKED ISSUE #113: determine minimum chars for person search
right now search is set for 3 chars (plus a space). once live, need to dial this in to the best and lowest value.

jrjohnson:
Indicate when a search is running and when there are no results.
Don't debounce in the live search, do that in the controller
Set the minimum term length to 3 not including spaces.

Fixes #113


saschaben:
Hmm. seems to have regressed to a less impressive state -- now will generate random result sets on many searches:

ilios 2015-02-15 10-11-32

Also seems to have some weird mid-search lagging/freezing.

jrjohnson:
I didn't change the search at all.  Are you sure all these people don't have a middle name of %jeffre%?  I believe that the demo data is built off of a limited list of name possibilities.

I did add a delay on the api end to simulate actual request lag. But I added the searching.... and no results.... tags to display when the search had been sent but the API hadn't yet returned any results.

saschaben:
I really don't know what is going on here -- this results certainly surprised me as well, considering what I saw in your PR. I don't think the demo data includes middle names, but we should check with @thecoolestguy to confirm.

saschaben:
OK -- just checked earlier branch, this appears to be pre-existing behavior. Let's figger out what is going on before any merging. Wondering if this might be a bug in ember.

saschaben:
HAH!
Livesearch appears to be finding not only matches to the string, but to matches of any two characters within the string, regardless of whether they are consecutive.

jrjohnson:
I'm guessing what you're seeing is because we're cheating in the dev api and going a quick concat string match since we don't have an actual database with AND operators.  
Basically we take the lowercase string firstName + middleName + lastName + email so susananderdonsudemo@ilios-demo and then look for any matches in there to each of the space separated search strings.

All of those jeffre matches were on email.  There are a bunch of users in the demo data with emails like first.*jeffrey*.last @ ilios-demoschool.edu

saschaben:
ahhh. makes sense. If we are going to include email in the search, we need to display it along with name for search results. Actually, we should display it anyway, because without email we run into the issue of not being able to visually distinguish between Jon Johnson (jrjohnson@masteroftheuniverse.net) and Jon Johnson (jj@thecoolhouse.biz) for selection.

jrjohnson:
Ok, added email.  It makes the box too big so we will need a styling pass once this is merged.

saschaben:
We tried to address this issue previously using the mouseover display of email --- with LiveSearch obviously that model no longer works. Let's see what @jenzweben can do with styling.

PR #131, opened 2015-02-15
WIP: Render a pre-ember page view

### LINKED ISSUE #71: render message for older browsers
In older browsers our site doesn't render at all.  We need to do some detection and display a nice error message instead of a white screen.

@jenzweben - can you design this screen?  It shouldn't depend on any of our existing content (header, navigation, footer) and be completely self contained like the login / landing page.

jrjohnson:
This accomplished two tasks.  First it puts up a loading indicator along with a noscript tag right away.  So if the browser takes a while to render our application or if javascript is not available the user will see some info immediately.  Second after 10 seconds it displays an error message that the application may not render at all in this browser.  This way we don't have to detect browser version we can just be certain that if that message displays something didn't work.

Fixes #71

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -2,7 +2,8 @@
   "predef": [
     "document",
     "window",
-    "-Promise"
+    "-Promise",
+    "$"
   ],
   "browser": true,
   "boss": true,
--- a/app/index.html
+++ b/app/index.html
@@ -13,51 +13,9 @@
     <link rel="stylesheet" href="assets/ilios.css">
 
     {{content-for 'head-footer'}}
-    <script type='text/javascript'>
-        setTimeout(function(){
-          $('#browsererrormessage').removeClass('hidden');
-          $('#initialpageloader .waveloader').remove();
-        }, 10000);
-    </script>
   </head>
   <body>
-    <div id='initialpageloader'>
-      <header class="main">
-        <span class='logo'>
-          <img src='images/ilios-logo.png' alt='Ilios Logo' title='Ilios Logo' />
-        </span>
-      </header>
-
-      <div class='content'>
-        <div class='waveloader is-full-screen is-full-size'>
-          <div class="wrapper">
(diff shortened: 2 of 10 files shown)


jrjohnson:
This is a work in progress.  I think this same functionality can be accomplished without messing with the default template so please don't merge. 
No matter how the code is completed though - it still needs review: @jenzweben, @saschaben  I added some border and shading to the initial mock in #71 and changed the wording a bit.  Please take a look and see if I'm way off base.  You will need to disable javascript in your browser to see the error message.
It is also worth testing in several configurations to make sure the loading indicator doesn't jump around between first load, and page load.  The best way to see this is to refresh the course list.

saschaben:
I think this actually works well for the error message. I can't get javascript disabled in FF --anyone know how to do that these days? -- but in chrome and safari this is rendering fine on Mac.

Not sure I get how this will also handle older browsers, but then again the coffee hasn't taken hold yet....

saschaben:
Update: FF requires an about.config setting change for javascript.enabled from true to false. talk about hiding things....

in any case, rendering fine in FF on mac as well.

jrjohnson:
The way the error works:
1.  Page loads
2.  Timer starts to display the warning message about no support
3.  Ember initializes

On supported browsers:
4. Ember loads and kills off the timer, the errors message, and the loading indicator.  Replacing the entire window with its own content.

On un-supported browsers:
4. Ember fails to load and shows no content
5.  The timer fires, destroying the loading message, but displaying the warning about not being supported.

The way to test the older browser behavior currently involves commenting out the lines
* {{content-for 'head'}}
* {{content-for 'head-footer'}}
* {{content-for 'body'}}
* {{content-for 'body-footer'}}

That is where ember gets loaded.

jrjohnson:
OK, @saschaben I'm happy with the code now.  This is ready to merge whenever you like.

PR #134, opened 2015-02-16
Be carefull when we call set on a controller

### LINKED ISSUE #133: Rapidly clicking between complicated routes causes an error
To duplicate go to /courses and choose a course then click back button, forward button, back button.  At some point it will stop working and an error will display in the console about calling set on a destroyed object.

jrjohnson:
The route hook setupController gets called even if the user has already clicked away from the current route and it is being destroyed.  This causes errors when we try and call set on the destroyed controller object.  This change ensures that we call setupController only after the destroy loop has completed and that we check the state of the controller before attempting to set anything on it.

Fixes #133

### DIFF
--- a/app/routes/course.js
+++ b/app/routes/course.js
@@ -1,14 +1,39 @@
 import Ember from 'ember';
 
 export default Ember.Route.extend({
+  availableTopics: [],
+  program: [],
+  afterModel: function(course){
+    var self = this;
+    var deferred = Ember.RSVP.defer();
+    Ember.run.later(deferred.resolve, function() {
+      var resolve = this;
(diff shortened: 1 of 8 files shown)


saschaben:
not sure if it has bearing on this fix, or if it simply is exposing a different issue, but topics is broken in a similar fashion -- add/remove topics, and the widget fails, throwing " Uncaught TypeError: Cannot read property 'destroy' of undefined". Causing this error will (duh) override the fix here.

jrjohnson:
Its probably the same type of issue, just caused by different set of conditions.  Can you report that as a separate bug?

This does fix the problems in #129

PR #163, opened 2015-02-18
added userName and userEmail properties to live search results, separated each with respective classname

thecoolestguy:
added/using 'result.userName' and 'result.userEmail' properties instead of combined 'result.Title' value, so each can be classed for css.

CSS classes added are 'livesearch-results-name' and 'livesearch-results-email'


jrjohnson:
If we need these properties to be called out we can't do it in the reusable live search component.  It uses a generic title so it can be called from other search options like finding a course, or a program.  We will have to re-archetect this component or else create a live user search option.

PR #186, opened 2015-02-20
edited course detail title header so that it wraps less, refs #177

jenzweben:


app/styles/components/_detailview.scss
```diff
@@ -132,21 +132,20 @@
...
     .actions {
       /*vertical-align: bottom;*/
-      @include span-columns(5 of 12);
-      @include shift(2);
+      @include span-columns(4 of 12);
       text-align:right;
 
       .status {
-        font-size: $h5-font-size;
+        font-size: 14px;
```
jrjohnson: All other fonts are in EM - can this be a $base-font-size *.8?

app/styles/components/_global.scss
```diff
@@ -55,6 +55,7 @@ button {
   }
   &.publish, &.unpublish {
     background-color: $ilios-green;
+    font-size: .80em;
```
jrjohnson: All other font sizes are specified as $base-font-size * .8 so if we ever change the base they will stay relatively the same.

PR #189, opened 2015-02-20
Complete reset of sass components

jrjohnson:
Several changes here:
1. Remove bourbon bitters (base) styles since they were causing @jenzweben to hate me.  Now normalize css is the only base style.
2. Created a reset.scss which sets some defaults that bitters was handling before
3. Add some basic stuff to global (link color, button style)
4. Remove all the currently unused or soon to be re-done components
5. Take a pass over defaults and ensure that they are sane.

It will probably be necessary to go back over everything again, but this is a good start at sane style.


jrjohnson:
@jenzweben take a look at this, I think it will make your overall job easier even if it require some extra initial work in adding back in styles that used to exist in bitters and are now removed.  If it works for you then assign it to someone to merge.

Issue #150, opened 2015-02-17
Session Detail Page


jrjohnson:
Finish session detail page.

@jenzweben please add your mock for session details.

jrjohnson:
![sessions-detail-page-v9](https://cloud.githubusercontent.com/assets/349624/6283131/a8b73a5a-b894-11e4-925a-29b6dfff700a.png)

jrjohnson:
@dartajax can you make sure your notes from our meeting yesterday find their way into this ticket?

dartajax:
The notes I have are the following ...

The link to go back to the main course page should be one of the following  ...
* Back to Main Course Page
* Back to Course Details

Session summary needs to show associated groups.

Independent Learning Modules need to be addressed.

Parent Objectives -- need to have a count not detail with a link to the Objective inspector.

dartajax:
What Valentina really wants is to have the page like the one above but also have the other sessions available -- in other words "View All / Edit " refers to sessions -- above and below "The Abdonimal Physical Exam Revisited" would be (in collapsed mode) the previous and / or following sessions (depending on how the screen has been sorted).  This would be huge for her if this could be possible.  She is aware the screen could be really long.

PR #197, opened 2015-02-21
Session Detail Page First Draft

jrjohnson:
This isn't done, but it is a start.  Everything but the offerings should be there and working.

I wanted to get something merged to we can take a style pass and also ensure that styles and functionality applied to course details are working for both course and session.

Refs #150


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/197" target="_blank">ReviewNinja

saschaben:
We may need to revisit the Session Objective display --- the 1:1 relationship of objective to parent works for this display in course, but the 1:many session relationship makes this display model ineffective.

saschaben:
I know all the technical reasons why we are loading a new page for session, but...since the course header is not persistent, and re-loads along with the session content, this creates a jarring contectual displacement for the user. And since we lose the context of location within a larger set of sessions in a course as well, we may have to rethink this a little. At the least we are going to need to re-label everything as "session-this" and "session-that".

saschaben:
Also: getting a lot of navigation errors when going back and forth from course page: " Error: Assertion Failed: calling set on destroyed object"

saschaben:
Missing functionality to set as Independent Learning, with instructional time instead of duration -- assuming we will handle this along with offerings?

jrjohnson:
@saschaben yes I agree that the transition is not working.  That's one of the reasons I wanted to get this out for discussion.  The mocks also have the course edit drop-down being available from this location, but I think maybe the right choice is to take the course off this page entirely except for the title.  I'm especially worried about the double publish buttons and status at the top, they are very confusing.

As far as ILM goes, I'm not sure we've added that functionality anywhere in the mocks.

saschaben:
@jrjohnson agreed and agreed: lots of challenges. None insurmountable. we will use Monday as our review session for this page, perhaps do some real time updating as a group. I believe @dartajax has already added a ticket for the missing ILM function, but we should review regardless, as we need to determine whether it goes in with the offering tools or with the general session tools.

This a a great start.

jrjohnson:
@saschaben this is ready for another review

PR #199, opened 2015-02-22
Preload Changes

### LINKED ISSUE #145: Fix prerender to inject into a script
Instead of injecting the JS into the page directly pre-render should add it to the end of vendors.  That way we don't get an inline script warning.

### LINKED ISSUE #155: Loading indicator jumps around
The pre-render loading indicator is flush left, but once the app loads the indicator moves into the center.

The best place to see this is on courses on a monitor large enough to have vertical white bars on the left and right of the application..  Just hit refresh and watch it jump.

jrjohnson:
Use an add-on to provide no script message.
Enclose initial loading in indicator in the same structure as the final site. (Fixes #155)
Refactor the pre-loader to put JS into include files and load them. (Fixes #145)

### DIFF
--- a/app/styles/components/_initialpage.scss
+++ b/app/styles/components/_initialpage.scss
@@ -1,0 +1,10 @@
+#initialpageloader {
+  p {
+    border: 2px dashed $ilios-orange;
+    color: $ilios-orange;
+    background-color: $background-grey;
+    margin: 2em;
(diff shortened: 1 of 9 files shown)


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/199" target="_blank">ReviewNinja

PR #200, opened 2015-02-23
Change all font-size declartions to be relative to $base-font-size

### LINKED ISSUE #187: modify course title detail view font to basefont
All other font sizes are specified as $base-font-size * .8 so if we ever change the base they will stay relatively the same. Do same to style for course title header. followup to #177, #186

jrjohnson:
Fixes #187


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/200" target="_blank">ReviewNinja

Issue #168, opened 2015-02-19
Hide Grid Header If No Data Associated - Learning Materials


dartajax:
The headers should probably be hidden if there is no data to display, no?  And across the board, this would seem like a good idea.

![lm](https://cloud.githubusercontent.com/assets/7493331/6260006/ca179cb0-b78d-11e4-8bc9-d272777f5087.JPG)

thecoolestguy:
We decided to have no header if there are no learning materials.

thecoolestguy:
Fixed in #205

PR #205, opened 2015-02-24
removed learning materials table from view if no learning materials exist

thecoolestguy:
also fixed a tr tag and did some tidying up, refs #168


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/205" target="_blank">ReviewNinja

PR #210, opened 2015-02-25
Working session objective manager

jrjohnson:


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/210" target="_blank">ReviewNinja

saschaben:
is the star on this for code review? If so please assign to me for UI review

saschaben:
cannot un-select objective once checked.
also:

we have some lag time on the checkbox action as well
we should be applying the same slide-in animation for this as we are for session/session list.

jrjohnson:
@saschaben are you sure about un-checking?  It works for me.  
We can't slide these in right now.  Too many layers to manage.  I'm 100% behind getting that going in a future release, but it adds significant complexity and will set us back weeks.

The lag time is also something to add to the list, but I want to get it up onto the real API before we start looking at which things need to perform better.

saschaben:
definitely can't un-check right now @jrjohnson

jrjohnson:
@saschaben - updated, give this another look.

saschaben:
check uncheck issue looks to be resolved. yay.
we still have the selected items dropping to the bottom of the list: they should stay in whatever sort order they are in to begin with (ID, alpha, whatever). Selected vs. unselected is not a good default ordering state.

jrjohnson:
@saschaben - ok they are sorting by title now.

saschaben:
right-o. let's code check and merge. We are still having some collision issues with the checkbox saves if someone clicks thru too quickly but we will address that as a bug later.

saschaben:
awaiting code review ninjastar.

PR #219, opened 2015-02-26
Help user make a good choice when publishing session and course

### LINKED ISSUE #202: Publish buttons on course and session displaying at the same time is confusing
@jenzweben please see it there is something we can do about this.
screen shot 2015-02-26 at 11 46 41 am

jrjohnson:
Fixes #202

@saschaben - see if this will work, or if we need to look at another way to make accidentally publishing a course more difficult.  Buttons are now hidden on course objective manager and when the course details more/edit gizmo is closed.

### DIFF
--- a/app/templates/components/course-header.hbs
+++ b/app/templates/components/course-header.hbs
@@ -12,8 +12,10 @@
     <h4>{{course.academicYear}}</h4>
   </span>
   <span class ='actions'>
-    <button class='publish' {{bind-attr disabled=course.isNotPublished}}>{{t 'general.unPublish'}}</button>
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/219" target="_blank">ReviewNinja

saschaben:
Short answer: No.

Longer answer: This is the least optimal solution to use, as discussed in the meeting. It removes a simple behavior from the system and makes it less transparent to the user. Since publishing is NOT a part of editing, but a workflow action that is distinct from all edit activity, it should not be bound to edit/non-edit view state.

I would rather relabel and re-style the course publish buttons universally to distinguish from session publish, rather than do this.

PR #234, opened 2015-02-26
fixed inline editing controls alignment problem, fixes #221

### LINKED ISSUE #221: inline edit fields and their controls should have consistent placement
Currently the course edit fields are aligned to provide some really unusual behavior if opened simultaneously: 

ilios 2015-02-26 09-40-36

-> Controls should remain persistent at the right margin of the entry form field, not jump to below the field;
-> If spacing is an issue it owuld be preferable to add a new row and place end date persistently below start date, rather than hacking the order of items so the dates stack appropriately;
-> there needs to be enough margin/padding around the controls to allow for visual distinction of which set of controls belong to which form field when multiple fields are open in a smaller screen

jenzweben:


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/234" target="_blank">ReviewNinja

app/styles/components/_global.scss
```diff
@@ -138,7 +140,7 @@ background: #FFF;
     z-index: 99999;
     overflow: show;
 padding: 4px;
-width: 100%;
+// width: 100%;
```
jrjohnson: Instead of commenting this out can you just remove it?

jrjohnson:
Replaced by #239.

PR #239, opened 2015-02-27
Updated #234

jrjohnson:


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/239" target="_blank">ReviewNinja

app/styles/components/_global.scss
```diff
@@ -128,17 +129,14 @@ textarea {
...
-    cursor: pointer;
-    z-index: 99999;
-    overflow: show;
-padding: 4px;
-width: 100%;
-
- }
 
+input[type=text] {
+  background: #FFF;
```
thecoolestguy: shouldn't we set text input fields' background's as a var?

PR #243, opened 2015-02-28
Faster Course List Loading

### LINKED ISSUE #207: Use filter for all queries
I've been handling fetching data incorectly.  When we grab a large set of data isntead of doing a store.find, do a store.filter and pass the query params as the second argument.  The ensures that we get fast results from the cache and it background loads everything else.
http://emberjs.com/api/data/classes/DS.Store.html#method_filter

jrjohnson:
Refactor course/objective/competency properties to be PromiseArrays instead of observers.  This ensures the courses list doesn't have to load everything.

Fix an issue with the course list double loading on some requests.

Fixes #207

### DIFF
--- a/app/components/detail-competencies.js
+++ b/app/components/detail-competencies.js
@@ -5,27 +5,23 @@ export default Ember.Component.extend({
   domains: function(){
     var domainContainer = {};
     var domainIds = [];
-    var competencies = this.get('competencies');
-    if(competencies){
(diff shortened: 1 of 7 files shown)


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/243" target="_blank">ReviewNinja

saschaben:
Nice improvement.

PR #237, opened 2015-02-27
Add Course Publication Workflow

jrjohnson:
Allows courses to be scheduled, published, and unpublished.
Displays publication issues on a separate route.

Fixes #141
Fixes #229

PR #244, opened 2015-03-01
Better Topic Saving

### LINKED ISSUE #135: Janky course topic function
Rapid adding and removing of topics causes the topics list to unload and an error to be thrown which causes further navigation or page action to fail. See also #134, #133, #129.

jrjohnson:
Several improvements to changing course and session topics.  
1.  Save faster
2. When saving don’t let the user take any other actions.

There was no way to make the save fast enough or smooth enough to prevent the ‘Janky’ behavior.  Instead I placed an overlay on top of the edit area to prevent the user from taking other actions.
Fixes #135

### DIFF
--- a/app/components/course-editing.js
+++ b/app/components/course-editing.js
@@ -38,16 +38,6 @@ export default Ember.Component.extend({
         });
       });
     },
-    addTopic: function(topic){
-      var course = this.get('course');
-      course.get('disciplines').addObject(topic);
-      course.save();
-    },
-    removeTopic: function(topic){
-      var course = this.get('course');
-      course.get('disciplines').removeObject(topic);
-      course.save();
-    },
     addMeshDescriptor: function(descriptor){
       var course = this.get('course');
       course.get('meshDescriptors').then(function(descriptors){
--- a/app/components/course-editing.js
+++ b/app/components/course-editing.js
@@ -40,25 +40,13 @@ export default Ember.Component.extend({
     },
     addTopic: function(topic){
       var course = this.get('course');
-      course.get('disciplines').then(function(topics){
-        topic.get('courses').then(function(courses){
-          courses.addObject(course);
-          topics.addObject(topic);
-          course.save();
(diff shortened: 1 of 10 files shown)


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/244" target="_blank">ReviewNinja

jrjohnson:
@saschaben - see if this pattern of overlaying a loading indicator to prevent actions between saves makes sense to you.  I'm guessing we will want to use it everywhere.
We may want to replace 'Loading Ilios...' with 'Saving...' and we can also get rid of the text entirely and simply put in the loading wave.  @jenzweben may also be able to make the transition and position better.

saschaben:
on attempting to either load or unload a topic after initially loading and opening the course, frequently getting the following error:

ilios 2015-03-01 18-35-42

suspect it happens when the course data is not fully loaded prior to topic action. Occurring on multiple courses, only on initial action after course load -- if it doesn't happen then, it won't happen at all. Hence my above hypothesis.

The delay in save and render is still verging on unacceptably long here: a 5-7 second delay might be OK for a full course save, or a publish, or clone, or somesuch, but for just adding or removing a tag, I doubt this will fly with users. The overlay function as a pattern -- if we can speed it up by a factor of five -- is just fine.

jrjohnson:
Your error is almost certainly fixed by #237.  You can confirm by not navigating to a course through the course list but by refreshing the URL directly in that case you shouldn't ever have this problem.  Some of the speed issue may be purely the artificial delay added to the dev API, but I would make that a different ticket since we need to confirm with a more accurate server.

saschaben:
OK -- looking at the time differential on loads I was wondering if this was a dev API thing. that makes sense.

I'll star this and add tix for the load issue and an updated loading wave. If we continue to encounter the load error after merge we can look into that, but I'll take your word on the #237 fix.

saschaben:
waiting for code review.

Issue #218, opened 2015-02-26
Style Objective Manager


jrjohnson:
objectivemanager.scss needs to be fixed, it lost a lot of its defaults in the #189 changeover.  Styles course and session objectives.

jenzweben:
I've styled as much as I can for the moment - there are a few things that we need to fix in the template.  For some reason, the header of the section of a selected item is having the selected item styling applied to it...

Attaching the mock here for reference:
![screen shot 2015-03-02 at 12 45 13 pm](https://cloud.githubusercontent.com/assets/5023894/6450137/9dcb3d3c-c0da-11e4-8c4d-8bd84069a944.png)

Screenshot of header taking on selected item styling:
![screen shot 2015-03-02 at 12 48 34 pm](https://cloud.githubusercontent.com/assets/5023894/6450192/fef3d696-c0da-11e4-8a1b-0d6bd579b843.png)

jrjohnson:
The header picking up selected style is intentional, indicates which competency is selected.

PR #249, opened 2015-03-02
fixed styling for list items and header but we need to tweak the templat...

jenzweben:
...e a bit..., refs #218

### DIFF
--- a/app/styles/components/_objectivemanager.scss
+++ b/app/styles/components/_objectivemanager.scss
@@ -5,16 +5,36 @@
     background-color: $background-grey;
   }
   .selected {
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/249" target="_blank">ReviewNinja

saschaben:
Per @jenzweben, still need to restyle the objective text itself, and remove "Select Parent Objective" tag.

jrjohnson:
hold off on this until I get the in place stuff finished.  I don't think its worth doing before that since the template has changed slightly.

PR #261, opened 2015-03-05
Working add new session workflow

### LINKED ISSUE #147: Active add new session action on course detail page
What happens when a use clicks add new in the sessions section?

@jenzweben please provide the mock for this.  I think we should use the slide down form we are going to use on the courses list.

jrjohnson:
Fixes #147


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/261" target="_blank">ReviewNinja

saschaben:
Is missing the session type selector.
on creation:addnew1
after creation: addnew2

Otherwise, seems fine.

jrjohnson:
Well it should have loaded eventually, but I just added a default state.  Let me know if this makes sense.  Before the list loads it will be a single item select box and the text is 'Loading Session Types...'

saschaben:
On initial access, this flow is working: list waits, loads with default, hey presto. BUT:

save a session. Enter the new session. Return to list. Add new session...and the list never reloads. just hangs on the loading state. no errors thrown.

ilios 2015-03-05 10-17-16

saschaben:
wondering if this could be a bug similar to the one we saw previously with page loads after a back and forth from the course list, a la #133 ...

jrjohnson:
@saschaben - this is ready to test again.

saschaben:
Looks like that took care of it. let's get this reviewed styled and merged and move on.

PR #265, opened 2015-03-05
Break user search into its own component

### LINKED ISSUE #166: Create separate livesearch component specifically for returning user search results
As user searches should return multiple useful values (full name, first name, lastname, email, uid, etc), instead of just a 'title', we need to set up a separate livesearch component solely for the searching of user information. (eg, in #161, the returned user information requires that each component be wrapped in its own respective class for styling -- this cannot happen if only the complete 'title' value is returned)... 

Other searches will use the current, generic livesearch component that returns a title only, but the user search should return several useful values.

jrjohnson:
Live search was not flexible enough to be styled

Fixes #166

### DIFF
--- a/app/components/course-overview.js
+++ b/app/components/course-overview.js
@@ -1,5 +1,4 @@
 import Ember from 'ember';
-import LiveSearchItem from 'ilios/mixins/live-search-item';
 
 export default Ember.Component.extend({
   editable: false,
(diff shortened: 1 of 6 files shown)


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/265" target="_blank">ReviewNinja

thecoolestguy:
fixes #166

PR #266, opened 2015-03-06
Editing topics in place

### LINKED ISSUE #259: Can we use the objective manager slide in other places?
I'm specifically thinking about topics and mesh terms on a course or session.  It would solve our save / loading issue if we displayed the topics and had a button to edit.  When editing was selected the current add / delete interface would slide in and changed could be made.  When they were done they can save or they can undo.

Thoughts?

jrjohnson:
Instead of allowing immediate editing of each topic use a manager with an undo action.

Fixes #259

### DIFF
--- a/app/components/detail-topic-list.js
+++ b/app/components/detail-topic-list.js
@@ -1,0 +1,7 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  topics: [],
+  sortBy: ['title'],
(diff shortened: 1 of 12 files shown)


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/266" target="_blank">ReviewNinja

saschaben:
looks good. Ready for code review. (note: is there any way to get the livesearch "cancel" icon next to the search entry field, rather than right aligned to the container itself?)

Issue #272, opened 2015-03-06
Objective manager should use the same pattern as topics


jrjohnson:
In #266 we made the topic manager undoable and sped up the process by not saving changes until the user hits the checkmark to save.  We need to do the same for the course and session objective manager.

PR #273, opened 2015-03-06
Replace checkboxes with radio buttons for course objective manager

### LINKED ISSUE #254: Radio buttons on course parent objective picker
@jenzweben I think I may have nixed this originally, but that limitation is gone.  Is there any reason not to use radio buttons on the course parent objective picker?  We can still use the checkboxes on the session one.  I think this is a better visual indication that you can only select one at a time.

jrjohnson:
Fixes #254


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/273" target="_blank">ReviewNinja

saschaben:
Looks fine. ready for code review and merge. Will be even better once we address #272 though.

PR #274, opened 2015-03-06
Better display for course objective parents from multiple cohorts.

### LINKED ISSUE #248: Design better consistent display for parent objectives
Need a way to display one or more parent objectives that is consistent and does not cause a context break between showing one and showing many.

For course: as discussed, let's go with the current _Parent Obj. Text (Parent Competency)_ when only one parent chosen (ie only one cohort), and _Name of Cohort (Parent Comepetency)_ for more than one, along with some sort of tooltip / info link to show full info. should use ellipsis on these as well.

For session: we need something that will be more generic for both 1 or 1+n patterns, that won't discourage the actual linking of multiple parents, nor make the content of those parents completely opaque.

jrjohnson:
Fixes #248


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/274" target="_blank">ReviewNinja

saschaben:
Awesomeness. Two issues though: 
1. For table listing, I think the `<tr>` alignment should be top, so that course objective description and mesh button don't hang in the middle; (this also gives me some ideas on how to handle session in a simpler way);
2. The display of the cohort name is not dependent on the addition of multiple parents to the objective; it is predicated on the addition of more than one cohort to the course. So, if there are two cohorts attached to a course, ALL parent objectives get a cohort identification, regardless if there is a parent for each cohort or not. There should never be a situation like the example now, of one objective with cohort labels and others without. It's either all with none or all with one.

jrjohnson:
@saschaben - I fixed the display as requested, nice catch, I completely mis-understood that requirement.  I added a new ticket for the alignment issue (#279)

saschaben:
Looks great. Ninja'd for functional review.

PR #276, opened 2015-03-06
stylized select elements to resemble action menu, added colon after title label, added dropdown arrow image to assets

### LINKED ISSUE #262: Style basic select list so that it matches the look & feel of the action dropdown

thecoolestguy:
unfortunately, the select option highlighting is still unable to be stylized in modern browsers, but this is a close as it gets, fixes #262


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/276" target="_blank">ReviewNinja

thecoolestguy:
a dropdown using a true \<select\> element can be viewed when adding a new session and selecting the session type

saschaben:
fine by me. buttons still need padding applied -- adding a new ticket for styling so we can merge.

app/styles/components/_global.scss
```diff
@@ -148,3 +148,31 @@ input[type=text] {
   overflow: show;
   padding: 4px;
 }
+
+/* style a regular select element to look like the actions menu */
+select {
+  /* set the dropdown arrow image as the background to simulate the actions menu dropdown */
+  background: url(/images/select-down-arrow.png) 99% / 2% no-repeat $default-background-color;
```
jrjohnson: I don't think we want to force the image to reside in / - I think images with no / will work.

app/styles/components/_global.scss
```diff
@@ -148,3 +148,31 @@ input[type=text] {
...
+  border-color: $header-grey;
+
+  /* handle the text and padding */
+  font-size: 1em;
+  height: 1.8em;
+  line-height: 1.1em;
+  padding-left: .5em; 
+ 
+  /* set up the corners of the select box to mimic the rounded button/input corners */ 
+  -webkit-appearance: none;
```
jrjohnson: instead of browser based selectors use the bourbon library to set appearance:
http://bourbon.io/docs/#appearance

app/styles/components/_global.scss
```diff
@@ -148,3 +148,31 @@ input[type=text] {
...
+  -moz-appearance: none;
+  appearance: none;
+  border-radius: $base-border-radius;
+  -webkit-border-radius: $base-border-radius;
```
jrjohnson: Use bourbon for this: http://bourbon.io/docs/#border-radius

app/styles/components/_global.scss
thecoolestguy: When I do that, it wants to pull from 'assets/images/'... Wasn't sure how to fix... I'll look for the solution.

app/styles/components/_global.scss
thecoolestguy: using bourbon method now...

app/styles/components/_global.scss
thecoolestguy: using bourbon method now

saschaben:
re-reviewed: fine. but see #277.

app/styles/components/_global.scss
thecoolestguy: looks like we should have a public/assets/images folder

http://www.ember-cli.com/#asset-compilation

I created one and moved the image there and it seems to work

PR #278, opened 2015-03-07
Improved Objective Management

### LINKED ISSUE #209: Allow for course objective parentage on multi-cohort courses
When a course has more than one cohort attached, objectives will have a unique parent objective for EACH cohort. Currently the model is challenged by:

A: Inability to select parent for each cohort (1. our return function keeps you from doing a two-fer; 2. selecting a parent for one cohort will override the selection for the other);
B: Inability to display multiple parents easily in the course details.

@jrjohnson, let's modify the selection model to allow for the unique selection of a parent for each cohort associated (and obviously, from that cohort's specific competency set); and we will need to revisit my clever click -return to page idea, as that toasts the selection of a parent for each cohort at the same time. not good.

@jenzweben let's discuss the display issues for this and see if we can come up with an alternate that allows for the use case of  > 1 cohort per course.

### LINKED ISSUE #272: Objective manager should use the same pattern as topics

jrjohnson:
Courses can select objective parents from any linked cohort (fixes #209)

Objective manager does not save live, it waits for user confirmation like the topic manager. (fixes #272)

### DIFF
--- a/app/components/course-objective-manager.js
+++ b/app/components/course-objective-manager.js
@@ -145,8 +145,10 @@ export default Ember.Component.extend({
         var newProgramYear = programYears.get('firstObject');
         newProgramYear.get('objectives').then(function(oldParents){
           oldParents.forEach(function(oldParent){
-            courseObjective.get('parents').removeObject(oldParent);
-            oldParent.get('children').removeObject(courseObjective);
+            if(oldParent.get('id') !== newParent.get('id')){
+              courseObjective.get('parents').removeObject(oldParent);
+              oldParent.get('children').removeObject(courseObjective);
+            }
           });
           courseObjective.get('parents').addObject(newParent);
           newParent.get('children').addObject(courseObjective);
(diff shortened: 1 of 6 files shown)


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/278" target="_blank">ReviewNinja

saschaben:
Very nice. Passes full muster on COURSE level. BUT.....

On SESSION Level: for a session objective with 1 + _n_ parent objectives, selecting a single parent currently deletes all parents. For a session with 1 + _n_ parent objectives, adding an additional parent or parents will save only the newly selected item(s), and will delete the pre-existing parent(s).

jrjohnson:
Part of the problem is that the display is not refreshing when changing from the states no parents, at least one parent, more than one parent - and back.  This is resolved with  #274 - I'm going to wait for that to be merged before debugging the issue with session parents.

jrjohnson:
@saschaben - please test again.

saschaben:
looks good.

PR #284, opened 2015-03-09
Active the Course print summary button

### LINKED ISSUE #139: Activate Print summary button on course details
@jenzweben - please attach your mock for that output.

jrjohnson:
Instead of trying to use the same templates as we use for course details I copied most of the HTML structure into a single print course template.  This may cause some divergence later, but it was the cleanest way to get a printable course view without cluttering the rest of the templates with ```{{#if printing}}``` garbage.

Load is pretty slow and as a next step we may want to pre-load the data in the route so we get our usual loading indicator instead of moving to the white screen.

I left the content mostly un-styles, but added enough hooks that it should be possible to clean it up without changing the structure.

Fixes #139

### DIFF
--- a/app/components/course-overview.js
+++ b/app/components/course-overview.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
   editable: false,
+  printView: false,
   directorsSort: ['lastName', 'firstName'],
   directorsWithFullName: Ember.computed.filterBy('course.directors', 'fullName'),
   sortedDirectors: Ember.computed.sort('directorsWithFullName', 'directorsSort'),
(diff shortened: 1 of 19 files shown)


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/284" target="_blank">ReviewNinja

saschaben:
Missing competencies.

jrjohnson:
@saschaben - competencies back in.  I added a loader for them; they take forever to load and render because they have to dig out so much objective and course data.

thecoolestguy:
ninja-starred for code

jrjohnson:
resolved conflicts - needs a cursory functional review and then can be merged.

PR #282, opened 2015-03-09
For ILM Sessions give the due date as the first offering date

jrjohnson:
Added dueDate and hours to ILM Session Model

Fixes #84

PR #285, opened 2015-03-09
Revert "For ILM Sessions give the due date as the first offering date"

saschaben:
Reverts ilios/frontend#282

I believe this needs to be reverted --- the ILM dates showing are always the value `currentDate (12:00:00am).`


review-ninja:
<a href="http://app.review.ninja/ilios/frontend/pull/285" target="_blank">ReviewNinja

PR #302, opened 2015-03-12
Create a loading-part component to handle internal loads

jrjohnson:
Right now this just calls the wave loader, but eventually we will replace it with the heartbeat or another smaller indicator

### DIFF
--- a/app/components/loading-part.js
+++ b/app/components/loading-part.js
@@ -1,0 +1,6 @@
+import Ember from 'ember';
+import layout from '../templates/components/loading-part';
+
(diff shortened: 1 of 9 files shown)


saschaben:
@jrjohnson I see this loading groups, but nowhere else. we still load blank space for directors and for session type / time on course page load.

jrjohnson:
Thats the only place it exists currently and the only place I replaced in this ticket.  You should ticket any other places you would like to see he internal loader.

saschaben:
that's what I thought -- just checking. this should get reviewed and merged then.

Issue #292, opened 2015-03-09
User Mirage to mock api data


jrjohnson:
https://github.com/samselikoff/ember-cli-mirage
Will handle both our fake server data for development as well as factories for testing.

jrjohnson:
closed in favor of #288

PR #303, opened 2015-03-13
Add Objective MeSH descriptor manager

### LINKED ISSUE #138: Add objective mesh picker to course details
We need a generic mesh attachment component that will work for course objective and for other areas where this relationship is created.  It should be based on the course objective manager.

Implement it initially when you click 'Add New' under Mesh Terms on course objectives.

@jenzweben - can you provide the mock for this?  Please work with @saschaben to identify any rules for associating mesh terms to a course objectives (many, just one) and how they might be different for session objectives or program objectives.

jrjohnson:
Added to courses as well as sessions and TESTS!
Also added some basic tests for the parent manager.
Fixes #138


jrjohnson:
This now takes too long to run - so the browser tests can't pass.  I will need to re-do fixtures to take out almost everything and make them faster.
Requires #304

jrjohnson:
OK - tests fixed and using ember-cli-mirage which is way better.  I severly limited the fixtures in dev so we may want to add those back in over time.  
Now also fixes #292

saschaben:
@jrjohnson assuming that the new fixtures are severely limited in what Mesh values can be returned, this seems to be working for search. But in order to actually test the mesh system, we really need to hook this up to the full mesh universe and determine if it provides correct result sets in a meaningful timeframe...

Issue #295, opened 2015-03-09
CODE THE OFFERINGS


saschaben:

jrjohnson:
Mock in #150

jrjohnson:
Closing with #327 - individual improvement will go in their own tickets.

PR #327, opened 2015-03-18
Very basic view of offerings

jrjohnson:
This is a mostly unstyled rendering of the offerings in a session to test assumptions about the layout and get something on the board to start iterating against.

Refs #295


saschaben:
This looks fantastic.

Currently though it appears that we are sorting groups of offerings within a given day by start time DESC, where it should list them ASC:

ex. 1
offering1

ex. 2

offering2

PR #343, opened 2015-03-19
Added offerings to course print view

### LINKED ISSUE #294: include offerings in print view
include all offering info on the print view

jrjohnson:
Has no tests because this is not our long term solution

Fixes #294


saschaben:
looks great.

PR #350, opened 2015-03-20
Add ILM options to the session

### LINKED ISSUE #165: Need to Have the Independent Learning Check Box on Session Edit / Add plus Due Date Field
In looking at the mockups, I am not seeing where the check box to indicate whether a session is an independent learning session type (Independent Learning Module).  Also missing - the due date - is this a formatted text field or date picker of some sort?

jrjohnson:
Bunch of file churn here as I removed the no longer relevant ilm-session-facet (was replaced by ilm-session a while ago)
I also added a sessionDescription factory which required adding a reference to several unrelated tests.

Fixes #165


saschaben:
As far as this goes, it's good. But it's only half the feature. Please see #353 and #354 bug entries.

PR #351, opened 2015-03-20
Add clerkship type selection to course overview

### LINKED ISSUE #321: Courses and Sessions >> Clerkship Type is Missing
I am not seeing the Clerkship Type drop-down.  Is that something we are no longer using?

jrjohnson:
fixes #321

### DIFF
--- a/app/components/course-overview.js
+++ b/app/components/course-overview.js
@@ -3,12 +3,13 @@ import DS from 'ember-data';
 
 export default Ember.Component.extend({
   editable: false,
+  course: null,
   directorsSort: ['lastName', 'firstName'],
   directorsWithFullName: Ember.computed.filterBy('course.directors', 'fullName'),
   sortedDirectors: Ember.computed.sort('directorsWithFullName', 'directorsSort'),
   levelOptions: [1,2,3,4,5],
   classNames: ['course-overview'],
-  sortedClerkshipTypes: function(){
+  clerkshipTypeOptions: function(){
     var deferred = Ember.RSVP.defer();
     this.store.find('course-clerkship-type').then(function(clerkshipTypes){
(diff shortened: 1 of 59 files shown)


thecoolestguy:
needs a rebase to merge...

saschaben:
Clerkship type requires the default selection to be blank, or "not a clerkship".

jrjohnson:
@saschaben this is ready for another look.

saschaben:
nearly: Currently, if the course is not a clerkship, then the field displays as blank on load, rather than displaying "not a clerkship". When you  open the course edit, the "click to edit" default text is what displays when "not a clerkship" is selected (or when the course is not a clerkship).

"Not a Clerkship" should _actually_ be the default display value. I should also be able to select "Not a Clerkship" and see that persistently displayed. I recognize that this wreaks havoc with the controller model for an inline field presenting "Click to edit" when it's blank. But it's what needs to happen. (and no, adding it as a value in the database and writing a migration script doesn't work, because "not a clerkship" isn't a type of clerkship, and can't be reported as such. sigh.)

jrjohnson:
I actually had this in my design, just forgot to implement.  It's one of the reasons I had to create a new component to handle this. Nice catch, quick fix... the ideal situation.

saschaben:
Sweet. Due to the delay in the data call, we will need to add an inline loading animation to this --- right now, when you switch from non-editable to editable view, there is a 1+ second delay where if the value is anything other than "Not a Clerkship" (ie it IS a clerkship), it flips to "Not a Clerkship for a second, until the data is loaded into the page. A little unsettling. I have added a separate bug issue for this (#360)

This can get reviewed and if it passes, starred and merged.

PR #361, opened 2015-03-23
Clear the user search when there is no input.

### LINKED ISSUE #312: Directors live search "keep typing" persists after searching
even after adding a user and clicking outside the area...and it shouldn't...this is an icing on the cake fix

screen shot 2015-03-16 at 1 38 55 pm

jrjohnson:
Fixes #312


thecoolestguy:
Needs functionality verification... Code looks good.

PR #364, opened 2015-03-23
Catch clear actions on mesh search so when the search box is cleared so are all the search results

### LINKED ISSUE #242: MeSH Close/Cancel does not work
When searching MeSH, hitting the cancel/close clears only the search term value in the input box; it does not clear the current search results nor close the current search return window.

jrjohnson:
Fixes #242


thecoolestguy:
code checks out, please review functionality

PR #369, opened 2015-03-23
Updated ember-cli-sauce

johanneswuerbach:
Use bundled connect command
Run 2 browsers in parallel (sauce free accounts can run up to 3) to speed up CI runs.


jrjohnson:
@johanneswuerbach love the parallelization - cuts test time in half!

PR #370, opened 2015-03-23
Set default status and role for new learning materials

### LINKED ISSUE #345: "Add Learning Material" default values not persistent
When adding multiple learning materials, after first save, default values for status and role do not re-load when the form is re-opened for adding the next material.

These should be persistent on every load of the widget, and not require a full page refresh.

jrjohnson:
Fixes #345


thecoolestguy:
The default status of learning materials was recently changed to 'final'.  We should make it so.

saschaben:
looks good. review and merge.

Issue #309, opened 2015-03-13
Update style for headers of information in publish "Review" pane


saschaben:
Right now, there is no visual distinction for the user between _required_ and _desired_ fields in the review pane(s). We need to clarify for the user so that there is no confusion regarding why something may be unpublishable -- and have no option to publish that is visible. A simple fix would be to un-bold all desired fields, and leave bold required fields. Or something along those lines.

COURSE LEVEL:
**Course Title** - required
**Cohorts** - required
_Topics_ - desired
_Objectives_ - desired
_MeSH Terms_ - desired

![course 2015-03-13 12-40-08](https://cloud.githubusercontent.com/assets/1753431/6646041/777d2616-c97e-11e4-9ae1-5d53e995b786.png)

SESSION LEVEL:
**Session Title** - required
**Offerings** - required
_Topics_ - desired
_Objectives_ - desired
_MeSH Terms_ - desired

![session 2015-03-13 12-39-40](https://cloud.githubusercontent.com/assets/1753431/6646027/54f3e59e-c97e-11e4-93f7-c4e481323b7a.png)

jenzweben:
How about this?  
![screen shot 2015-03-23 at 10 34 34 am](https://cloud.githubusercontent.com/assets/5023894/6786237/479a39d0-d148-11e4-9131-a9fcb578dfbe.png)

saschaben:
@jenzweben this looks good -- especially if we can provide a mouseover tooltip on the red asterisk and legend "you will not be able to publish until all required fields are completed" ( @jrjohnson ??)

PR #374, opened 2015-03-24
For ILM Sessions offerings are not required.

### LINKED ISSUE #353: Publish flow incorrect on ILM session fields
When a session is checked as "Independent learning", its nature changes: it has a due by date which is both required and unavoidable. It never has offerings. So when testing state for publishing, an ILM session should NOT check for having offerings, and publishing SHOULD be available without them.

jrjohnson:
Fixes #353

### DIFF
--- a/app/models/session.js
+++ b/app/models/session.js
@@ -73,7 +73,10 @@ var Session = DS.Model.extend({
     var requiredSet = [
       'title'
     ];
-    var requiredLength = ['offerings'];
+    var requiredLength = [];
+    if(!this.get('isIndependentLearning')){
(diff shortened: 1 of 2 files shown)


jrjohnson:
I made it possible to publish an ILM with no offerings.  I'm not sure how to present the publication check for ILM sessions should offerings be in the table - maybe that can be accounted for in #309?

saschaben:
The ILM review should instead of "offerings" have a header of "due by date" as a required field.

saschaben:
OK for now, but this workflow needs additional refinement from the user side. let's merge and discuss.

PR #380, opened 2015-03-24
replaced button element with span to fix search bug in safari

### LINKED ISSUE #176: Director Search box doesn't work in safari
Can't type any values into the director search box in Safari. This follows on the correction of #97.

thecoolestguy:
changed magnifying glass icon to span and removed action to fix safari bug, fixes #176

### DIFF
--- a/app/styles/components/_searchbox.scss
+++ b/app/styles/components/_searchbox.scss
@@ -21,22 +21,13 @@
         overflow: show;
         min-width: 220px;
     }
-    button[type=search], span[type=search] {
+    .search-icon {
         @include position(absolute, 0);
-        background-color: transparent;
         color: $base-link-color;
-        text-align: left;
+        display: inline-block;
         padding-left: 6px;
         margin-right: 2px;
-        img {
-            opacity: .7;
-        }
-    }
-    /* use inline-block displayed span instead of button for inset magnifying glass icon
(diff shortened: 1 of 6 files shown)


thecoolestguy:
This fixes the director search bug in Safari, but prevents clicking on the actual magnifying glass icon to acheive focus to the search box... We can add some js later to remedy this, but the actual search functionality is working on Safari now.  Please review.

app/styles/components/_searchbox.scss
```diff
@@ -21,7 +21,7 @@
         overflow: show;
         min-width: 220px;
     }
-    button[type=search] {
+    button[type=search], span[type=search] {
```
jrjohnson: This button isn't in the structure anymore so the style can be removed.  I also think span should probably get a class of search instead of a type which I think it just for buttons.

saschaben:
looks good so far, please alert when revisions noted by @jrjohnson are implemented (or not) and will re-review.

jrjohnson:
@thecoolestguy - if this is ready can you rebase and see if tests will pass?

saschaben:
Can we review this again? I think it's gone out of sync with the rest of code, cannot currently load resource locally.

saschaben:
Yay. Director search in Safari passes muster. Ninja this!

PR #256, opened 2015-03-03
added call to translation for 'New course'

thecoolestguy:

PR #436, opened 2015-04-07
changed span-columns, shift, and omega settings, refs #256

thecoolestguy:
Is responsive now... Still limited by non-responsive html table beneath it, but no longer wrapping and/or scroll bar appearing.


saschaben:
this is still very weird-feeling. Before we merge, let's review in the morning. I want to be sure to correctly enter any followup issues here.

Issue #376, opened 2015-03-24
Courses and Sessions >> Session >> Add Description >> Save not Occurring


dartajax:
I noticed while trying to add a description to a Session that did not have a description, the green check button wasn't processing the save event.

![wontsaveohno](https://cloud.githubusercontent.com/assets/7493331/6807503/12a0c07c-d20c-11e4-96ec-f2f27cffd723.JPG)

Clicking the green check icon does not do anything.  I hope this is not a duplicate - please close if so.

jenzweben:
Same here

jrjohnson:
This will be resolved by #359 I will leave both open for now.

jrjohnson:
Closed by #455

PR #455, opened 2015-04-17
Replace editable with inplace components

### LINKED ISSUE #359: Re-write editinplace componenets
The inplace-select in #351 is a vast improvement over the current components since it sends actions and allows more customization.  It needs to be generalized a bit by putting the common stuff in a mixin.  Then replace all the current text, text area, checkbox, edit-in-place components with this improved pattern.

jrjohnson:
New components don't attempt to save changes on their own, instead they send actions to the calling component.  This allows us to eliminate a lot of mysterious code that didn't really work anyway.

Also added tests for the places where we use this:
course overview changes, session overview changes, objective title, learning material changes.

Fixes #359 #376


jrjohnson:
Seems to be an issue with testing in IE 10.  I'm looking at solutions or we can wait for it to get fixed upstream.  I reported it.

PR #457, opened 2015-04-17
Kill sauce connect tunnel when tests are done

jrjohnson:
We were leaving the sauce connect tunnel open on every test which quickly becomes a problem when we reach our limit of 5 open connection.


jrjohnson:
@saschaben or @thecoolestguy please merge this ASPA it will make our other tests a lot more stable.

PR #472, opened 2015-04-21
Added programs landing page

### LINKED ISSUE #462: Create Programs Landing Page
programs

jrjohnson:
Programs are listed, filterable, removable, and new programs can be
added.  Also added a blank program route so the navigation works, but
the result is empty white space.

Fixes #462


saschaben:
good to go.

PR #476, opened 2015-04-23
Add Program Detail Page

### LINKED ISSUE #463: Create Program Detail Page
programs-detail

jrjohnson:
Also added program publication check, but its incomplete.

Fixes #463

### DIFF
--- a/app/components/inplace-boolean.js
+++ b/app/components/inplace-boolean.js
@@ -4,4 +4,12 @@ import InPlace from 'ilios/mixins/inplace';
 export default Ember.Component.extend(InPlace, {
   classNames: ['editinplace', 'inplace-boolean'],
   saveOnChange: true,
+  actions: {
+    toggleValue: function(){
+      this.set('workingValue', !this.get('workingValue'));
+      if(this.get('saveOnChange')){
+        this.send('save');
+      }
+    }
+  }
 });
--- a/app/components/inplace-select.js
+++ b/app/components/inplace-select.js
@@ -46,4 +46,12 @@ export default Ember.Component.extend(InPlace, {
 
     return proxies;
   }.property('options.@each', 'optionLabelPath', 'optionValuePath', 'selectPromptTranslation'),
+  actions: {
+    changeSelection: function(newValue){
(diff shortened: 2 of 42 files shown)


saschaben:
looking good. need to address formatting weirdness in header and divider lines, and there is something odd going on with the publish button, but we will review in meeting today and any follow-on issues can be scoped and logged then.

jrjohnson:
@saschaben  - please test again.  I fixed the publish button as well as adding the ability to create new program years.

saschaben:
We need to consider further the disallowing of changing the academic year setting on a program year: I'm a little (but only a little) concerned about the "pick it once and you're stuck" model. Sure, it's easy enough to just delete and recreate -- as long as no work has been done. But for now I think this pattern should be fine.

Otherwise, all good. An interesting side effect of the hard-coded api data: until you add a new program year, publish tool doesn't recognize years as existing, and so does not allow for publish -- even though pre-existing test data is there. As long as this doesn't bleed over into real use, it should be all good.

saschaben:
OK. 99.99% there. I am having a trouble saving a program duration of value = 1; each attempt to select a value of n=1 just reverts back to "click here to edit" default on save. I'm guessing this is due to the defaulting of the select list to the initial value (1). I think this may have been adding to the publish process confusion earlier (since we require duration).

We can go ahead and merge, but please create a new bug ticket to address this in a later prerelease if it's not a super-quick fix that can be slipped in now.

jrjohnson:
@saschaben I'm unable to duplicate the change to 1 issue.  I added a specific test case to check for this so it will run in our browser matrix to detect if this is a browser specific bug.

saschaben:
K. Getting this in Chrome and Safari. you can see the behavior here:

https://www.dropbox.com/s/2z5z365a0wihp4g/n%3D1_screencast.mp4?dl=0

jrjohnson:
@saschaben - ok I think I fixed it.  Give it another look please.  May need to do further refinement on the generic select component we are using here - I added that as #489

saschaben:
got it. current fix now works, default to 4 is a-ok.

PR #497, opened 2015-04-29
relocated pulse loader animation

jenzweben:
Tested in FF, Chrome & Safari at various window widths


saschaben:
nice!

 needs ninja review to merge.

PR #499, opened 2015-04-30
Added header to session offerings

### LINKED ISSUE #424: Add Header to Session Detail Page - Offerings Section
Currently looks like this:

screen shot 2015-04-06 at 1 30 43 pm

Should look like this:

screen shot 2015-04-06 at 1 28 50 pm

thecoolestguy:
fixes #424

### DIFF
--- a/app/styles/components/_detailview.scss
+++ b/app/styles/components/_detailview.scss
@@ -1,7 +1,7 @@
 //course, session, etc detail view
 .detail-view {
   background-color: $background-grey;
-    border-style: solid;
+  border-style: solid;
   border-color: $header-grey;
   border-width: 0 2px 2px;
   border-top: 0px;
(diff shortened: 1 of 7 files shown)


app/styles/components/_detailview.scss
```diff
@@ -1,13 +1,59 @@
...
+          @include span-columns(2 of 12);
+          @include omega();
+        }
+        .calendar-toggle {
+          @include span-columns(2 of 12);
+          @include omega();
+          margin-top: 1em;
+        }
+        .detail-actions{
+          @include shift(6.5 of 12);
```
jrjohnson: Does neat support a non whole number value here (6.5).  I guess its working so yes, but it might be a little bit iffy.

saschaben:
this looks brilliant.

app/styles/components/_detailview.scss
thecoolestguy: I think I was reading somewhere that it did, but it might have been something regarding bootstrap, so I'll try to find the article to make sure. But I do know it worked ALOT better when I added the .5 to the value...

PR #503, opened 2015-04-30
styled header row of sessions section, fixes #310

### LINKED ISSUE #310: Remove extra header showing up in Sessions section of Course Detail Page
Currently seeing this:

screen shot 2015-03-16 at 12 14 14 pm

Should look like this:

screen shot 2015-03-16 at 12 09 56 pm

thecoolestguy:


thecoolestguy:
I added the gray background color to the sesions table header to match the mock, and generically applied it to all table-based header rows... Please check all html table-based tables to make sure this did not mess anything up!

saschaben:
First lookover looks great.

app/styles/components/_detailview.scss
```diff
@@ -74,7 +74,30 @@
       }
     }
 
-    /* put the default detail-block settings here */
+      /*because it is the only section with filter-tools within the header,
+        special handling/tweaking is required for the sessions detail section specifically */
+      &#sessions {
```
jrjohnson: #sessions is sufficient here, but we should avoid using id based selectors.  Also its likely we will use filter tools in the header for the program year list on the single program page.  We can cross the bridge when we come to it, but using a class based selector here would make it easier.

thecoolestguy:
I added a filter-detail-block class for use in detail blocks that have filters and then I used the class for styling...

jrjohnson:
@saschaben - give this a final check please to ensure its still awesome.

PR #506, opened 2015-05-01
fixed width on my courses toggle, dual-state toggle now blue/green

### LINKED ISSUE #505: My Courses toggle size is too small
The 'my courses' toggle switch is not wide enough.

thecoolestguy:
fixes #505


saschaben:
back in shape.

PR #531, opened 2015-05-08
Create learner group detail page

### LINKED ISSUE #460: Create Learner Groups Detail Page
learner-groups-edit-v9

jrjohnson:
From here you can move a learner anywhere in the tree.  

This is a first draft with some rough edges.  Functionality should be 100% working though.

Fixes #460


saschaben:
Memebership labels aren't displaying accurately (or memebrship listings are borked):

1. anatomy lab prime shows 21 members and 18 subgroups

ilios 2015-05-08 08-43-05

I go into AL Prime, and switch Shirley from Prime (ie unassigned to a subgroup) to Anatomy lab 1.

ilios 2015-05-08 08-44-53

After save, Shirley is still a part of AL Prime, even though she now lists with the count of anatomy lab 1:
ilios 2015-05-08 08-46-14


ilios 2015-05-08 08-48-07

ilios 2015-05-08 08-49-51

ilios 2015-05-08 08-51-10


I can't tell if users are being moved in the tree, partially assigned in the tree, replicated in the tree.

If a user is in the master group, unassigned to a subgroup, they should display in the master group.
If a user is in a subgroup, they should display in that subgroup only, not as part of the master group.
If a user is in a subgroup, they should display only membership in that subgroup on the  inline control.

Is that correct?

jrjohnson:
ok - @saschaben give this another look

saschaben:
Cool. First pass it seems that users traversing the tree all function correctly now.

PR #532, opened 2015-05-08
left-aligned first column header and left-padded, fixes #527

### LINKED ISSUE #527: Move Program Title to the Left
I noticed that the Program Title column header should line with ... the Program Title.  I hope this is not a dupe.  It may be included in "general styling".  If so, I apologize and will close as dupe.

program1

thecoolestguy:


app/styles/components/_global.scss
```diff
@@ -175,6 +175,11 @@ textarea {
 /* handle the horizontal alignment ('text-align') properties */
 .text-left {
     text-align: left;
+
+    /* for left-aligned items in the first column of a table, add padding */
+    &.first-col {
```
jrjohnson: Can we do this with the :first-of-type selector instead of a special class? Just so we don't have to remember to apply it?

thecoolestguy: yes. I will fix that. good call.

PR #533, opened 2015-05-08
adjusted program years form styling, changed CSS for general form styling overall

thecoolestguy:


saschaben:
nice!

not sure about the button alignment on this form -- it seems to float a little, not sure if that is due to the year selector being very long, or what. but we might want to revisit later.

Also a functional note: program years should auto-clone all information from the preceding year. I will add a new ticket to make sure this is addressed.

jrjohnson:
@thecoolestguy - looks like the session overview special equipment test is now failing.

PR #540, opened 2015-05-11
Reorganize publication check

### LINKED ISSUE #496: "session type" list does not render in "Review Missing Items" mode
Per notes added to #488 --

When going to the "Review Missing Items" view of a session, the session type list is not getting loaded in:

lab1

now, pick "review _n_ missing items" from the publish control:

lab2

...and note that the session type has been de-selected. In fact, there are no session types to select!:

lab3

Doesn't appear to affect the data, just doesn't show up in this view. If you return to the normal session view, the session type is still selected correctly.

jrjohnson:
Made the course and session publication check nested routes from course so they use the same overview section.  Also makes moving between these areas more seamless. 

Fixes #496 - session type can now be changed from publication check.


saschaben:
This seems to be fine; we will need to review minor styling tweaks for visual consistency (location and placement of the "back to" links) but this is pretty good as is.

PR #548, opened 2015-05-12
Better transitions for objective manager

### LINKED ISSUE #371: Add New MeSH Term to Course (when one already exists) - Confused By Process
I am a bit confused and will attempt to document what I am thinking.  There may be a pull request pending for this issue and if so, I apologize.

I see this ...
addmesh

I would like to add an additional MeSH the course objective (indicated by the squiggly blank arrow - my specialty) -- I click on the word "pharmacokinetics" and the following screen appears.
addmesh2

I now see there are two MeSH terms -- that is because these are Course level MeSH terms not Course Objective level MeSH terms.  I have to scroll up to find the following ...
addmesh3

So I know that I am adding a MeSH term to an objective so that is where I would add one.  How do I know with which of the 23 Objectives I am associated this MeSH term?  At a minimum, if someone clicks to add a MeSH term to a Course Objective, should not the user be routed to the correct location to take care of this action?  The user should not have to figure out which location is for Course MeSH terms and which location is for Course Objectives MeSH terms.  I know I got confused.

Also, there should be a text element (label or whatever) indicating to the user to which Objective is the one to attach the MeSH term they are trying to find..

### LINKED ISSUE #539: Parent Objectives should slide in from the right
Originally reported in #222

jrjohnson:
Fixed the broken horizontal slide in of objectives
Used a scroll animation to move in and out of the manger in the right place

Fixes #539
Fixes #371

### DIFF
--- a/app/components/detail-objectives.js
+++ b/app/components/detail-objectives.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import scrollTo from '../utils/scroll-to';
 
 export default Ember.Component.extend({
   store: Ember.inject.service(),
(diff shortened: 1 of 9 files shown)


thecoolestguy:
@jrjohnson, this needs a rebase...

saschaben:
functional excellent, but induces vertigo.

Can we look at perhaps adding some transitional visual to reduce potential seizures and nausea?

PR #567, opened 2015-05-13
Allow choosing an instructor (or instructor group) for an ILM session

### LINKED ISSUE #500: Add Instructors To Independent Learning Sessions
**final decision**:
*Remove associated groups when ILM session is active.  Put an Instructors manager in the content below the Learner Groups manager for ILM.*

There needs to be a way to add Instructors to Independent Learning Sessions (ILM's) similar to the way we will be adding Learner Groups to ILM's.

In Ilios 2.4.x ...

yeastbeer
yeastbeer2
yeastbeer3

jrjohnson:
Fixes #500

### DIFF
--- a/app/components/detail-instructorgroups-list.js
+++ b/app/components/detail-instructorgroups-list.js
@@ -1,0 +1,7 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  instructorGroups: [],
+  sortBy: ['title'],
+  sortedInstructorGroups: Ember.computed.sort('instructorGroups', 'sortBy'),
+});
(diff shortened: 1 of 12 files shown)


saschaben:
OK. but I think this ticket is mis-identified. Why is this for choosing instructor groups, rather than instructors?

instructor groups are the exception, not the rule. Selection & assignment should be for instructors, as well as for groups --- but primarily instructors (solo).

jrjohnson:
Wow - big blind spot on my part.  I just tweedledeed my way through that one huh?  I'm going to close and modify this to pick instructors (with groups as an option).

PR #573, opened 2015-05-15
Lock the version of i18n

jrjohnson:
The latest update uses some asynchronousity that breaks all
translatable tests.

If tests pass this should be merged, it doesn’t change any application behavior.


saschaben:
if the tests pass _except_ for IE 11, should we still merge? Cuz that's where we're at now....

jrjohnson:
I update this pr to make IE11 optional as well so now all tests should pass.  This can be merged anytime we don't need to wait on tests again.

PR #574, opened 2015-05-17
Added Program Year Steward display and manager

### LINKED ISSUE #512: Create Program Year Stewarding Department / School Control
#511 is missing this control.  Mock in #464

jrjohnson:
Follows the competency tree pattern

Fixes #512


saschaben:
first pass looks pretty darn good. If there is a way to provide some styling to alert users that a school-level entity has been selected -- italic, maybe -- we should probably make that distinction in the pick list.
Also, the ability to collpase the selected list of depts for a school so that it doesn't take up a mile of page.

If we can do either quickly and easily, let's incorporate prior to merge. If they are bigger than breadboxes, please add new enhancement tickets for both.

jrjohnson:
Collapsable is #577 its a big feature
Differentiate Schools is #578 - this is probably not a lot of work, but needs a design from @jenzweben so we should merge this first.

Issue #585, opened 2015-05-18
Fix stewards test path


jrjohnson:
The path to program years changed in #576 so #574 broke our build.  Fix the path in [stewards-test.js](https://github.com/ilios/frontend/blob/master/tests/acceptance/program/programyear/stewards-test.js#L10)

PR #588, opened 2015-05-18
542 element styling

thecoolestguy:
This *should* have fixed all problems described in the issue and not break the css linter.

### DIFF
--- a/app/styles/components/_detailview.scss
+++ b/app/styles/components/_detailview.scss
@@ -1,4 +1,4 @@
-$colapse-control-shadow-grey: #e2e2e2;
+$collapse-control-shadow-grey: #e2e2e2;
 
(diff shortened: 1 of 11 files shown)


thecoolestguy:
I thought that the mesh terms selection was broken, but it looks like it is working after I merged in from master, so that's good!

dartajax:
To test responsiveness ...

Upon resizing the window the title disappears on Windows >> Chrome.

jrjohnson:
Code looks good.  Tests are failing due to #585 so this can be merged as if it is passes review.
As long as this doesn't cause any new problems I think we should merge ASAP so we can sort out additional style issues in more organized single tickets instead of this large bucket.

@dartajax - This seems to be an existing problem.  The title is white on grey and disappears.  Can you file a new ticket for that and we can fix separately?

Issue #570, opened 2015-05-14
Confirmation Msg After Publish / UnPublish


dartajax:
It would be nice if there could be a message to advise the user that a Course or Session (or other curricular object such as Program) has been successfully Published (or UnPublished).  

Per user feedback, it seemed really easy to use but it wasn't clear if the change "went through" since there is no longer a Save button to do this.  A message of some sort would be helpful.

saschaben:
perhaps we could consider implementing something similar to what Slack is actually doing, in the edits for account preferences: the saved icon flashes briefly upon the change: https://www.dropbox.com/s/c294xwduk1rlf9l/screencast%202015-05-14%2021-44-32.mp4?dl=0

jrjohnson:
We need notifications across the board.  I created #572 to document this.

PR #592, opened 2015-05-19
Add flash messages to course publication

### LINKED ISSUE #572: Add Notification Library (and use it)
There are several places where a notification messages would be great.  Adding new sessions, publishing a course, saving topics all come to mind.  We need to add a generic notification library and then make it available to use everywhere.

jrjohnson:
Flash messages on course publiction and session publication.

Fixes #572 #570

### DIFF
--- a/app/components/course-header.js
+++ b/app/components/course-header.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
   currentUser: Ember.inject.service(),
+  flashMessages: Ember.inject.service(),
   store: Ember.inject.service(),
   course: null,
   editable: true,
@@ -47,7 +48,7 @@ export default Ember.Component.extend({
   actions: {
     unpublish: function(){
       var course = this.get('course');
-      course.get('publishEvent').then(function(publishEvent){
+      course.get('publishEvent').then(publishEvent => {
         course.set('publishedAsTbd', false);
         course.set('publishEvent', null);
         course.save();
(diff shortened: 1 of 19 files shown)


saschaben:
While I like this, I have a couple concerns:

1. providing the response header generically at the top of the page for both course and session might not be effective, since especially for sessions, but even for courses, users will often have scrolled down from the top of the screen. I think it would be much more effective to have the response header appear more contextually near the action control that has triggered it: at the top of the course header for course, and at the top of the session header for session.

2. the flash in of the header is fine (although there seems to be some delay for session); the flash-out might be better with some sort of fade, to ease the abruptness of its disappearance.

@thecoolestguy @jrjohnson are their technical obstacles to this? @jenzweben what is your take? @dartajax what do you think the user response would be to the current PR?

jrjohnson:
The location of the messages is static.  I'm guessing we could probably affix it to the top of the screen though so wherever you are the message would appear at the top of your window.  In that case we would likely want to make it more transparent to give it an ephemeral feel and probably change the default disappear from 5 seconds to 2-3 seconds.

Looking at the code you're right the delay for sessions is due to the delay in saving the session.  You don't get the message until we get an OK response back from the server.  I messed up courses and just sent the message right away.  I will fix that.

Smoother removal of the messages probably requires work on liquid-fire. Through we can try something with a spacer to make it a bit smoother.

saschaben:
One more thing: the  screen jump caused by adding in the flash header: If we are going to provide this sort of feedback using this sort of model, we should look at using existing real estate on page (or creating a space); if we can avoid causing the jump, we make for a better user experience.

jrjohnson:
@saschaben - see how this looks to you now.  I fixed the messages to the top of the screen.  Keep in mind that @jenzweben will be able to tweak the color and opacity as well to get it dialed in.

saschaben:
much better. should probably have an opaque message, with some sort of fade or transition in/out, for clarity and usability.

saschaben:
FYI, I'm having real trouble loading this one locally, even with nomless/nommore. getting this:

err1

the result is that no links work in the ilios menu, and no page can load by url without throwing an error:

err2

jrjohnson:
Two issues you are having.  
1) the deprecations which can be ignored.
2) /course/595 is no longer valid.  its now /course**s**/595  we changes everything to be plural in #576

saschaben:
Ok. But the primary issue is that the menu links fail to function on click. Something has disabled the menu link functionality.

saschaben:
I"ve posted some screencaptures to show you the behaviors with just loading, and loading after nomless/nommore. also a comparative capture of working with master...

https://www.dropbox.com/sh/90ah4g1sdt9fm8l/AADYN7Ij72Nm036DgaDNBbMya?dl=0

jrjohnson:
@saschaben  Ok - fixed it.  z-index was hiding the link so you could not click it.  I have no idea why it was working on my machine.  I tried to write a test, but it was too hard.

saschaben:
still no z-index love here.  after cache clearing and deletion & reload of PR, same behavior.

thecoolestguy:
Yeah, for the record.  I can't click on any of the menu links either... I'll see if I can take a stab at the fix.

dartajax:
I am looking forward to checking this out.

thecoolestguy:
So, changing the z-index value from '200' to '201' on line 58 in _navigation.scss seems to do the trick, but  I don't know if it is breaking anything else.  It doesn't seem to be... @jrjohnson, do you want to make this change and see how it behaves?

jrjohnson:
Fixed z-index on 55 thanks @thecoolestguy .  This fixes #616

saschaben:
success!

@jenzweben take a look at tweaking opacity and size. We should also look for a better transition in and out. And I don't think it needs to be persistent for the length of time it currently has. Maybe half that.

If we could mimic the alert pane that is used in a similar model for waffle.io, that might be a good start for an example....

jrjohnson:
Transition out requires a change to the flash message library.  I submitted a PR for that, awaiting feedback from the author or merge.

The length is completely configurable.  I think we should merge before having @jenzweben do her style thing.  Its easier to manage that way.

saschaben:
OK. will merge after code review and then pass to @jenzweben for refinements. now go to sleep!

PR #599, opened 2015-05-20
Only display the manager title not the section or count

### LINKED ISSUE #521: Header for Manage Blocks
We are using the same header on all kinds of managers.  Topics is a good example:
screen shot 2015-05-07 at 11 22 26 am

We should do something better to separate the Topics (11) Manage Topics.  @jenzweben can you mock something up?

jrjohnson:
Changes cohorts, lms, objectives, topics, program year competencies
components so that when the manager is opened ‘Manage Topics’ is
displayed and not Topics (9) Manage Topics

Fixes #521


saschaben:
much cleaner!

PR #600, opened 2015-05-20
Add instructor-selection-manager to offerings

### LINKED ISSUE #568: improve pattern for selection of instructors and instructor groups
The current display is confusing and poorly prioritized.
instructors

in 2.4, we avoided this by concatenating the select lists of available groups and individuals, with groups simply listed first as selectable entities. @jrjohnson has noted that this is pretty tricky in the new model, so we have split the choosers. But right now presents a weird selection that implies the user should primarily be selecting instructor groups, which is incorrect. the default use case is a single instructor for a single offering/ILM.

Having the instructor group panel at top, always open, and taking up as much real estate on its own as it does is probably not a good call. We want the instructor (solo) selector to be primary here.

@jenzweben @jrjohnson can we get a revised mockup and implementation pattern for this, for both the above selection for an offering, as well as selection for an ILM...? Perhaps a version that brings the implemented/implementable code closer in line with our original mocks?

jrjohnson:
Combines the search for instructors and groups.

Fixes #568

### DIFF
--- a/app/components/detail-instructors.js
+++ b/app/components/detail-instructors.js
@@ -39,7 +39,6 @@ export default Ember.Component.extend({
           return group.save();
         }));
 
-        promises.pushObject(group.save());
       });
       this.get('instructorGroupBuffer').forEach(function(group){
         promises.pushObject(group.get('ilmSessions').then(ilmSessions => {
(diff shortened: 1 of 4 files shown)


saschaben:
very slick!!!

PR #606, opened 2015-05-20
Select field dropdown arrow resized globally and learner group membership aligned

### LINKED ISSUE #580: Learner Groups Member Re-assignment needs to look like a dropdown
Eventhough we might not technically be able to have a dropdown, we need to make the selector look like a dropdown - also the check & "x" should be to the right of the dropdown 

screen shot 2015-05-18 at 10 19 04 am

Should look like:
screen shot 2015-05-18 at 10 20 52 am

thecoolestguy:
fixes #580

### DIFF
--- a/app/styles/components/_editinplace.scss
+++ b/app/styles/components/_editinplace.scss
@@ -11,8 +11,7 @@
   }
 
   .editor {
-    input,
-    select {
+    input {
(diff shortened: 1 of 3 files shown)


saschaben:
this looks fine.

PR #609, opened 2015-05-21
Improved user search

### LINKED ISSUE #448: User Search doesn't update when an item is clicked
When a user is clicked in the user search results list (find directors for example) the choice isn't reflected by greying out the item.

jrjohnson:
Response to search action from search-box
Clear terms when something is picked
Sort results in the search action, not separately.

Applies to director search, ilm session instructor search, and offering instructor search.

Fixes #448

### DIFF
--- a/app/components/instructor-selection-manager.js
+++ b/app/components/instructor-selection-manager.js
@@ -1,7 +1,6 @@
 import Ember from 'ember';
 
 export default Ember.Component.extend(Ember.I18n.TranslateableProperties, {
-  subject: null,
   instructors: [],
   availableInstructorGroups: [],
   sortInstructorsBy: ['title'],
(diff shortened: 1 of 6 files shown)


saschaben:
I'm a little concerned that we now have two very different use patterns for selecting faculty (course director search vs. ILM instructor search/offering instructor search) which require the use to engage in two contrary patterns of behavior to do what they see as essentially the same thing. In ILM, I kept expecting my first selected instructor to be greyed out in the next search, but it wasn't -- because I hadn't saved. I see a lot of confusion along those sorts of lines being engendered here unless we can dial this in and have all searches follow the same pattern, and provide the same visual/cognitive results.

jrjohnson:
@saschaben - what you were seeing was a bug in the way instructors were getting passed in.  I believe this is now corrected.   Give it another look.

saschaben:
bug is corrected. this looks good. next step is to apply the manager pattern to all faculty search areas (and to figure out the best display style for the faculty lists)

PR #613, opened 2015-05-21
fixed course list hover color and tweaked sessions list hover color

jenzweben:


saschaben:
much better!

dartajax:
Per @thecoolestguy, this one is ready to merge so I am going to do it since .css won't cause the tests to fail.

Issue #458, opened 2015-04-18
Add WYSIWYG editor option to inplace-textarea / big-text components


jrjohnson:
Currently used for editing objectives we need to be able to enable a nice text editor in these fields.

saschaben:
perhaps something like https://www.npmjs.com/package/ember-cli-medium-editor (which includes a "pasteAsPlainText" config value...)?

jrjohnson:
This isn't the best addon, also the medium editor doesn't support lists, which I think is our number 1 use case for not going with plain text.

@saschaben - what functionality do we actually need in objective titles?

saschaben:
We need to provide at the very least: font styling (bold/ital/underline/etc); ordered/unordered lists; font sizing; The minimum functional requirements for format can be determined by looking at the existing 2.4.X editor toolbar -- and then adding the ability to directly import externally sourced/formatted cut&paste text in a clean format. please keep in mind that users are NOT going to learn markdown, any more than they know HTML. Controls must be WYSIWYG, and providing toolbar control over formatting is the most commonly understood way for them to interact with text. The markdown route makes good sense, but we still need to remember that our audience is really not sophisticated about this stuff.

I think we can approach the toolbar functions in phases --- we won't get too much grumbling if we can provide text styling (bold/ital/underscore) text sizing and lists to start, and then iterate over the other features. "Clean" cut&paste import is non-negotiable.

jrjohnson:
Ok - if we can drop support for lists then we will be in better shape.  Cut and paste is workable in that extraneous junk will get stripped out.  However we won't always correctly interpret what the user has entered.  So they will loose styling for size of characters, but may keep bold.  I'm not finding any editor that can do better than that.

saschaben:
we can't drop support for lists.

jrjohnson:
Ok - doesn't matter.  It runs out the medium editor can be modified to allow ordered and unordered lists.  So I think this is the right path for us.  I will put something together so we can demo it.

jrjohnson:
Notes from #619 
**UPDATE 5/26:**
good to go with medium editor. Issues to address are:
1. when you click to manage it needs to open the editor immediately, rather than requiring two clicks to open the editor.
2. we need to provide a plain text editor option along with WYSIWYG
3. Objective display needs to show styling
4. we need to find an appropriate means of scrubbing existing extraneous code tags from the objective data, along with our cut&paste solution to scrubbing new input data.

**Solutions**
[...]

PR #619, opened 2015-05-24
Add a Medium Editor to text area elements

jrjohnson:
**UPDATE 5/26:**
good to go with medium editor. Issues to address are:
1. when you click to manage it needs to open the editor immediately, rather than requiring two clicks to open the editor.
2. we need to provide a plain text editor option along with WYSIWYG
3. Objective display needs to show styling
4. we need to find an appropriate means of scrubbing existing extraneous code tags from the objective data, along with our cut&paste solution to scrubbing new input data.

**Solutions**
- determine a tag whitelist for accepted markup
- find a means of both scrubbing and validating the scrub

QUESTIONS:
can we do this without users having to look and approve?
can we add a column to track cleansing of data?

*Try out the medium editor when changing objective titles.*

*Thoughts?*

Refs #458

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -3,7 +3,8 @@
     "document",
     "window",
     "-Promise",
-    "$"
+    "$",
+    "MediumEditor"
   ],
   "browser": true,
   "boss": true,
--- a/app/components/inplace-textarea.js
+++ b/app/components/inplace-textarea.js
@@ -3,4 +3,11 @@ import InPlace from 'ilios/mixins/inplace';
 
 export default Ember.Component.extend(InPlace, {
   classNames: ['editinplace', 'inplace-textarea'],
+  options: '{' +
+    '"autoLink": true,' +
+    '"forcePlainText": false,' +
+    '"cleanPastedHTML": true,' +
+    '"imageDragging": false,' +
+    '"buttons": ["unorderedlist", "orderedlist","bold", "italic", "underline", "anchor", "header1", "header2", "quote"]' +
+  '}',
 });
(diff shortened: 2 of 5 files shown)


saschaben:
Hmmm. first thoughts:

1. the box around the text doesn't appear without a second click --- a little hard to understand what's going on. If we can, we should probably set this so that the box border shows up on the first editing click in the text.
2. the controls are applying the HTML, but the tags aren't being displayed. hard to judge because of this.
3. the inability to control font-family is annoying for the existing cut and paste cruft that exists. It would be great if we could somehow force this type of stuff into conformity....
4.. if we can tweak things a little per above, then I think that with the current control ribbon we should be OK for an initial release.

Overall: YAY an editor!!!!!

jrjohnson:
Yes, I think we're going to have to contribute a fair amount of code or else create our own addon to get this to play well with Ember.   I want to make sure we're on the right path before digging into that.  The essential thing to take away is the 'Medium Editor' toolbar.  Which appears over highlighted words as opposed to at the top of the editing area.  Does this make sense as a pattern?

The other issue here for me is that after you save you can only see the plain text.  I think in order to use this kind of editor which hides the HTML we're really going to need to display the formatted text on the list of objectives.

So lots to discuss, but maybe we're on the right track.

saschaben:
Yeah.

Lots of good links to discussion if you Google "ember-cli wysiwyg editor". I'm open to radical choices here if they answer our needs

jrjohnson:
Closing - @ebryn will look to see what the best solution is.

PR #625, opened 2015-05-27
Fix css conflict between header and editor actions

### LINKED ISSUE #401: Course Title Editing Box
The check & undo buttons should be on the RIGHT of the inline editing box and the box should be as wide as possible...

screen shot 2015-03-30 at 2 45 12 pm

jrjohnson:
This fixes the mis-alignment of the save buttons for editing course and
session titles.

Fixes #401


saschaben:
functional test passes.

PR #626, opened 2015-05-27
When a session-details component is loaded scroll to the top

### LINKED ISSUE #403: View / Edit Session: scroll-to session placement in list is mis-aligned
When a user clicks on a session that is in the lower part of the session list, the Session details page needs to be auto-scrolled down so that the user can see the Overview section currently, the Session details page is transitioning in at the middle...so users see this

screen shot 2015-03-30 at 2 55 31 pm

and they should see this:

screen shot 2015-03-30 at 2 59 48 pm

jrjohnson:
The fixes #403 so that when you follow a link to a session from a course you will scroll to the session header and not be stranded in the middle of the session details.

### DIFF
--- a/app/components/session-details.js
+++ b/app/components/session-details.js
@@ -1,9 +1,13 @@
 import Ember from 'ember';
+import scrollTo from '../utils/scroll-to';
 
 export default Ember.Component.extend({
   availableTopics: [],
   sessionTypes: [],
   session: null,
+  didInsertElement: function(){
+    scrollTo("#session-" + this.get('session.id'));
+  },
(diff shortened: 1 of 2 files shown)


saschaben:
Much better. Question: do we want to scroll to the current mark (top of title header), or up to just above the "back to session list" link?

thoughts?

jrjohnson:
Its no problem either way - @jenzweben do you care where on the page we scroll when selecting a session?

saschaben:
I would vote for a target like this:

ilios 2015-05-27 11-07-51

PR #628, opened 2015-05-28
Added a credit section to the README

jrjohnson:
Browserstack asked me to link to their service in order to gain access to their open source services.  It occurs to me that we should be thanking all of the companies that give us stuff for free. 
Added links for Travis, Sauce, Heroku, and BrowserStack
Did I miss anyone?

Any objections to including this in our README?


saschaben:
I think that is great. Can't think of other freebies offhand.

jrjohnson:
Well we both forgot github... so yea...

saschaben:
actually, I started to type it in, but realized we DO pay for it. so...

I'm fine with adding them, but if we are only doing free stuff, then they are sort of an extra.

jrjohnson:
I'm going to leave github for now, they give us lots of free stuff, even if we do pay for some things.

PR #637, opened 2015-06-01
added icon to learner groups header in place of ascii arrow

jenzweben:


saschaben:
I think this needs some revision -- the icon only appears in the first tier of the name, second tier still shows ascii:

ilios 2015-06-01 10-58-32

jenzweben:
The second "->" is being added from learner-group.js and I don't know how to make an icon display in a .js file...I'll make a ticket for this

jrjohnson:
Closing in favor of #664

PR #639, opened 2015-06-01
Sessionhover2

jenzweben:


saschaben:
not sure I'm loving this, stylistically. let's continue iterating as discussed and get team feedback tomorrow.

PR #640, opened 2015-06-02
waffle.io Badge

waffle-iron:
Merge this to receive a badge indicating the number of issues in the ready column on your waffle.io board at https://waffle.io/ilios/frontend

This was requested by a real person (user dartajax) on waffle.io, we're not trying to spam you.


jrjohnson:
I'm sort of OK with adding a badge, but it should be in the correct place in the readme.  Not just stuck on top.  @dartajax if you want to add this you will need to create a PR from your own fork since we can't update the waffle one.

PR #648, opened 2015-06-04
Simplified button styles

### LINKED ISSUE #316: hover states for inline edit buttons

jrjohnson:
Removed all the ‘add’ classes from detail buttons - that wasn’t really
doing anything.  Added hover state to inplace edit buttons.  Changed
the inline edit buttons from .save to .done which lets us reuse the
style.

I believe this Fixes #316

### DIFF
--- a/app/styles/components/_editinplace.scss
+++ b/app/styles/components/_editinplace.scss
@@ -27,7 +27,6 @@
       display: inline;
 
       button {
-        background: transparent;
         margin: 0;
         padding: 0 .2em;
       }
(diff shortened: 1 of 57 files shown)


dartajax:
I noticed that the Add New button on Instructor Groups and Student Groups moved over to the left by about an inch on my screen.

dartajax:
And then I found this too ...

what

dartajax:
This is what it looks like currently ...

what1

thecoolestguy:
Code looks good.  Needs final review and merge.

saschaben:
this looks good, seems to check out.

PR #650, opened 2015-06-04
Clear the new session list when navigating

### LINKED ISSUE #643: "session saved" message is too persistent
the message lasts until logout. It should clear after either a page change or a new save action for any issue. 
ilios 2015-06-02 14-10-14

jrjohnson:
I also updated the default settings for flash messages.

Fixes #643


saschaben:
looks good. merge.

PR #652, opened 2015-06-04
Sessionhover3

jenzweben:


saschaben:
without a cursor change for hover (clickable vs nonclickable), this still feels wrong.

-if we have the double highlight, the block highlight must be clickable to activate the link.
-the cursor must be an arrow vs a pointer hand when hovering over non-clickable areas.
-I still don't love the heaviness of the highlight on top of highlight on top of color background. It serves the purpose, but runs counter to the esthetic of the application. I would really like to see a solution with the cursor varying properly, both with and without the secondary highlight, before we make any decisions on this. @thecoolestguy @jenzweben let's see what we can do....

jenzweben:
OK I updated this - removed hover bg color for title & eyeball and added cursor pointer only for those two columns

saschaben:
@jenzweben I'm not seeing the updates -- I don't think PR code is getting your changes posted in.

PR #656, opened 2015-06-04
Sessionhover5

jenzweben:
OK this should be correct - removed hover bg color - hopefully it works

### DIFF
--- a/app/styles/components/_resultslist.scss
+++ b/app/styles/components/_resultslist.scss
@@ -36,6 +36,7 @@
         background-color: rgba(239, 251, 255, .8);
       }
     }
+     
 
     thead tr {
       background-color: $header-grey;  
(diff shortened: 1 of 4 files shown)


saschaben:
This looks MUCH better, and is working! A couple minor code issues that we may want to address later (like the td hover of the title not being active for the link, but still switching the cursor), but basically I think we are good here. One question is how we all (and by we all I mean @jenzweben) feel about the now significantly different hover states between our lists for course, and program, and program year, and this new one for session. :+1: ? :-1: ?

jenzweben:
I think that the other hovers are good as/is - using bg color highlight is pretty standard, expected behavior.  The only reason we're going with borders here is because Sessions has a background color already applied...

I'm open to using borders everywhere but I think it would probably be less usable.

PR #660, opened 2015-06-05
Added Readme Badges

jrjohnson:
Added the badges for slack and waffle.io


saschaben:
totally down with this.

Issue #446, opened 2015-04-10
Click to Edit Field Activates Field and sets Focus on Selected Field


dartajax:
It would be nice if, when a user wishes to add an "External ID" (for example) to a Course, the field chosen to modify ("External ID" in this case) would not only activate and show the save and cancel inline edit buttons but would automatically set focus on the text entry field so the user can start typing without having to use the mouse to set focus on the field.

dartajax:
This should be applied to all fields so from a user perspective, it is easier to know what is being modified and to be taken there automatically.

dartajax:
Another Example 

Programs 
* Select a Program 
* Click Program Title (Short)'

The field becomes activated but the focus is not set on the field - user still has to click in to edit.

PR #661, opened 2015-06-05
Focus on the search box when the icon is clicked

### LINKED ISSUE #494: Search Box Icon Select-ability
During usability testing, Gitanjali was trying to search for a Director to add and clicked on the search icon.  She was then unable to type in the search field because she thought she had already selected it.

Sooo.  We need to implement the search field in a way that puts the cursor in the input area when they click on either the search icon OR the search box input area...

jrjohnson:
Fixes #494

### DIFF
--- a/app/components/search-box.js
+++ b/app/components/search-box.js
@@ -15,6 +15,8 @@ export default Ember.Component.extend({
       this.sendAction('clear');
     },
     search: function(){
+      //place focus into the search box when search icon is clicked
+      this.$('input[type="search"]').focus();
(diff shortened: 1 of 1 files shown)


saschaben:
works for me!

dartajax:
This is only for search boxes I take it.  Should we have or do we already have a ticket for the setting focus on inline editing fields automatically?

jrjohnson:
@dartajax -yes search box only.  If there isn't a ticket for inline edit already please create one.

dartajax:
The inline edit ticket is #446.

PR #662, opened 2015-06-05
Change learner group subgroup member title based on context

### LINKED ISSUE #647: mislabeled section in learner group display
group1

This label needs to be context-driven. If viewing a top level group, we should refer to group (or "this level"). If viewing a sub-group (non-top level), it should refer to sub-group (or small group, or "this level").

jrjohnson:
Fixes #647

### DIFF
--- a/app/models/learner-group.js
+++ b/app/models/learner-group.js
@@ -215,4 +215,5 @@ export default DS.Model.extend({
       promise: promise
     });
   }.property('parent', 'parent.topLevelGroup'),
+  isTopLevelGroup: Ember.computed.empty('parent.content'),
 });
--- a/app/templates/components/learnergroup-details.hbs
+++ b/app/templates/components/learnergroup-details.hbs
@@ -8,7 +8,11 @@
 
     <section class='detail-block toplevelgroupmembers'>
       <div class='detail-title'>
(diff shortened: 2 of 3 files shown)


saschaben:
copying note from #647:

I just had a brainstorm: this label is better but still not really accurate from the POV of a user. If I navigate to Anatomy Lab, I should see that there are 26 sub-groups that contain all the learners (and ideally no one just floating in the top level group). But all those learners are "in" Anatomy Lab prime, since all their respective sub-groups are "in" Anatomy Lab prime. So rather than saying "Anatomy Lab Members NOT in this Group" at the top level (and we are ONLY talking here about top level), which is wrong on a number of levels -- members can't be non-members, and they are all in the group as noted above --why don't we say "Anatomy Lab Subgroup Assignments"? That's not only more accurate, but actually describes what we are really showing in that section....thoughts?

saschaben:
...or "[Top Level Group Name] Learner Assignments"...?

jrjohnson:
I can do either of this in a matter of seconds.  My vote is Anatomy Lab Learner Assignments.  Or even Anatomy Lab Learner Subgroup Assignments.

saschaben:
Learner Assignments it is. users always hated "subgroup". less of that is better.

saschaben:
Mergie strikes!

PR #663, opened 2015-06-05
Improve hover for results lists

### LINKED ISSUE #607: Clicking Anywhere in Session List Row Should Take User to Edit That Session
It seems with the mouse pointing as a finger and the entire row highlighted that clicking anywhere in the row of a session should perform the action of selecting the selected session for modification / review.

jrjohnson:
Instead of applying a background color indicate hover by changing the
borders on the table row.  This provides focus without indicating a
click will trigger an action.

I also consolidated the session highlighting and active-background
class.

I applied this state to all the results list (courses, sessions, learner groups, programs, etc...) so give it a look and see if thats an improvement or it this should just apply to the sessions list.

Fixes #607 
Replaces #656

### DIFF
--- a/app/styles/components/_resultslist.scss
+++ b/app/styles/components/_resultslist.scss
@@ -25,15 +25,20 @@
   .resultslist-list {
     @include span-columns(12 of 12);
     @include pad();
-    tbody tr {
 
-      &:hover {
-        background-color: rgba(239, 251, 255, .8);
+    tbody tr {
+      border-bottom: 1px solid transparent;
+      border-top: 1px solid transparent;
+
+      &:hover,
+      &:hover td {
+        border-bottom: 1px solid $hover-border;
+        border-top: 1px solid $hover-border;
       }
 
-      &.active-background,
-      &.active-background:hover {
-        background-color: rgba(239, 251, 255, .8);
+      &.donthover {
+        border-bottom: 0;
+        border-top: 0;
       }
     }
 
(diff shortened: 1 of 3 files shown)


jenzweben:
looks good!

saschaben:
This looks great! However, when applied to the course and program & learner group lists, it creates a weird scroll/spacing behavior that isn't at all a deal killer, just weird. Try it: go to learner groups, scroll to the bottom of the list, and then hover slowly over each record, ascending up the list. You'll see the horizontal scroll bar shift slightly, and the spacing between rows increase (and persist) so that the overall list lengthens by two pixels in height with each hover action....

check it out. bug or feature?

jrjohnson:
Hmmm - I thought the border-colapse I added solved that - I did on my machine.  But maybe that doesn't work all the time, or maybe the browser forgets to collapse the borders when you scroll around the page?  I'm thinking file this under css quirks we don't have time to fix right now.

jenzweben:
Not sure if this will help for the way you've coded stuff but I had put in borders the same color as the bg for non-hover state to ensure no jumping around

jrjohnson:
@jenzweben - I tried to copy that approach, but then the top border on the second item never seemed to show up.  It was like it never got the message to change (again might be a browser quirk).  I can go back to that approach if that slight bug is better than the current slight bug.  I have no preference.

jenzweben:
to solve that issue you have to add td to the hover style

saschaben:
totally OK to merge with this if we don't have a simple and super quick fix.

jrjohnson:
Let me try @jenzweben's fix and if it works we can go that route.

jrjohnson:
OK - take another look - @jenzweben had all the answers.

saschaben:
looks good. are we now using solid vs dotted line? I know we were having issues with dotted line rendering earlier, either is working for me....

jrjohnson:
I vote solid mainly because I don't want to change it any more.

saschaben:
is fine. If @jenzweben has concerns or thoughts let's do an enhancement ticket. move this one along, soldier.

thecoolestguy:
Yeah, dotted has notoriously caused cross-browser compatibility problems and, at 1px, it's not that big of a difference

saschaben:
ready for merge.

PR #665, opened 2015-06-05
Display Topics list as columns

### LINKED ISSUE #514: need better styling of saved topics
**final decision**:
4 columns alphabetical vertically down.  Setup the grid so they drop down when the page is too small.

_Right now the read-only topics list is unwieldy when there are more than four or five items -- take a look at a typical program year for an example._

_We need to style the list so that it is readily legible; either badge styles (like the edit format) without the "remove" x, or a more formatted list. Open to blue sky ideas here but let's get something!_

jrjohnson:
4 columns on big screens, 3 columns on small screens, one column on mobile.

It looks awesome on big lists, can we live with it when there are only 3 topics?

Fixes #514

### DIFF
--- a/app/styles/components/_detailview.scss
+++ b/app/styles/components/_detailview.scss
@@ -192,6 +192,16 @@ $collapse-control-shadow-grey: #e2e2e2;
     }
   }
 
+  .columnar-list {
+    @include media($medium-screen) {
+      @include column-count(3);
+    }
+
+    @include media($large-screen) {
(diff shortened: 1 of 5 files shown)


saschaben:
I think this looks awesome all round. merge it, Mergie McMergerson!

jrjohnson:
Sshhh... don't summon Mr. McMergerson yet.  Tests are broken.

PR #668, opened 2015-06-05
Create a program year publication check

### LINKED ISSUE #654: Program info not persistent for "review missing items" view
for a program year, if "review missing items" is selected from the publish list, the program title and duration information is not persisted on screen. Also, no "back to program year" link is available.

ilios 2015-06-04 12-02-45

shifts to this: 

ilios 2015-06-04 12-03-23

jrjohnson:
Turns out this was missing and we were using the program one.

Fixes #654


saschaben:
looks fine. we may want to consider putting in a smoother transition for this at some point, but this is ready for now.

jrjohnson:
Good catch.  I created that ticket as #674

PR #669, opened 2015-06-05
Fix display of learner group sub groups list

### LINKED ISSUE #667: learner group delete buttons require correction in styling
Text for the "yes" button doesn't appear; non-hover outline state doesn't appear for same.

image

jrjohnson:
Fixes #667


saschaben:
LGTM

Issue #396, opened 2015-03-30
Add Learning Materials Needs to be Styled


jenzweben:
This might be a work in progress but it needs styling.  Here's how it looks:

![screen shot 2015-03-30 at 1 13 27 pm](https://cloud.githubusercontent.com/assets/5023894/6905740/0e171614-d6df-11e4-8f04-0bf6d1ab610c.png)

Here's the mock:
![screen shot 2015-03-30 at 1 15 45 pm](https://cloud.githubusercontent.com/assets/5023894/6905749/155b1a88-d6df-11e4-93f1-46b76bd083ca.png)

saschaben:
when we style this, we need to remember to provide form styles for all three variants: document (shown above), link (requires inline display of URL), and citation (requires inline display of citation).

saschaben:
moving this to 0.3.0. Can be worked in conjunction with #452.

jrjohnson:
Fixed in #671

PR #671, opened 2015-06-05
Learning materials styled

thecoolestguy:
Needs acceptance tests and working file upload, refs #396

### DIFF
--- a/app/styles/components/_forms.scss
+++ b/app/styles/components/_forms.scss
@@ -122,7 +122,8 @@
 
 //put the universal (non parent-dependent) styles below here
 .form-label,
-.form-data {
+.form-data,
+.form-data-submit {
   margin-bottom: .4em;
 }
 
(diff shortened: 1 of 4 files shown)


saschaben:
looks soooooo much better!

need to make tickets for (if we haven't already);

 - adding LM file search to the Add file form;
 - adding dependency of state to actions for copyright (checkbox not checked = grey text, rationale box, inactive save; checkbox checked = black text, no rationale box, active save; checkbox not checked but rationale entered = grey text, active save)

thecoolestguy:
looks like the form is not displaying the default dropdown values

saschaben:
OK. So dropdown defaults are displaying for me, but form is not saving changes to status or user role properly -- they revert to defaults when pulling up secondary "attach to the course" form. API issues combined with the two-step form here may be at fault...?

jrjohnson:
scss-lint is failing.

PR #673, opened 2015-06-06
Separate Learner Groups by a nice icon where possible

### LINKED ISSUE #664: Use icon for learner groups header in place of ascii arrow
The learner group header currently uses -> to separate groups.  Use font awesome angle-right instead.

jrjohnson:
In some places the group list has to be a string so I used a plain >
character to separate group levels there.

Fixes #664


saschaben:
much prettier. eyeballed and passed.

PR #675, opened 2015-06-08
Code coverage analysis for code climate

### LINKED ISSUE #636: Exclude mirage directory from code climate
We're getting a bad score in code clime at least partially because the mirage fixtures and factories are getting checked and they should be skipped.

jrjohnson:
Current using the master branch of ember-cli-blanket and my fork of the
javascript-reporter until changes get merged and versioned.

Added code climate config file fixes #636


jrjohnson:
If tests pass I believe that this is ready.

PR #710, opened 2015-06-14
Enable IE10 tests

jrjohnson:
This update to ember-pikaday should allow IE10 tests to pass once again so they should no longer be marked as an allowed failure.


jrjohnson:
IE10 still not ready.

PR #715, opened 2015-06-18
Save learning material notes correctly

### LINKED ISSUE #714: Course and Session learning material notes are not saving
References https://github.com/ilios/ilios/issues/845

jrjohnson:
Fixes #714

### DIFF
--- a/app/components/detail-learning-materials.js
+++ b/app/components/detail-learning-materials.js
@@ -45,6 +45,7 @@ export default Ember.Component.extend(Ember.I18n.TranslateableProperties, {
       var buffer = Ember.Object.create();
       buffer.set('publicNotes', learningMaterial.get('publicNotes'));
       buffer.set('required', learningMaterial.get('required'));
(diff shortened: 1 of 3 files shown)


saschaben:
a thing currently failing in local fakeAPI builds, but not yet tested in API builds: when notes exist, this is not being correctly identified in the list view (label remains "no" rather than shifting to a nice green "yes"). let's make sure this gets validated with this PR

jrjohnson:
@saschaben That indicator is attached to the showNotes option.  Do we want to modify the columns to be something like:
- Show Notes
- Has Notes

and then have both indications?

saschaben:
@jrjohnson I think it better that we leave it as is for now and merely detect note existence with yes and no. We should add an enhancement to update the display to include an additional icon flag for when they are either hidden or shown (maybe along the lines of the eyeball or some such).

jrjohnson:
Further work in #720.  This can be merged.

PR #716, opened 2015-06-18
Save only one session when type changes

jrjohnson:
We were saving every session connected to a type.


saschaben:
local build throws Uncaught TypeError: session.save is not a function

thecoolestguy:
I am not able to save a session type change. Console returns: Uncaught Error: Assertion Failed: You cannot add 'undefined' records to the sessionType.sessions relationship (only 'session' allowed)

jrjohnson:
Ok - I think I fixed this.  Little bit of a race condition.  Please test again.

saschaben:
Appears to be better now. ILM still busted of course, but this granularly looks okey dokey.

PR #721, opened 2015-06-18
Changed 'add new instructor group' to use form structure 

thecoolestguy:


saschaben:
huzzah. looks great. ready to move along with code review

PR #722, opened 2015-06-19
Add a nice errors message when someone visits a URL that doesn't exist

### LINKED ISSUE #709: Add Not found page for frontend
When you visit on route that doesn't exist you get the browser unsupported error instead of a 404 page.  We need to add a route that will act as a 404 / Not Found page on the frontend.

```this.route("fourOhFour", { path: "*path"});``` will catch all requests so we just need a template that displays a nice message.

jrjohnson:
Fixes #709

@saschaben - I went with your ‘Rats!’ message.  If you were kidding I still think its a great 404 message.

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -35,6 +35,7 @@ Router.map(function() {
       this.route('publicationCheck', {path: '/publicationcheck'});
     });
   });
+  this.route('fourOhFour', { path: "*path"});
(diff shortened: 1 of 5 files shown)


saschaben:
I thought a little levity would be good. Glad to know in not alone!

PR #727, opened 2015-06-19
Fix user events so it works on the API

jrjohnson:


jrjohnson:
Tests are all passing so I'm going to go ahead and merge this one liner.

PR #728, opened 2015-06-22
Update ember and ember-data

jrjohnson:
This updates both ember and ember-data to the latest versions.  Ember 1.13 has the new glimmer engine so it is faster.  Ember-data is finally at a stable 1.x release.  

These were not insignificant updates and there are some other changes here as well.

Things to test:
- [x] Form to edit an offering (better format, but also new time-picker)
- [x] Unpublish course
- [x] Unpublish program year
- [x] Changing program year stewards
- [x] Transition animations (updated the liquid-fire transition library)
- [x] Navigation menu
- [x] Calendar display
- [x] Edit session
- [x] Edit learner subgroup

Also needs a through click around to make sure nothing weird happened.

There are still several deprecations being flagged by ember which you can see in the console or when you run ember server.  These will need to get sorted out before we can update to 2.0, but they aren't critical right now.


dartajax:
Removed screen shot from before.

PR #752, opened 2015-06-28
Authentication!

jrjohnson:
Standalone you can login with user: demo, pass: demo.

Works with the API to authenticate users by either username / password or shibboleth redirect.

Access is prevented to all un-authenticated users at any URL.  You will be redirected to the login page and then once logged in returned to your original requested path.


saschaben:
Huzzah!

saschaben:
Hmmm. local run fails with the following console error (this is after nomless):

fail1

Heroku build fails with our generic "bad browser" message and a string of 403 forbidden console errors.

saschaben:
ach. nevermind. was loading with the demo api. testing now with fake api on local.....

(but the Heroku issues stands...)

saschaben:
`username` is currently case-sensitive; should be case-insensitive.

jrjohnson:
Heroku issues appears to be an existing problem with the build configuration for pull requests attempting to load from the AWS cloud.  I will do a separate PR for that.  
If you want to test this with a backend you would need to checkout https://github.com/ilios/ilios/pull/864 first.

Is username really case insensitive?  That seems weird to me, but it is an easy fix.

saschaben:
generally, yes: username = case insensitive, pw = case sensitive. this was a battle fought in the 90s between DOS-win systems and *nix systems. while you still find the **nix case sensitive style amongst the geekerati (which admittedly around here is pretty mainstream), the rest of the world was won over by the other side.

jrjohnson:
Closed / Reopened to restart heroku build.

saschaben:
usability of auth basics checks out A-OK.

PR #754, opened 2015-06-30
Working logout function

### LINKED ISSUE #753: provide logout confirmation
we need to provide a confirmation page for when a user clicks logout:

> You have successfully logged out of Ilios. To help protect your account and the security of others, please close all browser windows.

> >*Click here* to return to Ilios.

jrjohnson:
When a user logs out they get redirected to the login page with a flash message confirming the logout.

Fixes #753

### DIFF
--- a/app/components/ilios-header.js
+++ b/app/components/ilios-header.js
@@ -1,6 +1,12 @@
 import Ember from 'ember';
 
 export default Ember.Component.extend({
+  session: null,
   currentUser: Ember.inject.service(),
   userName: Ember.computed.oneWay('currentUser.model.fullName'),
+  actions: {
+    logout(){
+      this.get('session').invalidate();
+    }
+  }
 });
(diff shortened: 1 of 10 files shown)


saschaben:
can we move the login form down on the page about 60px, so the "login" label doesn't conflict with the flash message? With that and a code review this should be fine to push.

I think this is fine for the beta; we should probably review (with JenZ) to determine a better long term display model here; the flash message is a bit short for a message like this, probably want something just a little more persistent in this particular situation.

jrjohnson:
The form is styled without anything special right now.  I would prefer to merge and leave it that way and then take some time to completely restyle the experience later. (for example error messages are a bit off).

saschaben:
That's fine, I'd still like to move it down the page 60px, as the overlay looks more like an accident right now, and less like a WIP. But if you are adamant, I'm OK. We'll just put it on the list for 0.5.0.

PR #756, opened 2015-06-30
Use the real codeclimate test reporter

jrjohnson:
This should make our coverage reporting work again.


jrjohnson:
Test are failing until #756 is merged so it should be merged first, but this is also ready for review.

PR #769, opened 2015-07-08
Update translation library

jrjohnson:
The update to ember-i18n 4.0 is a major one, but with a lot of benefits to be had.  This should be a complete replacement with no added / removed functionality.

Once this is merged we can create new locale files and add the ability for a user to select their preferred language and for an installation to select the default language.

@saschaben this is going to require a complete click through review of the application to look out for any missing translations or other lost functionality.


saschaben:
OK. Let's review and merge tomorrow, and I'll do a thorough step thru.

thecoolestguy:
I created a couple of issues regarding some small things we should fix later, but you can still merge after  you've reviewed.

PR #777, opened 2015-07-13
Add language picker to user menu

jrjohnson:
Also added a test to ensure translation files stay in sync

### DIFF
--- a/app/components/ilios-header.js
+++ b/app/components/ilios-header.js
@@ -3,10 +3,22 @@ import Ember from 'ember';
 export default Ember.Component.extend({
   session: null,
   currentUser: Ember.inject.service(),
+  i18n: Ember.inject.service(),
   userName: Ember.computed.oneWay('currentUser.model.fullName'),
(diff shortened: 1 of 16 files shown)


saschaben:
Hmm. this is failing for me locally: it can't load the select list. throws error on line 16 of ilios-header.js    
> `return this.get('i18n.locales').map(function (locale) {`

with a `Uncaught TypeError: Cannot read property 'map' of undefined`...

saschaben:
NB: requires nommore.

dartajax:
I got the picker to work and the language does change but the menu items >> Dashboard >> Course & Sessions >> Learner Groups >> Instructor Groups >> Programs don't reflect the language change.  Dashboard at the top doesn't change either.

jrjohnson:
I fixed the navigation and the titles.  Getting dates in the calendar recomputed is going to be a bit harder and required work with some other upstream modules.  I created #781 for this.

saschaben:
looks great.

thecoolestguy:
I also needed to nommore, but works/looks great once I did!

PR #789, opened 2015-07-27
Clean up browser console for developers

jrjohnson:
Since upgrading to Ember 1.13 our browser console has been cluttered with deprecation notices.  Dealt with this in two ways:

- Introduced an improved workflow for dealing with deprecations.  They are silenced by default and can be individually re-enabled and dealt with.

- Fixed a repeated warning from the simple-auth library.

There are still a few errors which cannot be caught in this way, but now the console is usable again.

### DIFF
--- a/config/deprecation-workflow.js
+++ b/config/deprecation-workflow.js
@@ -1,0 +1,19 @@
+window.deprecationWorkflow = window.deprecationWorkflow || {};
+window.deprecationWorkflow.config = {
+  workflow: [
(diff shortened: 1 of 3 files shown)


thecoolestguy:
The test keeps failing on the block comment in _navigation.scss (vs. using the // per-line style instead)... I can merge and then fix that after, but should I?  The other tests are passing...

thecoolestguy:
I'm going to merge this and fix the scss-lint disallowing of multi-line block comments.

PR #792, opened 2015-07-27
Use a custom authenticator for shibboleth

jrjohnson:
The ember-cli-simple-auth-token author declined to include this
functionality in the upstream library so I have added a custom
authenticator for ilios which handles our case.


thecoolestguy:
Looks like this is just failing because it's selecting the wrong day in the test.

PR #798, opened 2015-08-02
Update ember-cli version

jrjohnson:
Updated to ember-cli 1.13.6 which includes point release updates for ember-data ember ember
Changes to ember-data and JShint required fixing tests as well as the cohort model
Removed all of the blank component unit tests


jrjohnson:
Tests are failing because of an issue with the calendar in the 6 week August 2015.

jrjohnson:
The issue with tests is now solved in elemental-calendar upstream and fixed in #799

jrjohnson:
@saschaben not a lot of code review here, its mainly dependency stuff.  It does deserve a click around test though.

saschaben:
OK. clickthru shows things looking pretty much stable and good. I'm noticing some odd rendering slowness when moving learners from one group to another in Chrome (similar to the sort of latency we see when opening the parent objective manager in course, but I don't think it's a major issue, and is one we should address with a loading indicator when we can.

saschaben:
ready for a code ninja

PR #799, opened 2015-08-02
Fix calendar in safari

### LINKED ISSUE #786: Ilios Calendar Doesn't Display Properly on iPad or Smart Phone
I checked out an iPad from the Library and tested the calendar.  It doesn't display properly.

On the left side, there is a listing of the days of the week ...

Sun 7/19
Mon 7/20
Tue 7/21
Wed 7/22
Thu 7/23
Fri 7/24
Sat 7/25

And below that listing and over on the right side of the screen appears the listing of the times of day starting with 12:00am

                                                                                                                                                    12:00 am
                                                                                                                                                     1:00 am

and so on ..

jrjohnson:
Added an autoprefixer and updated the version of elemental-calendar to
fix calendar display in safari as well as an issue with the August 2015
calendar.

Fixes #786


dartajax:
It looks great in Safari now.

dartajax:
I tested it also on an iPad and that worked fine but brought up a usability issue that I will put in a new ticket.

PR #812, opened 2015-08-14
Remove Instruction-hour and recurring-event models

jrjohnson:
These are no longer part of the API


saschaben:
looks fine. Assuming that you can grep with the best of them (and have in this case)...
:+1:

PR #826, opened 2015-08-31
Add My Courses dashboard list

### LINKED ISSUE #749: Add Courses list to Dashboard
See #206 for details.

### LINKED ISSUE #755: "my courses" list not picking up all courses
on the Course List, the My Courses toggle is not capturing all associated courses for a given user.

The list should return an course where:

1. the user is a director of the course
1. the user is an instructor of a session in the course
1. the user is a member of an instructor group attached to a session in the course
1. the user is a member of a learner group attached to a session in the course
1. the user is a learner directly attached to a session in the course

cf demo_faculty6 in peer to peer mentoring for PISCES 2014-15, is an instructor as part of an instructor groups association, but is not retrieved by the course list.

jrjohnson:
The widget will only appear when a user has some related courses. 

Currently this will display for students too as they are in learnerGroups associated with a course.  We will probably be hiding some dashboard items from students generically so I didn’t handle that here.
 

Fixes #749
Incidentally fixed #755 (my courses toggle on courses list)

### DIFF
--- a/app/components/dashboard-mycourses.js
+++ b/app/components/dashboard-mycourses.js
@@ -1,0 +1,22 @@
+import Ember from 'ember';
+import DS from 'ember-data';
+
+export default Ember.Component.extend({
+  currentUser: Ember.inject.service(),
+  tagName: 'div',
+  classNames: ['dashboard-block'],
+  listOfCourses: Ember.computed('currentUser.model.allRelatedCourses.[]', function(){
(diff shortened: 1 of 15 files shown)


saschaben:
having this display for students is a plus. this is a handy widget for all users; the fact that we didn't have a way to give it to students previously was a bug, not a feature.

PR #836, opened 2015-09-21
added multi-edit functionality

### LINKED ISSUE #612: Implement Bulk Editing for Learner Groups
Per our discussion at the team meeting on May 21, in order to allow users to edit groups & subgroups when they are dealing with a large number of people, we're going to try adding "Bulk Edit" functionality.

**If anyone can think of a better name for "Bulk Edit", speak up!**

We need to make sure that when Bulk Edit is on, that the list of members NOT in this group become read-only (or we need to discuss the best solution)...

Bulk Edit = on

learner-groups-edit-v10-bulkedit

Default State (Bulk Edit = off)
learner-groups-edit-v10-default

djvoa12:
Additional Notes:
- added `ui-select` (dropdown select) addon dependency (see `package.json`)
  - created new `learner-select-field.hbs` which replaces default select-field that `ui-select` provides to fit our needs
  - dependency: `liquid-tether`
- refactored, where applicable
- added styles, where applicable
- added new words to `locales` for translation purposes
- created new component `multiedit-select`
- added test
  - unit
  - acceptance

fixes #612


app/components/inplace-select.js
```diff
@@ -64,5 +66,31 @@ export default Ember.Component.extend(InPlace, {
       newValue = newValue==='null'?null:newValue;
       this.send('changeValue', newValue);
     },
-  }
+  },
+
+  closeEditor: observer('multiEditModeOn', function() {
```
jrjohnson: InPlace is where shards code for select, text, textarea, etc... gets stored.  I think these methods belong somewhere else.

app/components/learner-select-field.js
```diff
@@ -0,0 +1,6 @@
+import UISelectFieldComponent from 'ui-select/components/ui-select-field';
+import layout from '../templates/components/learner-select-field';
+
+export default UISelectFieldComponent.extend({
+  layout
```
jrjohnson: I don't understand how this line works.  Shouldn't it need to be `layout: layout`?

app/components/learnergroup-members.js
```diff
@@ -1,7 +1,19 @@
 import Ember from 'ember';
 import DS from 'ember-data';
+import InPlace from 'ilios/mixins/inplace';
+
+const { computed, isEmpty, ObjectProxy, observer, RSVP } = Ember;
+const { all } = RSVP;
+
+export default Ember.Component.extend(InPlace, {
```
jrjohnson: OK, I see why you're using InPlace here, but I don't think this functionality is going to get used anywhere else so it should all just be placed in the component.  Am I wrong about that?

app/components/learnergroup-members.js
```diff
@@ -1,7 +1,19 @@
...
+  init() {
```
jrjohnson: Is there an advantage to doing this in an init() as opposed to just setting these as properties on the component?

app/components/learnergroup-members.js
```diff
@@ -122,38 +134,106 @@ export default Ember.Component.extend({
...
+      this.get('toBulkSave').pushObject(studentId);
+    },
+
+    removeStudent(studentId) {
+      this.get('toBulkSave').removeObject(studentId);
+    },
+
+    continueButton() {
+      if (isEmpty(this.get('toBulkSave'))) {
+        this.set('errorMessage', 'You must checkmark at least one.');
```
jrjohnson: Instead of having an error and a string that has to be translated, how about if the button is just disabled when nothing has been checked.  I think thats clear enough.

app/components/learnergroup-members.js
```diff
@@ -122,38 +134,106 @@ export default Ember.Component.extend({
...
+    },
+
+    changeSelection(newValue) {
+      newValue = newValue.get('value') === 'null' ? null : newValue.get('value');
+      this.send('changeValue', newValue);
+    },
+
+    bulkSave(value) {
+      this.get('toBulkSave').forEach((studentId) => {
+        this.sendPutRequests(value, studentId);
```
jrjohnson: I think there is a lot of potential for a race condition here.  As each student saves it also sends a put request with all the learner groups and all of their students.  Depending on the order it will be possible for a student to be removed once and then added back again when the next student saves.  Its possible that ember-date will save this from happening, but I'm not sure if that is true.

app/components/learnergroup-members.js
```diff
@@ -122,38 +134,106 @@ export default Ember.Component.extend({
...
+
+  errorMessage: '',
+  atLeastOneChecked: observer('toBulkSave.@each', function() {
+    if (isEmpty(this.get('toBulkSave'))) {
+      this.set('editorModeOn', false);
     }
-  }
+  }),
+
+  includeAll: false,
```
jrjohnson: It has been our convention to put the `actions` hash last in all components and controllers.  Please move everything above the actions key.

app/components/learnergroup-members.js
```diff
@@ -122,38 +134,106 @@ export default Ember.Component.extend({
...
+
+    bulkSave(value) {
+      this.get('toBulkSave').forEach((studentId) => {
+        this.sendPutRequests(value, studentId);
+      });
+      this.setProperties({ toBulkSave: [], editorModeOn: false });
+    }
+  },
+
+  errorMessage: '',
```
jrjohnson: It has been our convention to put the `actions` hash last in all components and controllers.  Please move everything above the actions key.

app/components/learnergroup-members.js
```diff
@@ -122,38 +134,106 @@ export default Ember.Component.extend({
...
+  atLeastOneChecked: observer('toBulkSave.@each', function() {
```
jrjohnson: Apparently observers are evil (according to some ember core team peeps).  This should be refactored into a computed property.

app/components/learnergroup-overview.js
```diff
@@ -27,9 +29,20 @@ export default Ember.Component.extend({
...
     },
+
     changeLocation: function(value){
       this.get('learnerGroup').set('location', value);
       this.get('learnerGroup').save();
     },
-  }
+
+    toggleSwitch() {
+      if (this.get('multiEditModeOn')) {
```
jrjohnson: Can be condensed to `this.set('multiEditMode', !this.get('multiEditMode'));`

app/components/learnergroup-overview.js
```diff
@@ -27,9 +29,20 @@ export default Ember.Component.extend({
...
+        this.set('multiEditModeOn', false);
+      } else {
+        this.set('multiEditModeOn', true);
+      }
+    }
+  },
+
+  multiEditModeOn: false
```
jrjohnson: Move property above actions hash

app/locales/es/translations.js
```diff
@@ -90,7 +90,9 @@ export default {
     'savedSuccessfully': 'Ahorrado Con Éxito',
     'notFoundMessage': "Malditas! Éso no puede encontrar. Favor de verificar la dirección de la página, y trata otra vez.",
     'myCourses': 'Mis Cursos',
-    'waitSaving': 'salvando... un momento...'
+    'waitSaving': 'salvando... un momento...',
+    'continue': 'Continúa',
```
jrjohnson: @dartajax please check these spanish translations.

app/locales/fr/translations.js
```diff
@@ -90,7 +90,9 @@ export default {
     'savedSuccessfully': "Réussie sauvé",
     'notFoundMessage': "Zut! Je ne pouvais pas trouver cette chôse. S'il vous plaît vérifier votre adresse de la page et essayez à nouveau.",
     'myCourses': "Mes Cours",
-    'waitSaving': "Travaille dessus... un moment..."
+    'waitSaving': "Travaille dessus... un moment...",
+    'continue': 'Continuez',
```
jrjohnson: @saschaben please check these french translations

app/components/inplace-select.js
```diff
@@ -1,6 +1,8 @@
 import Ember from 'ember';
 import InPlace from 'ilios/mixins/inplace';
 
+const { observer } = Ember;
+
 export default Ember.Component.extend(InPlace, {
```
jrjohnson: I don't think inPlace select is the place for this.  InPlaceSelect is used when we want to put a checkbox next to an item and save it immediately.  These should just be regular select lists because the save happens when the continue button is clicked.

app/mixins/inplace.js
```diff
@@ -1,18 +1,21 @@
 import Ember from 'ember';
 
-export default Ember.Mixin.create({
-    i18n: Ember.inject.service(),
+const { computed, inject, Mixin } = Ember;
+const { service } = inject;
```
jrjohnson: I've never seen this done, it makes the service definition way better.  Nice!

app/locales/fr/translations.js
```diff
@@ -160,6 +162,8 @@ export default {
     'cohortMembersNotInGroup': "Membres de la Cohorte sans affectation à {{groupTitle}} ",
     'notInThisGroup': "Pas dans ce groupe",
     'selectGroup': "Choisi Groupe",
+    'multiEdit': 'Modifier Multiples',
+    'includeAll': 'Inclure Toute La Cohorte'
```
saschaben: Should be "Incluez tous les membres de la Cohorte"

app/styles/_variables.scss
```diff
@@ -24,6 +24,7 @@ $ilios-orange: #c60;
 $background-grey: #fbfbfb;
 $toggle-background-grey: #eceeef;
 $light-grey: #fefefe;
+$light-grey2: #d3d3d3;
```
jrjohnson: This would usually be the right place for this, but there are so many greys here already I think this should go into learner-group-members.scss as the variable $learnerGroupGrey

app/locales/fr/translations.js
```diff
@@ -160,6 +162,8 @@ export default {
     'cohortMembersNotInGroup': "Membres de la Cohorte sans affectation à {{groupTitle}} ",
     'notInThisGroup': "Pas dans ce groupe",
     'selectGroup': "Choisi Groupe",
+    'multiEdit': 'Modifier Multiples',
```
saschaben: should be "Modifier de multiples groupes"

app/styles/components/_learner-group-overview.scss
```diff
@@ -0,0 +1,13 @@
+.toggle-onoff {
```
jrjohnson: This is too generic and will end up hitting this box when we place it other places where it is not aligned like this.  I think we probably need to create an alignment class that is more generic and maybe put it in _detailView.

app/templates/components/learner-select-field.hbs
```diff
@@ -0,0 +1,6 @@
+{{#if selection.value}}
```
jrjohnson: I don't like using a span and icon here when a generic HTML select tag will do the same job.  Is there a benefit to doing it this way?

app/locales/es/translations.js
dartajax: Salvando... un momento...

dartajax:
Do you have more translations for me to check, @jrjohnson ?

app/locales/es/translations.js
```diff
@@ -90,7 +90,9 @@ export default {
     'savedSuccessfully': 'Ahorrado Con Éxito',
     'notFoundMessage': "Malditas! Éso no puede encontrar. Favor de verificar la dirección de la página, y trata otra vez.",
     'myCourses': 'Mis Cursos',
-    'waitSaving': 'salvando... un momento...'
+    'waitSaving': 'salvando... un momento...',
```
dartajax: Salvando... un momento...

app/locales/es/translations.js
```diff
@@ -160,6 +162,8 @@ export default {
     'cohortMembersNotInGroup': 'Miembros de la Clase NO Asignados a {{groupTitle}} ',
     'notInThisGroup': 'No en este grupo',
     'selectGroup': 'Seleccione Grupo',
+    'multiEdit': 'Editar Múltiples',
+    'includeAll': 'Incluya Toda La Cohorte'
```
dartajax: Should be 'Incluye Todos Los Miembros del Cohorte'

app/locales/es/translations.js
```diff
@@ -160,6 +162,8 @@ export default {
     'cohortMembersNotInGroup': 'Miembros de la Clase NO Asignados a {{groupTitle}} ',
     'notInThisGroup': 'No en este grupo',
     'selectGroup': 'Seleccione Grupo',
+    'multiEdit': 'Editar Múltiples',
```
dartajax: Should be 'Editar múltiples grupos'

app/templates/components/learnergroup-members.hbs
```diff
@@ -1,18 +1,59 @@
...
   </div>
-{{/each}}
+{{/if}}
+
+<div class="learnergroup-list">
+  {{#each proxiedMembers as |userProxy|}}
+    <div class="form-col-12">
+      <label class="form-label learnergroup-username">{{userProxy.fullName}}</label>
+      <div class="form-data learnergroup-group-membership">
+        {{inplace-select
```
jrjohnson: I think these can just be HTML <select> tags with actions.  Take a look at the inplaceSelect handlebars for an example.  Any reason not to do it that way?

app/templates/components/learnergroup-members.hbs
```diff
@@ -1,18 +1,59 @@
...
+    <p class="continue-error-message">{{errorMessage}}</p>
+  {{/if}}
+
+  <button class="continue-button" {{action 'continueButton'}}>{{t 'general.continue'}}</button>
+{{/if}}
+
+{{#if editorModeOn}}
+  <div class="choose-action__container">
+    <div class="choose-action__options">
+      {{#ui-select field-component="learner-select-field" showPlaceholderAsOption=false placeholder='general.select' on-select="changeSelection" value=value inline=true items=proxiedOptions as |item|}}
```
jrjohnson: Also looks like it can just be a plain HTML <select> tag.

app/templates/components/learnergroup-members.hbs
```diff
@@ -1,18 +1,59 @@
...
+        {{#ui-option value=item}}
+          {{item.label}}
+        {{/ui-option}}
+      {{/ui-select}}
+    </div>
+
+    <span class='choose-action__actions'>
```
jrjohnson: Why make this a two step process?  Could just show the select menu and have a save button instead of going through the continue step?

app/templates/components/learnergroup-overview.hbs
```diff
@@ -3,13 +3,19 @@
...
+      {{toggle-onoff label="learnerGroups.multiEdit" on=multiEditModeOn action="toggleSwitch"}}
+    </div>
+
     {{learnergroup-members
       topLevelGroup=learnerGroup.topLevelGroup
       cohort=learnerGroup.cohort
       members=learnerGroup.usersOnlyAtThisLevel
       showMoveToCohortOption=true
       showMoveToTopLevelGroupOption=true
+      multiEditModeOn=multiEditModeOn
```
jrjohnson: I'm mostly OK with the way this breaks down using the learnergroup-members component with a multi-edit option, but it may be a lot cleaner to create a multiedit-learnergroup-members component and swap that out here instead of passing in this option.  Take a look, but feel free to leave it this way if that makes the most sense.

app/components/learner-select-field.js
djvoa12: It's the ES6 syntax for object literal shorthand for property names that are same as value variable names.

```javascript
let name = 'Jon';
let hash = { name: name } ;
```
is equivalent to
```javascript
let hash = { name } ;
```

app/components/learnergroup-members.js
djvoa12: Yes, I am going to move them back to the component.

app/components/learnergroup-members.js
djvoa12: I am going to remove the `init` method for this particular case because the component is huge and has dependencies. When you have a smaller component (w/o other dependent components) that is going to be used a lot, it's good practice to use the `init` method because setting them explicitly would put them on component's prototype, which is shared across all instances.  It's just good practice.

app/components/learnergroup-members.js
djvoa12: Okay. Users might be confused initially, but I suppose they can figure it out.

app/components/learner-select-field.js
jrjohnson: Sweet!

app/components/learnergroup-members.js
djvoa12: Done.

app/components/learnergroup-overview.js
djvoa12: :+1:

app/components/learnergroup-overview.js
djvoa12: Done.

app/mixins/inplace.js
djvoa12: Yes, ES6 provides many new syntax, including destructuring for organization. :+1:

app/locales/fr/translations.js
djvoa12: Done.

app/styles/_variables.scss
djvoa12: Done.

app/locales/fr/translations.js
djvoa12: Done.

app/templates/components/learner-select-field.hbs
djvoa12: I used this select addon that I helped build. It's more flexible in terms of styling, can support placeholders (as an option as well), and fires primary action when one changes the value.

app/locales/es/translations.js
djvoa12: Done.

app/locales/es/translations.js
djvoa12: Done.

app/locales/es/translations.js
djvoa12: Done.

app/mixins/inplace.js
jrjohnson: It's the double destructure that I really like.  So you get `store: service()` in the component.  That's very clean.

app/components/learnergroup-members.js
djvoa12: Yup. I am working on converting them over.

app/locales/es/translations.js
djvoa12: No need for this anymore as I am removing the button.

app/locales/fr/translations.js
djvoa12: No need for this anymore as I am removing the button.

app/templates/components/learnergroup-members.hbs
djvoa12: This is the addon component. It needs to be called this way, until angle-bracket component come out.

app/templates/components/learnergroup-members.hbs
djvoa12: That's a good idea. I am going to remove the `continue` button.

app/templates/components/learnergroup-members.hbs
jrjohnson: I mean I don't think there is any reason for the addon component here.  It's just a <select> element which is bound to an action.  Just using the straight <select> as it is in the inplace-select.hbs will allow for simplification later and lets the browser do the work.

app/components/inplace-select.js
djvoa12: We should discuss this more in person. I believe `InPlace` mixin was already included before I started.

saschaben:
This is starting to look good -- we are however still missing a piece of the expected functionality. if you look at the original mock you see that we not only have the ability to multi-select users from the currently selected group for re-assignment, we also have the ability to select any user from the hierarchy as well (the listings in the box "[group name] learner assignments". Use case here would be when you are looking at the 'anatomy lab' parent group -- which has no primary members, since everyone is plugged into one of the subgroups--- and you want to move one person from each of the subgroups to a different group --eg, taking a single learner from lab 1, lab 2, lab 3, lab 4, etc, and putting all those into lab 27. As is stands now, to do this would require in the above example 4 separate single actions, despite the bulk option. If this is out of the range of doable right now, let's all talk.

also, the "include entire cohort" option should really just be labeled "select all members in current group" -- since cohort has a specific meaning in this context, and it isn't synonymous with the members of a single group or subgroup displayed.

jrjohnson:
Good catch @saschaben I completely missed that.  @djvoa12 this will mean changing the way you have this structured because the list of proxiedMembers in learner-group-members is not the same in multi-edit mode.  It will instead need to include all the learners in the top level group.  

@saschaben - should this list include learners who are not in the group tree (aka cohort people)?

saschaben:
@jrjohnson list should include just the tree people. not the cohort people.

jrjohnson:
:+1:  @saschaben please review functionality and merge if it is ready

saschaben:
This looks awesome! Functionally I think it all checks out (assuming the work you are doing on the model allows for performance on the API that somewhat matches local test data); one remaining item that I see is that the field entries for Location and default instructor should be disabled when a user flips into mutli-mode:

add

I actually think that disabling or greying out is in this instance a better pattern, since it helps to orient the user in terms of modality / context of use.

jrjohnson:
I'm not sure I agree with those being disabled?  You can still change attributes of your current group.  and you can still scroll down to change other things.

saschaben:
(...and I know it's not part of this ticket, but....when are we going to get some back button / active link breadcrumbing action on this page? now thatit's getting truly functional, it's annoying as hell to not be able to easily traverse the tree hierarchies.)

PR #845, opened 2015-10-02
Split code out into calendar addon

jrjohnson:
All of this code is now in the ilios/calendar add-on which can be used both by the fronted and by the embedded app.


app/locales/en/translations.js
```diff
@@ -294,6 +294,7 @@ export default {
     'month': 'Month',
     'today': 'Today',
     'backToCalendar': 'Back to Calendar',
+    'loadingEvents': 'Loading Events...',
```
jrjohnson: @saschaben / @dartajax please provide a translation for `Loading Events...`;

saschaben: I think for the french it will be best to use `Rassembler les événements...` ("getting events...")

dartajax: Cargando eventos ...

jrjohnson:
This is ready for review

jrjohnson:
@thecoolestguy I rebased, but tests should run again just in case.

PR #851, opened 2015-10-05
added small group generator

### LINKED ISSUE #697: Add the offerings Small Group Generator
We can't let the ball drop on this: implement the tool which allows the option of creating a collection of offerings, driven by the list of groups selected.

https://www.dropbox.com/work/usability?preview=SmallGroupGenerator.png

This includes modifying the "Add Offering" button to be a drop down, with the options of "Add a Single Offering" and "Add Multiple Offerings (small groups)"

### LINKED ISSUE #758: Offering Time am/pm set is buggy
cant set pm start/end times on an offering

djvoa12:
Added an editor that can create a single offering or multiple offerings at once (small groups).

- Fixed AM/PM bug in `time-picker`
- Switched out old learner group editor with the new one in `offering-manager`
- Tests:
  - Unit
  - Integration
  - Acceptance (creating/editing single/multiple offering(s))

fixes #697 
fixes #758 (may need to replace time-picker with other addons due to rigid data-binding)


saschaben:
some squiggly time issues:

first, the bug in #758 is still apparent here.
Second, the time displayed in the session list for "first offering" is showing current time rather than the earliest date/time offering available when the session is an ILM. This is a quirky issue but we should resolve how to address it.

app/components/offering-editor-learnergroups.js
```diff
@@ -0,0 +1,47 @@
+import Ember from 'ember';
+
+const { Component, computed, isEmpty, isPresent } = Ember;
+const { sort, filter } = computed;
+
+export default Component.extend({
+  classNames: ['offering-editor-learnergroups'],
+
+  revisedLearnerGroups: computed('learnerGroups', {
+    get() {
```
jrjohnson: How does having a get() method and a readonly() extension work here?

app/components/offering-editor.js
```diff
@@ -0,0 +1,190 @@
...
+    changeEndTime(date) {
+      let newEnd = moment(date);
+      let endDate = moment(this.get('endDate'));
+
+      endDate.hour(newEnd.format('HH'));
+      endDate.minute(newEnd.format('mm'));
+      this.set('endDate', endDate.toDate());
+    },
+
+    addInstructorToBuffer(instructor){
```
jrjohnson: These method names *ToBuffer reflect the old pattern of creating a buffer object to send values to.  Seeing as how this component stores these values internally I think they can just be called addInstructor/addInstructorGroup for clarity.

app/components/offering-editor.js
```diff
@@ -0,0 +1,190 @@
...
+
+    removeInstructorGroupFromBuffer(instructorGroup){
+      this.get('instructorGroups').removeObject(instructorGroup);
+    },
+
+    addLearnerGroup(group) {
+      let learnerGroups = this.get('learnerGroups');
+
+      group.get('allDescendants').then((descendants) => {
+        if (isEmpty(descendants)) {
```
jrjohnson: This is good code, but it needs a comment along the lines of 
//If the group has no descendants then we add it, otherwise we add the descendants, but not the parent.

jrjohnson:
I think we're missing a piece of functionality from the learner group chooser.  The selected learner groups (on the left) is a flat list with the parent group shown via the name, but not indented.  This is OK except it means we're missing the ability to remove a whole set of children by clicking their parent.  
Does that make sense?  @saschaben can you take a look?  Do we need that ability, or can users just clickety click all the child groups away?

app/components/offering-editor.js
```diff
@@ -0,0 +1,190 @@
...
+        }
+      });
+    },
+
+    removeLearnerGroup(group) {
+      this.get('learnerGroups').removeObject(group);
+    },
+
+    create() {
+      let startDate = moment(this.get('startDate'));
```
jrjohnson: This is where you can use startDate.hour(8) startDate.minute(0) to default all the new offerings to start at 8am.

app/components/offering-editor.js
```diff
@@ -0,0 +1,190 @@
...
+    this.setProperties({ instructors: [], instructorGroups: [], learnerGroups: [] });
+  },
+
+  classNames: ['offering-editor'],
+
+  singleOffering: true,
+  isMultiDay: false,
+  filter: '',
+
+  startDate: new Date().setHours(8, 0, 0),
```
jrjohnson: These properties should probably be called defaultStartDate / defaultEndDate unless they are getting set somewhere?

app/components/session-offerings.js
```diff
@@ -3,28 +3,153 @@ import Ember from 'ember';
...
-    add: function(){
-      var self = this;
-      var offering = this.get('store').createRecord('offering', {
-        session: self.get('session'),
-        startDate: moment().hour(8).minute(0).second(0).format(),
-        endDate: moment().hour(9).minute(0).second(0).format(),
+    addSingleOffering({ startDate, endDate, room, learnerGroups, instructors, instructorGroups }) {
+      const store = this.get('store');
+      const session = this.get('session');
+      const offering = store.createRecord('offering', { session, startDate, endDate, room });
```
jrjohnson: this should be let instead of const.  The object isn't getting modified, but it isn't really a constant.  JS seems to be a little wishy washy right now about how this works, but I would prefer a let here.

app/components/session-offerings.js
```diff
@@ -3,28 +3,153 @@ import Ember from 'ember';
...
+
+      offering.save().then((offering) => {
+        offering.get('learnerGroups').then((offeringlearnerGroups) => {
+          learnerGroups.forEach((learnerGroup) => {
+            offeringlearnerGroups.pushObject(learnerGroup);
+            learnerGroup.save();
```
jrjohnson: Just a reminder to do all of these changes and then save the offering at the end instead of saving each learner group / instructor / instructor group.  That should be a a single put request which should be much faster.

app/components/session-offerings.js
```diff
@@ -3,28 +3,153 @@ import Ember from 'ember';
...
     },
-    removeNewOffering: function(offering){
-      this.get('newOfferings').removeObject(offering);
+
+    addMultipleOfferings({ learnerGroups, startDate, endDate }) {
+      this.set('saving', true);
+
+      const store = this.get('store');
+      const session = this.get('session');
+      const offeringPromises = [];
```
jrjohnson: offeringPromises should be a `let` not a `const`

app/components/session-offerings.js
```diff
@@ -3,28 +3,153 @@ import Ember from 'ember';
...
+
+      learnerGroups.forEach((learnerGroup) => {
+        const room = learnerGroup.get('location');
+        const offering = store.createRecord('offering', { session, startDate, endDate, room });
```
jrjohnson: `let` not `const`

app/models/learner-group.js
```diff
@@ -19,6 +19,7 @@ export default DS.Model.extend({
     }
   ),
   instructorGroups: DS.hasMany('instructor-group', {async: true}),
+  offering: DS.belongsTo('offering', {async: true, inverse: 'learnerGroups'}),
```
jrjohnson: Can you remind me why this is correct?  Seems wrong since a learnerGroup has many offerings and never just one.

app/styles/components/_offering-editor-learnergroups.scss
```diff
@@ -0,0 +1,40 @@
+@mixin subgroup-style {
```
jrjohnson: This should be styled using the existing _tree-list.scss.  That file can be modified, but we want to use the same visual style everywhere this UX exists.

app/styles/components/_offering-editor.scss
```diff
@@ -0,0 +1,48 @@
+.offering-editor {
```
jrjohnson: This form looks great, but it should be styled with the existing form styles in place on the rest of the site and match those other forms.  Otherwise its too hard to maintain long term.

app/styles/components/_session-offering.scss
```diff
@@ -0,0 +1,7 @@
+.plus-button, .minus-button {
```
jrjohnson: Plus / minus button style should go in its own scss file so we can move it into a pod later.

app/styles/components/_session-offering.scss
```diff
@@ -0,0 +1,7 @@
...
+  outline: none;
+}
+
+.offering-section-top {
+  overflow: hidden;
```
jrjohnson: What does this line do?  Seems like a strange place to hide the overflow.

app/templates/components/session-offerings.hbs
```diff
@@ -1,22 +1,35 @@
...
+        <span class="switch-handle"></span>
+      </label>
+    </div>
+
+    <div class='detail-actions'>
+      {{expand-collapse-button value=offeringEditorOn action="toggleEditor"}}
+    </div>
   </div>
 
+  {{#if offeringEditorOn}}
```
jrjohnson: For consistency this should by {{#liquid-if offeringEditorOn class='vertical'}}  and the block should end with {{/liquid-if}}

saschaben:
@jrjohnson The chooser is flattened knowingly: there are some usability challenges which are obviated by doing this, and it simplifies both code and patterns. I think the additional clickety click to remove selected groups rather than bulk deselecting is for now worth the upside of the change.

app/components/offering-editor-learnergroups.js
djvoa12: It's the latest `getter` for CPs. You can also have a `set` but we are not explicitly setting the CP so not necessary. `readonly()` will block all attempts to set it.

http://emberjs.com/api/classes/Ember.ComputedProperty.html

app/components/offering-editor.js
djvoa12: Done.

app/components/offering-editor.js
djvoa12: Done.

app/components/session-offerings.js
djvoa12: Done.

app/components/session-offerings.js
djvoa12: Should the offering be saved first (`post` request to server) for it to get an id and then all the child associations to be made afterwards (with `put` requests)?

app/components/session-offerings.js
djvoa12: Done.

app/components/session-offerings.js
djvoa12: Done.

app/styles/components/_offering-editor-learnergroups.scss
djvoa12: I made modifications to `_tree-list.scss` so the style looks consistent, but `subgroup-style` needs to remain in here because the structure for `offering-editor-learnergroups` is not consistent with others. It uses one layer of `li` tags , instead of two layers.

app/styles/components/_session-offering.scss
djvoa12: Its children are floats so I want that container to cover all of its children. The default behavior for parent containing floating children is that it will be not visible when you look for it.

app/components/session-offerings.js
jrjohnson: yes. I think that's the most reliable way to handle it.

app/styles/components/_offering-editor-learnergroups.scss
jrjohnson: Makes sense given the feature discussion here.

app/templates/components/session-offerings.hbs
djvoa12: I will make the change. Do you think, in general, the slide out form is slightly unnecessary for user-interaction? It would be great if it was fast-acting with no lag, but it's the opposite, visually slow, and doesn't add much to the user-experience.

app/styles/components/_session-offering.scss
djvoa12: Done.

app/templates/components/session-offerings.hbs
jrjohnson: We have it everywhere so we should have it here.  When people are tired of it (as you're right they will be) then we will remove it from everywhere.

app/styles/components/_offering-editor.scss
djvoa12: Done. I commented out the CSS for future reference, instead of removing it all.

app/models/learner-group.js
djvoa12: I removed this line along with others. The changes are local currently, not reflected on here yet.

app/components/offering-editor.js
```diff
@@ -0,0 +1,199 @@
...
+export default Component.extend({
+  init() {
+    this._super(...arguments);
+
+    this.setProperties({
+      instructors: [],
+      instructorGroups: [],
+      learnerGroups: [],
+      startDate: moment().format('MM/DD/YYYY'),
+      endDate: moment().add(1, 'days').format('MM/DD/YYYY')
```
jrjohnson: endDate should default to one hour after start date.

app/components/offering-editor.js
```diff
@@ -0,0 +1,199 @@
+/* global moment */
```
jrjohnson: moment should now be used as 
`import moment from 'moment';`
and not as a global

app/components/offering-editor.js
```diff
@@ -0,0 +1,199 @@
...
+      }),
+    });
+
+    let cohorts = this.get('cohorts') ? this.get('cohorts') : [];
+
+    return cohorts.map((cohort) => {
+      let proxy = cohortProxy.create({
+        content: cohort,
+        selectedLearnerGroups: this.get('learnerGroups')
+      });
```
jrjohnson: should also pass `filter: this.get('filter')` into the cohort proxy.

jrjohnson: Nevermind.  The filter is no longer an option on this component.  If we want to implement it it would need to be further down the tree in the cohort learner group component.  In that case just remove the filter property from line 27

saschaben:
this is looking great. we are still a little challenged in the time-picker area as noted. a couple of things:

1. when an offering is created, and then opened in edit mode, the time selector should display the existing start/end times for the offering. currently selectors are defaulting to 1pm.

2. when an offering is created, if the start/end time is left at the default display (8/9), it will save as 12am regardless.

If we can't get around this problem with the current control, can we add a default blank value for time, and use form validation to force a selection?

Also, assuming we will do this on the form validation side as well, but we need to make sure we don't allow end times which precede start times.

app/components/offering-manager.js
```diff
@@ -1,7 +1,12 @@
-import moment from 'moment';
-
+/* global moment */
```
jrjohnson: global moment can go away as the import statement is on line 4.

app/components/time-picker.js
```diff
@@ -2,11 +2,24 @@ import moment from 'moment';
...
+
 export default Ember.Component.extend({
   classNames: ['time-picker'],
   date: null,
-  dateBuffer: Ember.computed.oneWay('date'),
-  hour: momentFormat('dateBuffer', 'h'),
+  dateBuffer: oneWay('date'),
+
+  hour: computed('dateBuffer', function() {
+    const defaultHour = this.get('defaultHour');
```
jrjohnson: I think the default hour stuff shouldn't be part of this component.  When it is created it gets passed a data, if the calling component wants a default then it should make one up before calling time-picker.

app/styles/components/_tree-list.scss
```diff
@@ -11,16 +11,21 @@
       margin-left: 2em;
 
       li {
-        font-weight: normal;
+        font-size: 14px;
```
jrjohnson: All font sizes are relative to $base-font-size
https://github.com/ilios/frontend/blob/edee5bb3f39931a6649daf198f663f1631a9b506/app/styles/components/_table.scss#L3

app/components/offering-editor.js
djvoa12: Because

app/components/time-picker.js
djvoa12: This was extremely tricky to pull off. It doesn't let me do both: set a visible formatted date on the start-date input AND have the time data passed down to reflect the default time. When the date gets formatted to "MM/DD/YYYY", it resets the hours and minutes.

app/components/offering-editor.js
djvoa12: Done.

app/components/offering-editor.js
djvoa12: Done.

app/components/offering-manager.js
djvoa12: Done.

app/styles/components/_tree-list.scss
djvoa12: Changed it back.

jrjohnson:
:+1:

jrjohnson:
@saschaben please review and merge if ready.  @djvoa12 will work separately on the remaining time / date issues, but I want to get the majority of this merged so we can be certain it will be in the release next week and so the time / date can be a more focused discrete PR.

saschaben:
@jrjohnson OK. Will eview fo primary functionality and UX less the time bug.

PR #857, opened 2015-10-06
Model cleanup

jrjohnson:
Aligned the fronted models with the current API.


jrjohnson:
The single failing test here is not our problem and I submitted a PR upstream to resolve.  This can, and should, be merged with this failing test.  Otherwise it will hold up work in a lot of other places.

thecoolestguy:
looks ok to me! merging even though it has an error, based on @jrjohnson's notes above...

PR #863, opened 2015-10-07
Add the Froala WYSIWYG Editor

### LINKED ISSUE #458: Add WYSIWYG editor option to inplace-textarea / big-text components

jrjohnson:
This seems to be our only choice for a relatively modern toolbar based editor which plays nicely with ember.

Before this is merged it needs to be added to and tested on:
- [x] Edit objective title
- [x] New objective title
- [x] Session Description
- [x] Learning material description
- [x] Learning material notes
- [x] Fix lists in rendered output
- [x] fix lists inside editor

Fixes #458


jrjohnson:
@saschaben can you enumerate all of the places this editor should be displayed?

@saschaben / @dartajax please test the editing of a course objective title to get a feel for this thing and see if it will work for us.  A heroku build should be available shortly to make it easy to try out.

saschaben:
Editor needs to be in place for:

Program objective
Course objective
Session objective
Session description
Learning material description
Learning material notes

saschaben:
Hmm. Looks like it may be a little buggy. delimited lists don't work -- it adds an unnecessary UL tag which breaks them; ordered lists don't work and I can't tell why.

Also: we will need to modify the UI display so that we are rendering any formatting in both edit AND non-edit modes, otherwise we run the high risk of making most entries with any real level of formatting unintelligible.

saschaben:
Attempting to edit/update a session objective throws an error destroying the objective; also attempting to show html code on same has the same impact. Only occurring on session, not course.

jrjohnson:
Yes, please just test on course objective.  The other editors may or may not appear, but I haven't hooked them up.

saschaben:
multiple edits causes course objective to be cleared:

err2

saschaben:
"Show HTML" is blank after cut& paste from word.

jrjohnson:
@saschaben please test again - you will need to run`nomless` to get the latest version of everything.  The only remaining bug I see is that clicking the show HTML button causes a black screen.  This only happens if you have not focused the cursor into the editor window before clicking the button.  I tried several pastes from word and all appeared to render correctly.
I didn't know what:
>delimited lists don't work -- it adds an unnecessary UL

meant as I wasn't able to get an extra UL tag to show up anywhere.

jrjohnson:
This is ready for review

jrjohnson:
I lied, not quite ready, needs some translations.

jrjohnson:
OK, now we're ready for review.  All translations in.

app/components/big-text.js
```diff
@@ -8,28 +12,36 @@ export default Ember.Component.extend({
...
+  showIcons: computed('displayText', 'text', 'renderHtml', function(){
+    if(this.get('renderHtml')){
+      return this.get('displayText').toString() !== this.get('text');
+    } else {
+      return this.get('displayText').toString() !== this.get('cleanText');
     }
+  }),
+  cleanText: computed('text', function(){
     //strip any possible HTML out of the text
-    return text.replace(/(<([^>]+)>)/ig,"");
```
stopfstedt: seems kinda hinky.
why not use jQuery here: http://api.jquery.com/text/

app/components/big-text.js
```diff
@@ -8,28 +12,36 @@ export default Ember.Component.extend({
...
-  }.property('text'),
-  displayText: function(){
-    var text = this.get('cleanText');
-    if(this.get('expanded') || text.length < this.get('totalLength')){
-      return text;
+    return this.get('text').replace(/(<([^>]+)>)/ig,"");
```
stopfstedt: likewise.

app/components/big-text.js
jrjohnson: Never occured to me actually.  Good change

jrjohnson: @stopfstedt jquery.text() is for working with elements.  We're working with a string passed from somewhere else like the DB.  This would lead to something like:
```return  Ember.$("<div/>").html(text).text();```  I'm not sure that isn't more confusing.  This isn't a security step, its purely to easy in the display.  Thoughts?

stopfstedt: right, forgot that you'll actually have to wrap your string in markup first.
that _is_ more confusing.
let's go with your initial solution for now.

stopfstedt:
:+1:

PR #868, opened 2015-10-11
fixed date-time issue, created validation, wrote tests

djvoa12:
- Fixes all date & time picker related bugs
- Includes date validation (end-date cannot be on or earlier than start-date)
- Includes time validation (end-time cannot be on or earlier than start-time)
- validation error-messages to display via `ember-flash-messages`
- Minor refactoring of `offering-editor`,  `offering-manager`, etc.

### DIFF
--- a/app/components/offering-editor.js
+++ b/app/components/offering-editor.js
@@ -11,13 +11,16 @@ export default Component.extend({
   init() {
     this._super(...arguments);
 
-    this.setProperties({
-      instructors: [],
-      instructorGroups: [],
-      learnerGroups: [],
-      startDate: moment().format('MM/DD/YYYY'),
-      endDate: moment().add(1, 'days').format('MM/DD/YYYY')
-    });
+    const instructors = [];
+    const instructorGroups = [];
+    const learnerGroups = [];
+    const startDate = new Date();
+    const endDate = new Date(startDate);
+    endDate.setDate(startDate.getDate() + 1);
+    const startTime = new Date().setHours(8, 0, 0, 0);
+    const endTime = new Date().setHours(9, 0, 0, 0);
+
+    this.setProperties({ instructors, instructorGroups, learnerGroups, startDate, endDate, startTime, endTime });
   },
 
   classNames: ['offering-editor'],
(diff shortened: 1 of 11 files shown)


app/components/offering-editor.js
```diff
@@ -1,23 +1,28 @@
 import Ember from 'ember';
 import DS from 'ember-data';
 import moment from 'moment';
+import { translationMacro as t } from "ember-i18n";
```
jrjohnson: the translationmacro doesn't work for us right now because when we change language it doesn't change.  This is fixed in the latest version, but that breaks some tests.  You can update the package if you want and fix the broken tests, or you can just use a computed property which depends on 'i18n.locale' and then call this.get('i18n').t() to do the translation.

app/components/offering-editor.js
```diff
@@ -91,6 +101,42 @@ export default Component.extend({
     }).sortBy('title');
   }),
 
+  datesValidated() {
+    const resetToday = new Date().setHours(0, 0, 0, 0);
+    const resetStartDate = this.get('startDate').setHours(0, 0, 0, 0);
+    const resetEndDate = this.get('endDate').setHours(0, 0, 0, 0);
+
+    let isStartDateInthePast = resetStartDate < resetToday;
```
jrjohnson: A start date in the past is allowed.

app/styles/components/_ember-cli-flash.scss
```diff
@@ -0,0 +1,9 @@
+$warning-color: #d9534f;
```
jrjohnson: There is already a flashmessages.scss this should go there.

jrjohnson: Actually this already seems to be there and this is duplicate styling.

saschaben:
so close! still seeing the "if I don't take an action/make a change on a time value, it reverts to 12am on save" issue on offering edit. Appears to be fully resolved on offering creation.

saschaben:
This looks good to go.

jrjohnson:
:+1:

PR #888, opened 2015-10-14
Fix publish / unpublish

### LINKED ISSUE #831: Remove administrator id from publish event serialization
As reported here:
https://github.com/ilios/ilios/issues/984

Publishing is broken due to the administrator id being sent with requests.  It is auto-generated on the API.

jrjohnson:
Remove un-postable fields from publish-event model.
Do not attempt to delete publish events with no related records.

Fixes #831

### DIFF
--- a/app/mixins/publishable.js
+++ b/app/mixins/publishable.js
@@ -52,10 +52,6 @@ export default Ember.Mixin.create({
         publishTarget.set('publishEvent', null);
         if(publishEvent){
           publishEvent.get(this.get('publishEventCollectionName')).removeObject(publishTarget);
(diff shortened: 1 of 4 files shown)


stopfstedt:
might as well remote the `totalRelated` property from `publish-event.js` while at it.

```
stefan@herrfuchs: ~/dev/projects/frontend on 831-publisheventsave
$ grep -r 'totalRelated' app
app/models/publish-event.js:  totalRelated: Ember.computed.sum('relatedCounts'),
```

stopfstedt:
otherwise this LGTM. :+1:

jrjohnson:
Made the small change.  If tests pass this should be good to merge.

PR #905, opened 2015-10-19
Publish all sessions

### LINKED ISSUE #797: Publish All / Publish All (as is) Functionality Missing From Courses And Sessions
In the current version of Ilios (2.4.8), there is a button to Publish All -- this is at the Course Level and it publishes all sessions and the course itself. There should also be a way to Publish All (as is) at this level.  This would present the user with the list of all the things that are missing from both the Course and the Sessions and assuming the user has reviewed this information allow for the Course and Sessions to be published as is.  The user should also have the ability to choose which to Publish as-is and which to leave as-is in Not Published mode.

In the 2.4.x, this is a button.  Maybe in the new version, it could be included in the drop-down menu at the course level.  Either way, we should scope this out and provide this functionality.

jrjohnson:
WIP

I’m starting a PR to get feedback on the overall functionality and display.  @saschaben can you take a look?

Steps to test -
1) Use http://56247d26d6865d4b8b000005.zoo-keeper-levels-51188.netlify.com/ as proxy wont work until #888 is merged
2) Pick a course with sessions and click the ugly green button in the session header
3) test, test, test

Questions I have:
Do we need this much information on this screen?
Would the info make more sense if we removed everything that was done and only displayed what is missing?
Do the Publish As Is / Mark As Scheduled checkboxes make sense?
Do the checkboxes in the table header to publish all as is and mark all as scheduled make sense?
I'm planning to collapse the unpublishable and publishable sections.  I think if they are collapsed they belong on top of the section where you can make publish / mark decisions.  Is that correct?
How can we make it clear that when you click save you are not only saving those sessions that you acted upon but also publishing everything in the publishable section?
Where should the save button go?  I can put more than one on the screen if it needs to be in a couple of places.
Other thoughts on function, other than the fact that it is pretty ugly right now?
Fixes #797


saschaben:
Questions I have:
Do we need this much information on this screen?
> yes. I don't think we can get away without showing it.

Would the info make more sense if we removed everything that was done and only displayed what is missing?

> interesting idea. I'm not sure. I suspect that we want to retain it though; remember to keep thinking about this as a "pre-flight checklist" rather than a "what's missing to get published" list. That brings more into focus the desire to see both the complete and incomplete data.

Do the Publish As Is / Mark As Scheduled checkboxes make sense?

> Yes, but I think we want some UX-ification for styling here. The first time around the block we spent forever going back and forth on nomenclature and presentation for publish - not published - scheduled. I don't want to do that again.

Do the checkboxes in the table header to publish all as is and mark all as scheduled make sense?

> I think so. but same note as above on UX-ification.
I'm planning to collapse the unpublishable and publishable sections. I think if they are collapsed they belong on top of the section where you can make publish / mark decisions. Is that correct?

How can we make it clear that when you click save you are not only saving those sessions that you acted upon but also publishing everything in the publishable section?

> This is one that needs input from Valentina, I think. If we can get some screenshot mocks of:
1. the way it is now
2. with the sections collapsed on top
3. with the sections collapsed on bottom
and get Dave to put them in front of Valentina, I think that would be ideal.

Where should the save button go? I can put more than one on the screen if it needs to be in a couple of places.

> I think at top & bottom of the list (the changeable list that is) should work.

Other thoughts on function, other than the fact that it is pretty ugly right now?

> if we are keeping the tool function dropdown control on each row of the session list, we should be adding a "publish session" option along with the "edit" option. I still am very much in favor of having this sort of control available...

saschaben:
one item I am noticing is that what displays as "published" in the session list is shown in the preflight checklist as unpublishable -- the checklist is correct. I am hoping this is just crap test data, rather than a problem on how we are reading the values...

dartajax:
publishallornot

Is it supposed to be right under "Add New"?

jrjohnson:
@saschaben I have no idea how to to make the boxes look better.  I can do the check all / un-check all as buttons.  But the individual lines I don't know what to use.  If I use monkey buttons they are way to big and a slider is really too small.  I was thinking about a switch with P and S, but I just don't think thats very clear.

Any ideas?

I don't have the tools to do a good mock, I will just put it together and we can move the sections around and take screen shots.

saschaben:
I missed one question earlier. Please see above, in bold.

saschaben:
@jrjohnson regarding the checkboxes: is there a monkey button control that stacks the buttons, rather than side by side? I would think that wouls take up only slightly more space than the boxes + labels do now....

jrjohnson:
We dint have anything. I can try and create something. Is it worth adding an extra development day to do?

saschaben:
@jrjohnson let's leave it be for now and we can add it as an enhancement.

jrjohnson:
Ok - I cleaned this up about as far as I can take it without feedback.  It needs tests, but I wanted to wait to nail down the functionality first.  The sections are now collapsible and I switched to buttons for the select all.

saschaben:
Seems like we're getting there. Functionality at first pass is good. We need serious work to provide better user context for this layout. We can discuss this morning and hopefully get some good input this pm from Valentina.

I'll review again a few times to make sure we are solid.

dartajax:
We met with Valentina today and together we have agree upon the following captions for this functionality.

Collapsable sections:

Sessions Imcomplete: cannot publish (x)
Sessions Complete: ready to publish (x)
Sessions Requiring Review (x)

The shorter buttons used in @jrjohnson's more recent mock seemed to be the ones we agreed upon ...

buttons

We agreed that the action button could be captioned simply "Save"

jrjohnson:
@saschaben / @dartajax thoughts on this?  Maybe a green button?
screen shot 2015-10-21 at 5 13 40 pm

saschaben:
Clever approach, @jrjohnson. I like it, but I suspect that feedback would tell us that this may confuse users initially --- since it doesn't "really" look like a button. I really like not requiring an additional step to display the info, but maybe we should build that review sentence adjacent to a "save now" button...?

dartajax:
That's cool but I agree with @saschaben that it looks to me like a label not a button.  The user would see the caption change if one of the check boxes is changed to "Publish As-Is" it would say Publish 1, Schedule 5, and Ignore 5 Sessions?  So they would see it dynamically change?  Or maybe that review sentence would change instead.

app/locales/en/translations.js
```diff
@@ -7,6 +7,7 @@ export default {
     'yes': 'Yes',
     'no': 'No',
     'on': 'On',
+    'go': 'Go',
```
jrjohnson: @saschaben / @dartajax please translate

app/locales/en/translations.js
```diff
@@ -293,6 +294,11 @@ export default {
     'unPublishProgram': 'UnPublish Program',
     'publishProgramYear': 'Publish Program Year',
     'unPublishProgramYear': 'UnPublish Program Year',
+    'incompleteSessions': 'Sessions Incomplete: cannot publish',
+    'completeSessions': 'Sessions Complete: ready to publish',
+    'reviewSessions': 'Sessions Requiring Review',
+    'all': 'Publish {{sessionCount}} sessions',
+    'publishAllConfirmation': 'Publish {{publishCount}}, schedule {{scheduleCount}}, and ignore {{ignoreCount}} sessions',
```
jrjohnson: @saschaben / @dartajax please translate

jrjohnson:
This needs translation, but other than that it should be ready to go.  It has the minimum possible tests.

app/locales/en/translations.js
saschaben: 'incompleteSessions': '"Séance incomplète: ne peut être pas publiée"
+    'completeSessions':  "Séance complète: prêt à publier"
+    'reviewSessions': "Séances à faire réviser"
+    'all':  "Publier {{count}} des séances"
+    'publishAllConfirmation': 'Publier {{publishCount}}, réserver {{scheduleCount}}, et ignorer des {{ignoreCount}} séances', "

thecoolestguy:
You probably want to wait for dave, but if you want to merge now and adjust later, I think this will do it:

'go': 'vaya',

'incompleteSessions': 'Sessiones Incompletados: No Poder Publicar',
    'completeSessions': 'Sessiones Completados: Listo a Publicar',
    'reviewSessions': 'Sessiones a Revisar',
    'all': 'Publique {{sessionCount}} sessiones',
    'publishAllConfirmation': 'Publique {{publishCount}}, Programe {{scheduleCount}}, y ignore {{ignoreCount}} sessiones',

PR #906, opened 2015-10-19
Improved Topic Management

### LINKED ISSUE #629: Fix topic manager to use a buffer
The topic manager currently interacts directly with the disciplines relationship on the course / session it is assigned to.  This is a bug because changes to the topics get shared around the application before they are saved.
Instead we need to use a buffer like the program year competencies manager.

### LINKED ISSUE #820: Don't preload course topics
Currently the course route is loading topics because when I built this I was younger and dumber (also injecting the store as a service wasn't yet a thing, ah youth).  Anyway, it makes the course load very slow and should be pretty easy to pull out and put in the detail-topic component.

https://github.com/ilios/frontend/blob/master/app/routes/course.js#L14

jrjohnson:
Course and Program Year now load much faster since the list of available topics is not loaded until it is needed.
Fixes #820

The topic manager now works with a buffer instead of on the subject directly.  So when you pick new topics, but do not save them they don’t show up other places on the page.

I also added a spinner to icon when topics are being saved.  I think we can use this on all the managers, see what you think.

Fixes #629

Review link: http://56249ae971e20a33b200001d.zoo-keeper-levels-51188.netlify.com

### DIFF
--- a/app/components/detail-topics.js
+++ b/app/components/detail-topics.js
@@ -22,9 +22,7 @@ export default Ember.Component.extend({
           defer.resolve(topics);
         });
       });
-    }
-    
-    if(this.get('isSession')){
+    } else if(this.get('isSession')){
       this.get('subject').get('course').then(course => {
         course.get('school').then(school => {
           school.get('topics').then(topics => {
@@ -32,9 +30,7 @@ export default Ember.Component.extend({
           });
         });
       });
-    }
-    
-    if(this.get('isProgramYear')){
+    } else if(this.get('isProgramYear')){
       this.get('subject').get('program').then(program => {
         program.get('school').then(school => {
           school.get('topics').then(topics => {
@@ -42,6 +38,8 @@ export default Ember.Component.extend({
           });
         });
       });
+    } else {
+      throw new Error("detail-topic component called without isCourse, isSession, or isProgramYear context");
     }
     
     return PromiseArray.create({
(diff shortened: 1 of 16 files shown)


saschaben:
looks fine. ready for code review

app/components/detail-topics.js
```diff
@@ -1,28 +1,85 @@
...
+    
+    if(this.get('isCourse')){
+      this.get('subject').get('school').then(school => {
+        school.get('topics').then(topics => {
+          defer.resolve(topics);
+        });
+      });
+    }
+    
+    if(this.get('isSession')){
```
stopfstedt: suggesting  `else if` here.

app/components/detail-topics.js
```diff
@@ -1,28 +1,85 @@
...
+      this.get('subject').get('course').then(course => {
+        course.get('school').then(school => {
+          school.get('topics').then(topics => {
+            defer.resolve(topics);
+          });
+        });
+      });
+    }
+    
+    if(this.get('isProgramYear')){
```
stopfstedt: likewise.

app/components/detail-topics.js
jrjohnson: Yea. I'm not sure how to do this right. Ideally I would probably create subclass components for each of these states but that seems like over kill. I felt like else if made it harder to understand because the blocks seemed related. Maybe I should stick a return inside each block so the next one isn't reached?  Or move the sets out into functions here?  Or I can just go else if, but it seems harder to read that way.

jrjohnson:
@stopfstedt - how does that look.  I added a thrown exception as well if none of them are set.

stopfstedt:
LGTM :+1:

stopfstedt:
@homu r+

PR #907, opened 2015-10-19
added form validations

djvoa12:
Notes:
- Replaced `inplace.js` mixin with `inplace-validation.js`
- Functionality stays the same with optional validation added
- Used highly-rated `ember-validations` addon from Dockyard
- Validation applicable for the following components:
  - `inplace-text`
  - `inplace-textarea`
- Refactored components:
  - `inplace-text`
  - `inplace-textarea`
  - `inplace-boolean`
  - `inplace-date`
  - `inplace-select`
- Edited few tests
- Validation test (acceptance)
- i18n validation error translations
  - Used `validation-i18n.js` from instance-initializer folder as `ember-cli-i18n` support is not yet available out of the box
  - Custom validators are located in `app/validators`


app/mixins/inplace2.js
```diff
@@ -0,0 +1,58 @@
+import Ember from 'ember';
```
jrjohnson: We're going to completely replace inplace, but in case we don't get it done how about we call this inplaceValidated?

djvoa12: `InPlace` mixin is used currently used in the components below: 
- `inplace-boolean`
- `inplace-date`
- `inplace-select`.

`InPlaceValidated` is used in two components now:
- `inplace-text`
- `inplace-textarea`

I will need to see if the code can be applicable to the three components at the top. Sometimes, one-for-all mixin can have issues as it doesn't provide much flexibility.

jrjohnson: I'm ok with splitting these up and removing the mixin entirely or partially. This might be a place where repeating yourself is a good thing.

jrjohnson:
I think the interaction and the display are very good here.   The only thing I would like is for the save checkmark to be disabled when there are errors.  I like only displaying the top error both visually and because putting a bunch of errors in the stack would be overwhelming.

I don't like having to define a Mixin for every field on the site.  Is there a way to specify these properties directly or does it have to be inherited?

In most cases the requirements are specified in the API docs:
https://ilios3-demo.ucsf.edu/api/doc

djvoa12:
@jrjohnson The save button already is disabled when there are validation errors. I am trying to figure out currently how to pass-in or specify the requirements another way (slightly blocked).

Also, I discovered couple bugs related to `inplace-date` and `inplace-select` while I was refactoring. We can talk about those later.

saschaben:
Validation rules:

External_id is an optional field, allow null - alphanumeric, Max length should match db field length, min length > 2.

Session Description / learning material description / learning material notes: an optional field, allow null - all chars acceptable, this is a memo field so maxlength isn't really an issue. No minimum length.

ILM session instruction hours: integer, allow fractional values and values <1, no negative values, numeric only, if available no null values (must be greater than 0).

Program title / Course title / session title: required, min length 3, all chars acceptable.

Program duration: whole number integer only, required, no nulls.

Program short title: optional, min > 1, max = db field length

Learning material author: required, min 2 char, alpha only

Learning material

tests/acceptance/program/overview-test.js
```diff
@@ -102,7 +102,7 @@ test('change duration', function(assert) {
 test('leave duration at 1', function(assert) {
   server.create('program', {
     school: 1,
-    duration: null,
+    duration: 1
```
djvoa12: @jrjohnson I had to edit few tests to make them pass. This particular one was a little tricky. I changed to the duration value because it displays `Click to edit` when duration is `null`. Let me know if this is alright.

jrjohnson: Thats ok since a value is required here you should also change the program creation to set a default of 1 for duration.  It looks like it was setup to not require a value, but that is wrong.

djvoa12:
@saschaben @dartajax Here's a list of validation errors that need to be translated:
- "is not included in the list"
- "is reserved"
- "is invalid"
- "doesn't match {{attribute}}"
- "must be accepted"
- "can't be empty"
- "can't be blank"
- "must be blank"
- "is too long (maximum is {{count}} characters)"
- "is too short (minimum is {{count}} characters)"
- "is the wrong length (should be {{count}} characters)"
- "is not a number"
- "must be an integer"
- "must be greater than {{count}}"
- "must be greater than or equal to {{count}}"
- "must be equal to {{count}}"
- "must be less than {{count}}"
- "must be less than or equal to {{count}}"
- "must be other than {{count}}"
- "must be odd"
- "must be even"
- "must be alphanumeric"

Note: You don't have to worry about any word inside `{{ ... }}`.

dartajax:
"is not included in the list" == "no está incluido en la lista"
"is reserved" == "es reservado"
"is invalid" == "es inválido"
"doesn't match {{attribute}}" == "no coincide con {{attribute}}"
"must be accepted" == "tiene que ser aceptado"
"can't be empty" == "no puede estar vacío"
"can't be blank" == "no puede estar en blanco"
"must be blank" == "tiene que estar en blanco"
"is too long (maximum is {{count}} characters)" == "es demasiado largo (el máximo es {{count}} caracteres"
"is too short (minimum is {{count}} characters)" == "es demasiado corto" (el mínimo es {{count}} caracteres"
"is the wrong length (should be {{count}} characters)" == "es la longitud incorrecta (debe ser {{count}} caracteres"
"is not a number" == "no es un número"
"must be an integer" == "debe ser un numero entero"
"must be greater than {{count}}" == "debe ser mayor que {{count}}"
"must be greater than or equal to {{count}}" == "debe ser mayor que o igual a {{count}}"
"must be equal to {{count}}" == "debe ser igual a {{count}}"
"must be less than {{count}}" == "debe ser menos que {{count}}"
"must be less than or equal to {{count}}" == "debe ser menos que o igual a {{count}}"
"must be other than {{count}}" == "Debe ser otra cosa que {{count}}"
"must be odd" == "debe ser número impar"
"must be even" == "debe ser número par"
"must be alphanumeric" == "debe ser alfanumérico"

app/instance-initializers/validation-i18n.js
```diff
@@ -0,0 +1,28 @@
+import Ember from 'ember';
```
jrjohnson: This needs a comment referencing https://github.com/dockyard/ember-validations/issues/224 so we can see where it is coming from and when it can be removed.

app/styles/components/_inplace-text.scss
```diff
@@ -0,0 +1,14 @@
+.inplace-text {
```
jrjohnson: This should be in _editinplace.scss since it is common styling for all inplace elements.

app/styles/components/_inplace-textarea.scss
```diff
@@ -0,0 +1,14 @@
+.inplace-textarea {
```
jrjohnson: This should be in _editinplace.scss since it is common styling for all inplace elements.

app/components/course-overview.js
```diff
@@ -29,6 +29,14 @@ export default Ember.Component.extend({
       promise: deferred.promise
     });
   }.property(),
+
+  externalIdValidations: {
+    'buffer': {
```
jrjohnson: I'm not a huge fan of buffer here.  Maybe we can use a oneWay computed value on the mixin and then instead of buffer we can use userInput here?  Don't need to make the change in this version, but think about it for when we create a lot more validators.

app/components/inplace-boolean.js
```diff
@@ -1,14 +1,19 @@
 import Ember from 'ember';
-import InPlace from 'ilios/mixins/inplace';
+import InPlaceValidation from 'ilios/mixins/inplace-validation';
```
jrjohnson: What does validation of a boolean do?  I guess we could prevent it from being false, but that seems like it wouldn't ever happen.

jrjohnson: If it is just so everything is using the same mixin then you should remove the plain inplace and just keep the inplace-validation.

app/instance-initializers/validation-i18n.js
djvoa12: Done.

app/styles/components/_inplace-text.scss
djvoa12: Done.

app/styles/components/_inplace-textarea.scss
djvoa12: Done.

app/components/inplace-boolean.js
djvoa12: In case of `inplace-boolean` component, `inplace-validation` mixin doesn't provide much value other than saving the new value. Should I have `inplace` or `inplace-validation` here? They both work.

app/components/course-overview.js
djvoa12: The validation criteria, such as `externalIdValidations` need to be passed into any `inplace-{name}` to work. In all five components, it's watching the `buffer` to validate it. Maybe we can talk more about it in person on Monday? I am not how to use `oneWay` on the mixin in this case.

jrjohnson:
:+1: @homu r+

PR #909, opened 2015-10-19
select options update on subgroup creation

### LINKED ISSUE #730: Learner Group subgroup additions are not immediately added to select lists
If a sub-group is added or removed, that change is not reflected in the drop down selection lists for user management without a hard page refresh (boo.)

djvoa12:
Select options now updates themselves on sub-group creation. No more page refreshing necessary.

Fixes #730


jrjohnson:
:+1:

PR #910, opened 2015-10-21
fixed back-to-courses link bug

### LINKED ISSUE #377: Course and Sessions >> Navigation Question >> Back to Courses List Disappears after Selecting Session
I noticed that the link "Back to Courses List" is there when the user shows up to Course and Sessions detail initially but disappears after a Session has been pulled up.

link1
link2
link3

djvoa12:
Fixes #377

### DIFF
--- a/app/controllers/course.js
+++ b/app/controllers/course.js
@@ -1,14 +1,19 @@
 import Ember from 'ember';
 
-export default Ember.Controller.extend({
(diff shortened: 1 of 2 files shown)


jrjohnson:
Interesting.  I wonder why we had it in an if block.  I'm going to go ahead and approve and we can figure out if that was an artifact or a requiremnet.
:+1:

Issue #848, opened 2015-10-02
dashboard: my courses list is pulling the wrong courses


saschaben:
As far as I can tell, the My Courses list is not pulling the correct data. On demo, demofaculty3 is in medicine, and is teaching in BMB, but My Courses appears to be pulling only older Dentistry courses.

@jrjohnson, please move to /Ilios if this is an API issue.

PR #914, opened 2015-10-21
Refactor courses list

jrjohnson:
The courses list is now more reliable and faster as we’re not doing a
bunch of work in the model hook, but in the controller instead.

I misread #848 and did this by mistake.


jrjohnson:
This is ready for review, it's kind of a big change, sorry.

stopfstedt:
:+1: LGTM

PR #918, opened 2015-10-22
stack trace to the console

### LINKED ISSUE #913: Add the stack trace to the console when we encounter an error
Currently we're only getting the error message, which makes it hard to debug.  Instead we need to figure out how to get the whole error object with stack trace into the console.  This allows the developer to click on a file in the stack trace and it will open to that line in the course tab of the developer tools.

djvoa12:
@jrjohnson Let me know if this solves the issue. I made it so it logs the error object itself and the stack trace. If not, I may need to use [`console.trace()`](https://developer.mozilla.org/en-US/docs/Web/API/Console/trace).

Fixes #913

### DIFF
--- a/app/initializers/error.js
+++ b/app/initializers/error.js
@@ -20,6 +20,7 @@ export default {
       // Global error handler in Ember run loop
       Ember.onerror = (error) => {
         console.log(error);
+        console.error(error.stack);
 
         if (error) {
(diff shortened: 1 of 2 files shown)


jrjohnson:
:+1: I'm going to merge and we can test as we go.

Issue #899, opened 2015-10-15
flesh out MeSH picker details


stopfstedt:
for comparison, here a screenshot of the V2 MeSH picker:

![selection_014](https://cloud.githubusercontent.com/assets/1410427/10529997/c5a1fc12-7357-11e5-9c6b-eea2ab8e5260.png)

display the same detail of information in V3.

stopfstedt:
@saschaben - getting there.

![selection_017](https://cloud.githubusercontent.com/assets/1410427/10651942/64e3434a-7807-11e5-8cb1-0fadae390dd0.png)

- duplication issue in scopenotes seems to be fixed.
- i'm expanding the search box width to claim more space.
- truncation of scope notes via CSS.

PR #919, opened 2015-10-22
show more details in MeSH picker

stopfstedt:
refs #899

### DIFF
--- a/app/mirage/config.js
+++ b/app/mirage/config.js
@@ -178,6 +178,13 @@ export default function() {
     this.delete('/api/meshqualifiers/:id', 'meshQualifier');
     this.post('/api/meshqualifiers', 'meshQualifier');
 
+    this.get('/api/meshtrees', getAll);
+    this.get('/api/meshtrees/:id', 'meshTree');
+    this.put('/api/meshtrees/:id', 'meshTree');
+    this.delete('/api/meshtrees/:id', 'meshTree');
+    this.post('/api/meshtrees', 'meshTree');
+
+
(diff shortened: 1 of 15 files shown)


stopfstedt:
selection_018

saschaben:
this is looking really really good. We should probably think about how to deal with the scope notes and the current ellipsis cutoff --- at this stage, there's no way to expand the note fully -- but the current responsiveness and data display is awesome as is.

stopfstedt:
@saschaben thanks for the feedback. would displaying the full scope notes be an option?

meanwhile i fixed all regressions in existing tests that were introduced by these markup changes. 

still need to put the newly displayed data points under test coverage. but really close to being done now.

saschaben:
@stopfstedt I'm not sure we want to allow full scope notes without limit, as I'm not sure how long they might turn out to be. But perhaps a longer string prior to truncation; right now you have it set to truncate at ~65 chars. Why not extend that to 250 and call it a day. Anything over that is prolly too much for a body to pay attention to anyway. we can revisit later.

stopfstedt:
@jrjohnson please review. @saschaben see follow-up #939. thanks!

jrjohnson:
@homu r+

jrjohnson:
Moved ahead as it looks like functional review already happened.

saschaben:
@stopfstedt does this ticket override #899, or is that still a separate issue?

stopfstedt:
@saschaben :this fixes #899

Issue #860, opened 2015-10-06
Currently Applied Filter Needs To Have an Indication on the Dashboard


dartajax:
If a user is applying a filter to the calendar and closes the filters, there needs to be an indicator of what is being filtered.

I guess this relates to #858 in the sense that not only do we need to see what is being filtered (this issue), we also need to provide the functionality to have the filter(s) removed completely (that issue).

I am not sure what kind of mock I can come up with but how about this ...

Step 1 >> Show Filters and select several.

![step1](https://cloud.githubusercontent.com/assets/7493331/10318585/0b22f060-6c1f-11e5-8d69-7ec1ff301c4f.JPG)

Step 2 >> Hide the Filters 

![step2](https://cloud.githubusercontent.com/assets/7493331/10318745/c29bd5e0-6c1f-11e5-9dc5-542cabf4a821.JPG)

When the filters are hidden and the calendar below has been filtered, there needs to be an indicator of what filter has been applied.  This is ugly and not ideal but provides the idea that something should be done to indicate the applied filters.

PR #929, opened 2015-10-24
added clear filters

### LINKED ISSUE #858: Clear All (for Filters) Needed on Dashboard
When filters have been activated and selected, there should be link or button of some sort to remove the filters completely and show everything.

Ugly mock of the strip (buttons adjusted as well) below ....\

newstrip2

djvoa12:
Ready for review and merge when approved.

Note: My editor removed lots of spaces that shouldn't be there.

Fixes #858

### DIFF
--- a/app/components/dashboard-calendar.js
+++ b/app/components/dashboard-calendar.js
@@ -3,6 +3,8 @@ import Ember from 'ember';
 import DS from 'ember-data';
 import momentFormat from 'ember-moment/computeds/format';
 
+const { computed, isPresent } = Ember;
+
 export default Ember.Component.extend({
   userEvents: Ember.inject.service(),
   schoolEvents: Ember.inject.service(),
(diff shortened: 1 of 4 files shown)


saschaben:
sweet.

jrjohnson:
I would rather see this as a list of applied filters with an x to clear them all at the end.  This seems to be a pretty common pattern for this type of thing.

saschaben:
@jrjohnson that is the existing 2.4.8 pattern --- definitely would be preferable, but requires some UI thought regarding the display of the filter list. This as is could be an interim solution while we hash out the display question. That filter list can get really long. In the current 2.4.8 iteration it wound up a less than ideal compromise between available space, legibility, and code constraint. Would love to find a better answer this time....

djvoa12:
That's another PR I am going to create to fixes issue #860.

jrjohnson:
:+1:

PR #934, opened 2015-10-24
Fix Dashboard My Courses

### LINKED ISSUE #848: dashboard: my courses list is pulling the wrong courses

jrjohnson:
The list of courses was getting rendered before all promises were resolved.  I refactored the property which aggregates a users courses and updated the display of the My Courses Dashboard component to include a loading spinner.

Fixes #848

### DIFF
--- a/app/components/dashboard-mycourses.js
+++ b/app/components/dashboard-mycourses.js
@@ -5,6 +5,8 @@ export default Ember.Component.extend({
   currentUser: Ember.inject.service(),
   tagName: 'div',
   classNames: ['dashboard-block'],
+  courseSorting: ['startDate:desc'],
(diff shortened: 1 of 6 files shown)


jrjohnson:
@saschaben Demo faculty 3 wasn't attached to BMB in my demo DB.  Maybe that changed?  Or possible just a database timing issue.  I think this is now fixed, but I wasn't able to validate the specific case you had in #848

saschaben:
as far as I can see, this is solve the issue, and is full of awesome goodness. 

code review.

PR #960, opened 2015-10-27
don't save user when assigning course director.

stopfstedt:
this fixes yet another 403/_verboten_ causing bug.


jrjohnson:
:+1: no user facing changes
@homu r+

PR #962, opened 2015-10-28
learner group management: on assignment save the group, not the instructor.

stopfstedt:
this was causing 403s to be thrown around. bugfix.


jrjohnson:
:+1: no user facing changes
@homu r+

PR #964, opened 2015-10-28
create, update, delete program-year

### LINKED ISSUE #954: unable to add new program year
attempting to add new program year generates a 400 error:
addyearerr

in one instance, it sent a partial save -- leaving a year without a cohort:

addyearerr2

### LINKED ISSUE #955: unable to edit program year
generates error: `<ilios@component:programyear-list::ember2039> had no action handler for: edit`

### LINKED ISSUE #956: no control to delete program year
oversight. This will be an uncommon but necessary admin function. will be for superadmin permission only.

until we can address it correctly we will deal with this manually.

djvoa12:
- [x]  Create new program-year (Fixes #954)
  - Added constraints
  - If one or more programs years existed already, it takes a specs from the latest one
- [x] Update an existing program-year (Fixes #955)
  - Edit icon works now
- [x] Delete an existing program-year (Fixes #956)
  - Confirmation for removal
- [x] Tests
  - Unit
  - Acceptance


saschaben:
Looking really nice!

A couple functional review comments:

1. while we don't need a program year delete confirmation immediately, we should add a ticket to make one. It will follow form of the program delete confirmation.
2. when a program has no program years, we run into the same dropdown control issue we have everywhere: there is not enough room at the bottom of the container div to show the selections. This is a currently insoluble z-index issue, but the simple hack-around it is to make sure to provide enough blank screen space to let the control list display. we can add some space above and below the "no years available" text message and we'd be done with this one for now.

3. more generally if we don't already have a ticket to increase the spacing between the edit icon and the delete icon on program year (and globally), let's make sure to get one in.

djvoa12:
1) Confirmation is now included
2) Done
3) Done (also properly spaced in this PR)

@saschaben

saschaben:
:+1:

app/styles/components/_resultslist.scss
```diff
@@ -1,4 +1,9 @@
 // scss-lint:disable SelectorDepth NestingDepth
+
+$bootstrap-success1: #dff0d8;
```
jrjohnson: Use the style for course saving here instead of a new pattern.

jrjohnson:
@saschaben please review function again.  Slight adjustment to the confirmation when a new program year is saved.  Should work proxied against demo.

saschaben:
proxied, this fails on new program year creation with 2 

`Error: Adapter operation failed
    at new Error (native)`

errors

app/templates/components/programyear-list.hbs
```diff
@@ -74,17 +77,30 @@
...
+                  {{#link-to 'programYear.index' program programYear class="program-year-link"}}{{fa-icon 'edit'}}{{/link-to}}
+                  <span class="clickable trash {{if programYear.showRemoveConfirmation 'confirm-removal'}}" {{action 'remove' programYear}}>{{fa-icon 'trash'}}</span>
                 </td>
               </tr>
+
+              {{#if programYear.showRemoveConfirmation}}
+                <tr class='confirm-removal'>
+                  <td colspan=8>
+                    <div class='confirm-message'>
+                      {{t 'programs.confirmRemove' courseCount=program.courseCount}} <br />
```
jrjohnson: I think this is the wrong removal confirmation.  Shouldn't it be programYears.confirmRemove?

djvoa12: Yup. Good catch!

saschaben:
:+1:

saschaben:
@dartajax please merge asasp

PR #967, opened 2015-10-29
upgraded ui-select to v0.0.10

djvoa12:


jrjohnson:
@saschaben can you make sure that adding users to learner groups and bulk editing still works and star?

saschaben:
adding users appears to function correctly.

bulk edit has display error in control (below)
err2

my initial bulk test threw a bunch of errors, but I can't reproduce so I think it was cache.

djvoa12:
You have to `npm install` to get the latest version.

saschaben:
ah. got it.

Everything looks good now (after `npm install`, duh.) One minor fix: the label on the bulk edit control should be plural:  "Remove/Switch _learners_ to..." not learner. Even for a single selection, the implication should be "move all the selected _items_ to somewhere else".

saschaben:
@djvoa12 if you can tweak that label, quickly, let's go ahead -- otherwise let's add an enhancement ticket and I'll star this one.

djvoa12:
jesus finally green

saschaben:
:+1:

jrjohnson:
@homu r+

jrjohnson:
@homu retry

PR #971, opened 2015-10-29
added css id's for all toggle switches on dashboard page for selenium testing

thecoolestguy:


jrjohnson:
You should set some kind of default value for CSS id on the component, maybe a random string.  Otherwise if you don't pass in cssId you will end up with several elements with id='-input' which is probably not valid and also would result in a duplicate ID.

thecoolestguy:
Setting up a default cssId makes alot of sense! Thanks! I will do that.

thecoolestguy:
Closing to redo in different branch.

PR #976, opened 2015-10-30
revised error handling process

djvoa12:
Refactored and added additional control flow to reduce bugs caused by the error handlers.


jrjohnson:
:+1: no user facing change

Issue #779, opened 2015-07-13
Logout Users when they close their browser


jrjohnson:
Currently because we store our authentication token in LocalStorage users stay logged in until the intentionally logout.  Closing the browser window should cause a logout.

The easiest path is to store the token in a cookie instead of LocalStorage and to set the cookie to expire at the end of the session.

Looks like we can do this with a simple config change, will need to investigate further to ensure this is compatible with our JWT solution:
https://github.com/simplabs/ember-simple-auth/tree/master/packages/ember-simple-auth-cookie-store

saschaben:
This includes switching of permissions from one value to another  (see #1072)

saschaben:
I am suspecting that this is actually having an impact on usability in a number of areas. escalating.

jrjohnson:
This was resolved for the app in #1289 - but it is entirely possible that we are having authentication issues with the embedded app since there is no logout option there.

PR #981, opened 2015-11-03
Add Authorization and limit access to navigation items

jrjohnson:
Authorize basic user actions.  Right now we are just hiding the items from the menu which will prevent most problems.  This is not for security, its for ensuring that the user doesn't get annoyed by having access to things they can't actually do.

Can be tested here:
http://5638561571e20a1a4200001c.ilios-frontend.netlify.com/

### DIFF
--- a/app/abilities/admin-dashboard.js
+++ b/app/abilities/admin-dashboard.js
@@ -1,0 +1,11 @@
+import Ember from 'ember';
+import { Ability } from 'ember-can';
+
+const { inject, computed } = Ember;
+const { service } = inject;
+const { alias } = computed;
+
+export default Ability.extend({
+  currentUser: service(),
+  canView: alias('currentUser.canViewAdminDashboard'),
+});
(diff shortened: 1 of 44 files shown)


jrjohnson:
@saschaben this is ready for review agains the demo API.

saschaben:
will review as soon as demo is rebuilt

saschaben:
Nav bar as built for a given user/privilege model does not get flushed from cache when logging out / re-logging in as a new user with differing permissions.

log in as demo_dev1. you get full access. log out, re log in as demo_faculty3; full access is still available. full page refresh clears and correctly rebuilds nav selections.

jrjohnson:
That seems like an edge case maybe a new ticket?

saschaben:
per discussion , this is due to ticket #779.

we will merge and address this gaping security hole with that ticket.

saschaben:
:+1:

thecoolestguy:
@homu r+

jrjohnson:
@homu retry

jrjohnson:
Homu is unhappy, merging anyway.

Issue #944, opened 2015-10-26
Get Rid Of School (Medicine) Box on Dashboard when "My Schedule" is Displayed


dartajax:
This is because if a user's schedule is displayed, it may contain data (offerings) from more than one school (Medicine and Pharmacy for example).

![notneeded1](https://cloud.githubusercontent.com/assets/7493331/10745560/37bb9a46-7c00-11e5-826a-3b5479a11f5e.JPG)

PR #1024, opened 2015-11-04
Handle users who are in multiple schools

### LINKED ISSUE #1014: Course List year selector throws error
1. select a given year for the course page, and allow the list to load;
2. re-select the same, currently displayed year from the year select list;

> Result: `Cannot read property 'get' of undefined` error is thrown.

jrjohnson:
This affects lists for courses, learner groups and instructors groups as well as the school events on the dashboard.

Currently API endpoint for permissions is not available so this has to be tested with:
http://563a7ef171e20a1a05000009.ilios-frontend.netlify.com
demo / demo login is now a user in two schools.

I ended up having to refactor instructorGroups completely because it wasn't working so instead of spending time fixing an old pattern I updated it to be the same as courses.
Fixes #1014 #944

### DIFF
--- a/app/components/dashboard-calendar.js
+++ b/app/components/dashboard-calendar.js
@@ -3,7 +3,7 @@ import Ember from 'ember';
 import DS from 'ember-data';
 import momentFormat from 'ember-moment/computeds/format';
 
-const { computed, isPresent, RSVP, ObjectProxy } = Ember;
+const { computed, isPresent, RSVP } = Ember;
 const { PromiseObject } = DS;
 
 export default Ember.Component.extend({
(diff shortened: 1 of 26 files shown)


saschaben:
Looks good overall. Failling points:

Programs: switching schools causes error: `Mirage: Your Ember app tried to POST '/errors', but there was no route defined to handle this request. Define a route that matches this path in your mirage/config.js file.`

Calendar: switching schools with filters on (set to topic/detail) causes error: `n.getCohortsForYear is not a function` & `TypeError: Cannot read property 'filter' of undefined`

Courses & Instructor groups: adding a new course or instructor group and then switching schools should probably clear the "saved successfully" message.

jrjohnson:
Damn.  I forgot programs entirely.  Good catch on the filters.  Will address both.

I went back and forth on the save message and ended up picking the easiest path.  I don't know what the correct way to go here, would it be more confusing if it went away?

saschaben:
@jrjohnson I think we leave the save message as is for now, and address this as a UX enhancement issue later. I suspect handling it best for the users will require some backflips and hocus pocus. Don't need to deal with that now.

jrjohnson:

saschaben:
same netlify build to test? or new URL?

jrjohnson:
New one.  I updated the URL above.

saschaben:
looks good. can't break it. :+1:

PR #1037, opened 2015-11-05
More details in cohort manager

### LINKED ISSUE #1033: Badges in Manage Cohorts Need Indication of School and Program
Badges in Manage Cohorts don’t indicate where they come from (which school and year).

This will be increasingly more important as multi-school curricular activities become more common down the road.

Here you can see everything ...
image

Click to edit (Manage Cohorts) though and get this ...

image

It is possible to get two (or three) of the same badges up there and not know which one should be removed ...

image

jrjohnson:
- Added the school and program to the cohort badge
- Added the school to the list of cohorts
- Added a nice save spinner

Fixes #1033

Review at http://563ae5ded6865d7bd0000023.ilios-frontend.netlify.com

### DIFF
--- a/app/components/detail-cohorts.js
+++ b/app/components/detail-cohorts.js
@@ -6,6 +6,7 @@ export default Ember.Component.extend({
   course: null,
   cohorts: Ember.computed.oneWay('course.cohorts'),
   isManaging: false,
+  isSaving: false,
   previousCohorts: [],
   classNames: ['detail-cohorts'],
   programs: function(){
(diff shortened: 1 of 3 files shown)


saschaben:
@jrjohnson is there any way we can get the program/cohorts to sort by owning school? Right now they are randomly ordered, which is a little weird from a user perspective. If it isn't easy, then let's add a separate ticket and address it after launch. But if it is a simple fix to order the cohort/program groups, then let's cram that in. This is a solid fix for the release,

jrjohnson:
Unfortunately not simple.  It is doable, but will probably take  bit of time to figure out.

saschaben:
ok. will add a followup issue. 

:+1:

PR #1040, opened 2015-11-05
Changed cssId to computed property to ensure randomness between elements

thecoolestguy:


jrjohnson:
Does this work?  I didn't know it was possible to replace a computed property with a value. Sweet.

jrjohnson:
:+1: no user facing change

PR #1043, opened 2015-11-05
changed to calendar description to "my calendar"

### LINKED ISSUE #850: Simplify Dashboard Label
Currently the Label on the Dashboard reads ....

With the Calendar off (when logged in as Demo Faculty4) ...

"Demo Faculty4's Calendar Monday October 5th, 2015"

With the Calendar on, it displays ....

"Demo Faculty4's Calendar"

This should be simplified (in either case) to "My Schedule" since the user knows they are logged in (can be verified in the upper right corner) and they presumably know today's date or it is visible on their computer.

djvoa12:
Fixes #850


saschaben:
:+1:

PR #1053, opened 2015-11-06
ILM Details

### LINKED ISSUE #883: provide event detail for ILM
ILMs are not currently getting processed correctly for event detail display. We need to add this.

ILMs have the following changes in display:

1: "On [Date value], [time value]" changes to "Due by [Date value], [[time value]]"
> > "Due by October 31, 2015, 5:00pm"
2: Session type displays the same as now, but should be appended with a tag of "__(ILM)__" 
> > This session is "Self-Directed Learning __(ILM)__"
3: The instructor list will typically be empty on ILMs

jrjohnson:
I fixed basic ILM display on the detail page.  We will probably need to refine this, but it does at least show up now without breaking.

I also made the school events / my events slider option a query param so it sticks on refresh and back button use.

Fixes #883

### DIFF
--- a/app/components/dashboard-calendar.js
+++ b/app/components/dashboard-calendar.js
@@ -3,15 +3,16 @@ import Ember from 'ember';
 import DS from 'ember-data';
 import momentFormat from 'ember-moment/computeds/format';
 
-const { computed, isPresent, RSVP } = Ember;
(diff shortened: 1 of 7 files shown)


saschaben:
:+1:

Issue #977, opened 2015-10-30
Calendar Feed Functionality Needed on The Dashboard


dartajax:
We need to provide the functionality contained in Ilios 2.4.x for the students to be able to generate and download the URL that they can paste into the mobile devices to set up their calendar feed.

For reference, here is how it looks in 2.4.8 ...

![feed](https://cloud.githubusercontent.com/assets/7493331/10860523/ac02e598-7f26-11e5-809b-6a5eadbdda37.JPG)

I will check the mocks for this as well to see if it has been scoped out for 3.0.

jrjohnson:
Fixed by https://github.com/ilios/calendar/pull/6 we will need to release a new version and then update our dependancy.

Issue #930, opened 2015-10-24
Calendar offerings for a given session with same time block should aggregate


saschaben:
per existing functionality, if you have multiple offerings for a single session which occur at EXACTLY the same time block (same start and end datetime values), they should display in the week and month views as a single block, with the room and instructor values listed as "multiple". Day view displays all offerings as unique events with all detail. Clicking on an aggregated block like this will take a viewer to the day view, at which point they can click an individual offering and get to the detail event page. (see offerings manager in 2.4.8 for further example).

as it stands, we are not aggregating these displays. This is only really an issue for course directors, since neither students nor regular instructors will ever normally be attached to multiple events at the same time.

![blockerr](https://cloud.githubusercontent.com/assets/1753431/10708682/9dfe3de6-79c5-11e5-9d31-4f8b90137079.png)

While this isn't strictly a bug, I am labeling at such since it is existing behavior and expected by the users.

Also, I'm assuming the is a frontend (and not a calendar) issue, but that's very likely a wrong assumption.....

PR #1058, opened 2015-11-06
Calendar Updates

### LINKED ISSUE #941: Calendar Day View cuts off at 4pm
This disallows the view of any activity 5pm or on. Need to add a vertical scroll just as with week view. or make overflow visible. or somethin'.

jrjohnson:
Fixes #941 #977 #930

### DIFF
--- a/app/components/dashboard-calendar.js
+++ b/app/components/dashboard-calendar.js
@@ -34,6 +34,9 @@ export default Ember.Component.extend({
   loadingEventsTranslation: Ember.computed('i18n.locale', function(){
     return this.get('i18n').t('calendar.loadingEvents');
   }),
+  icsInstructionsTranslation: computed('i18n.locale', function(){
+    return this.get('i18n').t('calendar.icsInstructions');
+  }),
   setup: Ember.on('init', function() {
     //do these on setup otherwise tests were failing because
     //the old filter value hung around
--- a/app/locales/en/translations.js
+++ b/app/locales/en/translations.js
@@ -331,7 +331,8 @@ export default {
     'sessionObjectives': 'Session Objectives',
     'sessionLearningMaterials': 'Session Learning Materials',
     'noAssociatedCompetencies': 'No Associated Competencies',
-    'dueThisDay': 'Due this day'
+    'dueThisDay': 'Due this day',
+    'icsInstructions': 'To add your Ilios calendar to another application or service, use this URL.  This URL is like a password. Anyone who knows it can view your calendar!'
   },
   'auth': {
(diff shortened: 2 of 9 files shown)


jrjohnson:
@saschaben this is ready to review and can be tested against demo or stage.

saschaben:
on show all calendar for november 8,9,10, getting the `TypeError: Cannot read property 'totalLevels' of undefined` error. nothing shows in network tab; think this may be either bad data or we are pulling a deleted event....?

saschaben:
on the upside: ics thing seems to be perfect. ILM views now appear to be working (though we are indeed setting them to 4pm rather than our previous 5pm, not a big deal but weird); and everything else is pretty much passing muster as far as I can tell.

jrjohnson:
- 1 Existing issues caused by the Valour Lunch course which goes from August to June.  (#1068)
- 2 We've never had the user events on the show all calendar.  My solution would be to rename that switch to 'School Events' - does that work?  Or do we need to combine the user and school events onto the same calendar?
- 3 I'm having trouble reproducing this.  Might it be related to 1 (in that the bug there broke going back)?
- 4 new ticket #1069

saschaben:
- 1 Existing issues caused by the Valour Lunch course which goes from August to June.  (#1068)
> - OK. I missed that one. Is there a possible fix for this (otherwise we're gonna very quickly have some negative feedback from our VALOR students & faculty)

- 2 We've never had the user events on the show all calendar.  My solution would be to rename that switch to 'School Events' - does that work?  Or do we need to combine the user and school events onto the same calendar?
> - Wow. Never mind. I just wrote a big o'l response to this but as I stepped thru to validate the use case to reproduce,  I found that the error is actually an artifact of the bug reported in #1. The correct items show if the display is not busted by the multiday event (in which case NOTHING shows. Suckity suckity.)

- 3 I'm having trouble reproducing this.  Might it be related to 1 (in that the bug there broke going back)?
> -  I don't think so. To reproduce: proxy to demo3. login as demofaculty3. open calendar to show all events. toggle to view Pharmacy. click on Monday 11/9 "Quality  Assurance/Improvement" 10am. Hit back button in the browser (or keyboard). dahsboard will load the calendar for Medicine.

- 4 new ticket #1069 
> - Yay. Boo.

jrjohnson:
AH!  The School!!  I was thinking the all evens slider.  I have no idea why I couldn't make that connection.  Yea.  School isn't sticky.  It really should be.  I created a new issue for this #1070

saschaben:
:+1:

PR #1060, opened 2015-11-07
fixed am/pm flip bug

### LINKED ISSUE #1034: PM Switches Back to AM When End Time Hour Picker is Used >> Add Offering
The AM/PM picker switches back to AM when the End Time Hour is Updated.  Obviously, if it already AM, there is no problem but if PM, it messes up the offering's start and end time.

This session goes from 11:00 AM - 1:00 PM so first I will change the start time.
image

Now change the end time AM/PM to PM.
image

Now change the end time hour to 1 and notice that without doing anything, the PM switched to AM.
image

NOTE:  This only happens when the end time hour is changed to a different end time hour than it was previously.

djvoa12:
Fixes #1034

### DIFF
--- a/app/components/offering-editor.js
+++ b/app/components/offering-editor.js
@@ -165,21 +165,27 @@ export default Component.extend({
       this.set('isMultiDay', !this.get('isMultiDay'));
     },
 
-    changeStartTime(date) {
-      let newStart = moment(date);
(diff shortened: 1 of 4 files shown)


saschaben:
:+1:

saschaben:
:+1:

PR #1081, opened 2015-11-10
added small group validation

### LINKED ISSUE #1067: small group creator fails silently if no group selected
if no groups are selected for the offerings generator, and it is set on the default of small groups, the save operation fails silently.

the only two viable cases are either we automatically save any action without groups as a single offering, or we require groups to be added when small group is selected. In the case of the 2nd option, the validation message would have to read: "Please select at least one learner group to attach to your small group offering. If you wish to schedule this offering without groups, please select the "offering" button above."

fr translation for option two, just in case we need to go that route: "S'il vous plaît choisir au moins un groupe d'apprenants à joindre à votre offrande de "small groups". Si vous souhaitez programmer cette offrande sans groupes, s'il vous plaît choisir le bouton "offre" au-dessus ".

djvoa12:
Fixes #1067

### DIFF
--- a/app/components/offering-editor.js
+++ b/app/components/offering-editor.js
@@ -224,30 +224,37 @@ export default Component.extend({
     },
 
     create() {
-      if (this.datesValidated() && this.timesValidated()) {
-        let datesHash = this.calculateDateTimes();
-        let learnerGroups = this.getAllLearnerGroups();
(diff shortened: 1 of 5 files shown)


saschaben:
:+1:

dartajax:
Can I merge this one?

jrjohnson:
@homu r+

PR #1087, opened 2015-11-10
Put the error data into a data key

jrjohnson:
This makes it a little bit easier to deal with in the API.

Refs ilios/ilios#1063


jrjohnson:
:+1: no user facing change

PR #1088, opened 2015-11-10
Default lm::link attribute to 'http://'.

### LINKED ISSUE #1013: learning material weblinks require "http://" prefix
if the user does not provide the `http://` prefix to a web address, adding the learning material weblink will fail. This is an onerous action to put on the user.

Solution: if possible, we should provide the `http://` as the default value on focus of the cursor in the address form field; if that seems too complex and hoop-jumpery, we should simply allow un-validated web addresses to be added.

stopfstedt:
fixes #1013

selection_028

### DIFF
--- a/app/models/learning-material.js
+++ b/app/models/learning-material.js
@@ -19,7 +19,7 @@ export default DS.Model.extend({
   filename: DS.attr('string'),
   mimetype: DS.attr('string'),
   filesize: DS.attr('number'),
-  link: DS.attr('string', {defaultValue: 'http://'}),
+  link: DS.attr('string'),
   absoluteFileUri: DS.attr('string'),
   type: computed('filename', 'citation', 'link', function(){
     if (this.get('filename')) {
--- a/app/models/learning-material.js
+++ b/app/models/learning-material.js
@@ -45,13 +45,13 @@ export default DS.Model.extend({
     }
     
   }),
-  isFile: computed('link', 'citation', 'absoluteFileUri', function() {
+  isFile: computed('type', function() {
     return (this.get('type') === 'file');
   }),
-  isLink: computed('link', 'citation', 'absoluteFileUri', function() {
+  isLink: computed('type', function() {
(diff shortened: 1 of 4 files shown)


saschaben:
:+1:

app/models/learning-material.js
```diff
@@ -45,4 +45,13 @@ export default DS.Model.extend({
     }
     
   }),
+  isFile: computed('link', 'citation', 'absoluteFileUri', function() {
```
jrjohnson: this can just depend on type, type will depend on the others.

stopfstedt: tried that, didn't work. will try again.

stopfstedt:
@jrjohnson defaulting the `link` attribute to `http://` works fine, however that data get's passed to the server on PUT/POST, polluting the `learning_materials` table. 

We can handle this either on the client or server-side, I'm leaning towards the latter. Thoughts?

jrjohnson:
Ah right - so if its a citation you still end up with a link... yea.  How about we detect non fully qualified URLs on setLink on the entity and then add an http:// prefix there?  You could then return to checking for a URL as well.

saschaben:
a little uncomfortable with adding the prefix value after the user submits --- what about `https://`or `ftp://` or `ftps://`or other ....? I'd rather let them just throw garbage at the server in some respects.

jrjohnson:
I was thinking we would just detect any :// and not prefix those.

stopfstedt:
found a better solution for populating newly instantiated weblink-LMs that leaves the link attribute blank for non-weblinks.

jrjohnson:
@saschaben please review

saschaben:
This is OK, I guess, as far as it goes. we will need to add validation (in another issue ticket) to disallow passing either a blank value or a value that = `http://` with no further trailing string as an address.

saschaben:
:+1:

jrjohnson:
@homu r+

PR #1095, opened 2015-11-10
re-positioned close button on search form.

### LINKED ISSUE #330: Add (New) Session (or edit) >> MeSH vs. Directors >> Black "x" appears in Different Location
see the screen shots below ...

blackxdirectors
blackxmesh

stopfstedt:
fixes #330

### DIFF
--- a/app/styles/components/_searchbox.scss
+++ b/app/styles/components/_searchbox.scss
@@ -32,7 +32,7 @@
   }
 
   .close {
-    @include position(absolute, 0 0 0 'auto');
+    @include position(relative, 0);
     background-color: transparent;
     border-bottom-left-radius: 0;
     border-top-left-radius: 0;
--- a/app/styles/components/_searchbox.scss
+++ b/app/styles/components/_searchbox.scss
@@ -30,20 +30,4 @@
     padding-left: 6px;
     width: 20px;
   }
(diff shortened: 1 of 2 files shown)


jrjohnson:
This ends up with a double close button on chrome.  I think we should just remove the .close button and let people use the backspace key.  Then as browsers increase support for the 'search' type input the display will look good for them.  Alternatively remove the type='search' from the user search box.  So chrome doesn't get punished for doing thing right.

saschaben:
I think I agree with @jrjohnson --- let's remove the .close button and see how the users respond to it. If we need to we can flip this pretty easily, and re-add the .close button and remove the type="search" instead, until the other browsers catch up.

stopfstedt:
on it. thanks!

jrjohnson:
@dartajax you can test this at http://5644db9f71e20a2b60000025.ilios-frontend.netlify.com/  ensure that the director search and mesh search still work for a course.

dartajax:
I will test that right now.  Thanks for nelifying it.

dartajax:
:+1:

dartajax:
Although I found something else to report in a new issue.

jrjohnson:
@homu r+

PR #1097, opened 2015-11-11
fixed issue #1082

### LINKED ISSUE #1082: Learner Groups Not Available For Selection After Adding Instructor
How is this possible?  I know I did the following things ..

* selected an instructor for an existing session offering / event 
* attempted to at the same time to remove an instructor group -- that was impossible since I was doing the add of the instructor at the same time 
* clicked Done 
* noticed that the Learner Group got removed from the offering
* tried to fix it and noticed that the learner groups were not loading up for me to select one to fix this situation
* noticed the session below still had learner groups available for selection 

image

Reloading the page did not help the situation. 

This may not be the best description of this scenario but I have some more for tomorrow on this topic as well.

djvoa12:
Fixes #1082 
- Fixes `Available Learner Groups` not populating for offerings with no groups
- Fixes duplicate issue on the `Selected Learner Groups` side

### DIFF
--- a/app/components/offering-manager.js
+++ b/app/components/offering-manager.js
@@ -47,7 +47,9 @@ export default Ember.Component.extend({
 
   learnerGroups: null,
 
-  filteredCohorts: computed('cohorts.[]', 'filter', function(){
+  changeFlag: false,
+
+  filteredCohorts: computed('cohorts.[]', 'filter', 'changeFlag', function(){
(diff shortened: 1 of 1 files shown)


saschaben:
woot!

:+1:

jrjohnson:
:+1: via @saschaben 
@homu r+

PR #1098, opened 2015-11-11
save offering/user(s) association on the offering side.

### LINKED ISSUE #1080: Add New Offering With Instructor Fails
It could be re-writing to the User end point incorrectly or when it shouldn't be.

Steps below ...

image

image

image

image

image

stopfstedt:
fixes #1080.

### DIFF
--- a/app/components/session-offerings.js
+++ b/app/components/session-offerings.js
@@ -34,27 +34,24 @@ export default Component.extend({
     addSingleOffering({ startDate, endDate, room, learnerGroups, instructors, instructorGroups }) {
       const store = this.get('store');
       const session = this.get('session');
-      let offering = store.createRecord('offering', { session, startDate, endDate, room });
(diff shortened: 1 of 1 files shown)


stopfstedt:
@jrjohnson ok, this fixes the imminent issue of saving offering relationships from the wrong end.
however, this introduces a regression with the assignment of default instructors, default instructor-groups to offerings via small groups generator. they do not stick anymore. i've been unable to link these to the offering by chaining promises _before_ the offerings are created. still, better than before. any pointers are welcome. thanks!

jrjohnson:
:+1:  via Sascha
@homu r+

stopfstedt:
rebased against master.

jrjohnson:
:+1: group choice to move ahead with this as it solved many issues.  @stopfstedt should document remaining problems in a new ticket.

jrjohnson:
@homu retry

jrjohnson:
@homu r+

PR #1102, opened 2015-11-11
raised course query filter limit

### LINKED ISSUE #1064: "filter by course" list incorrect
the calendar filter for course is not accurately displaying the available courses for a given year:

CASE: 
1. go to courses & sessions. list courses for medicine 2015-16.
2. search for Foundations of Patient Care Year 1 (FPC1). Note publish status.
3. return to dashboard.
4. open calendar and turn on filters
5. flip to filter by course
6. select year 2015-16.
7. scroll thru the courses listed for filtering. you will not find FPC1

For any selected year, the filter course list should reflect:
* for a non-student: all courses for that year regardless of publish status
* for a student, all courses for that year with a status of published or scheduled.

Assigning this to @djvoa12 initially only because he has had his head under the hood of filters recently. If needed please reassign the @jrjohnson.

djvoa12:
Fixes #1064 

New limit is `1000`

### DIFF
--- a/app/components/dashboard-calendar.js
+++ b/app/components/dashboard-calendar.js
@@ -305,14 +305,15 @@ export default Ember.Component.extend({
   selectedCourses: [],
   allCourses: Ember.computed('selectedSchool', 'selectedAcademicYear', function(){
     let defer = Ember.RSVP.defer();
-    this.get('selectedSchool').then(school => {
-      this.get('selectedAcademicYear').then(year => {
+    this.get('selectedSchool').then((school) => {
+      this.get('selectedAcademicYear').then((year) => {
         this.get('store').query('course', {
           filters: {
             school: school.get('id'),
             year: year.get('title')
-          }
-        }).then(courses => {
+          },
+          limit: 1000
+        }).then((courses) => {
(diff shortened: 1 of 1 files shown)


jrjohnson:
@saschaben what do you think the absolute maximum number of courses that might be attached to a program year might be?  Is 200 sufficient?  Or should we make it 1000?

saschaben:
We should probably raise it, but not to 1000. 300 should be ample.

jrjohnson:
I should have mentioned that there is no cost to picking a number that is too high.  We're shooting for all of the results.  So 10000 is not more expensive than 10 unless you actually have 10000 in which case I guess you would still want them all - but probably be calling Dave to add paging to the list.

saschaben:
1000 is a big scary number. but I guess if this is a no-cost no-fault thing then that's fine. I can't actually conceive of a curriculum with more than a few hundred courses listed in a year, but hey.

saschaben:
:+1:

djvoa12:
New limit = 1000

saschaben:
:+1:

jrjohnson:
@homu r+

PR #1104, opened 2015-11-12
Update drop down component

### LINKED ISSUE #902: Drop down control not touch responsive
Our current dropdown  select control does not respond on touch screen devices. This is an issue regardless of full mobile implementation as we have many users on tablets as their primary device for calendar, and this issue precludes use of  logout, translation, and causes failures with tooltips..

jrjohnson:
Couple of fixes in this:
- It should now be possible to logout on touch devices
- Logout was borken on netlify (or anything not proxied) it is now fixed
- Netlify couldn't be refreshed, but it now can be with the addition of some redirect rules
- Disabled autocorect and autocomplete on the uesrname login field.  Making it way less annoying to login on an iPhone (and probably most any mobile device)

Review URL: http://5643f0b4d6865d2e31000004.ilios-frontend.netlify.com/

Fixes #902


saschaben:
:+1:

thecoolestguy:
@homu r+

PR #1110, opened 2015-11-13
only save the program-year when (un)assigning a director from/to it.

stopfstedt:
don't save the director/user.


jrjohnson:
:+1: @homu r+

PR #1111, opened 2015-11-13
added validation for learning material url

### LINKED ISSUE #1106: add validation to learning material weblink
the LM weblink entry field should have validation added to restrict:

Must have length > 0
must NOT === `http://`

djvoa12:
Fixes #1106

### DIFF
--- a/app/components/new-learningmaterial.js
+++ b/app/components/new-learningmaterial.js
@@ -7,25 +7,41 @@ const { computed } = Ember;
 const { alias } = computed;
 
 export default Ember.Component.extend(EmberValidations, {
+  init() {
+    const type = this.get('learningMaterial.type');
+
+    switch (type) {
+      case 'file':
+        this.set('isFile', true);
+        break;
+      case 'link':
+        this.setProperties({ isLink: true, 'validations.urlBuffer': { presence: true, url: true } });
+        break;
+      case 'citation':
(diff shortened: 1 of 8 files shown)


jrjohnson:
@djvoa12 failing tests

jrjohnson:
@saschaben is going to have to take a look at this one.  I'm not sure we want to go this far with validations.  This checks if the URL is a valid one, but doesn't allow things like `afp://campusserver/course.txt` which might be valid in some cases.  I don't think we want to go through all of the possible valid URLs we might need here.  Probably just ensure that they haven't left an empty http:// behind.

saschaben:
@jrjohnson is correct: we don't want to validate for well-formed / valid urls here, as it is too restrictive for the context. We just want to make sure we don't have blank entries, or entries with orphaned default values (eg "http://"). This is a GIGO situation, we can only guide users toward proper data entry here, not force them. The code at this point here is a little too restrictive.

djvoa12:
Updated url validation definition. Ready for review.

jrjohnson:
I reviewed at http://564a538b71e20a665b000002.ilios-frontend.netlify.com/courses/595?details=true

:+1:  looks good, waiting on tests.

jrjohnson:
@homu r+

PR #1113, opened 2015-11-13
notes and publicNotes display straightened out

### LINKED ISSUE #1083: LM notes flag incorrect
looks like we are incorrectly setting the LM notes flag. Right now, under the LM header column "notes", we show "yes" if there is a note AND it is set to visible for students. Otherwise we set it to "No".

What it should be (and this will require a little bit of tweaking), should be "Yes" if there is a note of any value --- regardless of who can see it --- and an eye icon or some other notification if it is hidden/visible. The validation should not be on the visibility, but on the existence of the note.

djvoa12:
Fixes #1083 

Notes: `undefined`, publicNotes: `true`

screen shot 2015-11-13 at 3 15 59 pm

Notes: `a`, publicNotes: `false`

screen shot 2015-11-13 at 3 17 08 pm

### DIFF
--- a/app/templates/components/detail-learning-materials.hbs
+++ b/app/templates/components/detail-learning-materials.hbs
@@ -95,11 +95,15 @@
         	        {{/if}}
                 </td>
                 <td class='text-center' colspan=2>
-        	        {{#if lmProxy.publicNotes}}
+                  {{#if lmProxy.notes}}
         	          <span class='add'>{{t 'general.yes'}}</span>
+
+                    {{#if lmProxy.publicNotes}}
+                      {{fa-icon 'eye'}}
+                    {{/if}}
         	        {{else}}
         	          <span class='remove'>{{t 'general.no'}}</span>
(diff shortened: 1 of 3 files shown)


jrjohnson:
This ends up displaying weird data on first creation because publisNotes defaults to true.  So we get No (eyeball) on first creation.  Instead only display the eyeball if notes has content.  So the options are
1.  No
2. Yes
3. Yes (eyeball)

djvoa12:
I am not really seeing this behavior you are talking about. It seems to work fine for me.

jrjohnson:
Sorry - could have been clearer.  The behavior I want to remove is the exact one from your example screen shot at the top of this PR.  The eye shouldn't appear next to a 'No'.  If there is No notes then it doesn't matter if they are public or not.

djvoa12:
Fixed. Ready for review.

jrjohnson:
:+1: looks good.
@homu r+

PR #1114, opened 2015-11-14
updated true overall learner-group count

### LINKED ISSUE #1021: Count on Detail Screen Should Reflect Sum of All Sub-Groups in Learner Groups
See screenshots for clarification.  There are 328 members in the Anatomy Lab group.  They are all in sub-groups as shown in the second screen shot but the count at the top of the detail screen indicates there are zero members of the Anatomy Lab group (and all of its associated sub-groups)

anatomylabcount

The ugly arrow is pointing to the location where the count should reflect the 328 members contained in Anatomy Lab (the total of all of the members in all of the 26 sub-groups).

anatomylabcount2

### LINKED ISSUE #1074: Learner Group Total Count is Incorrect
This is roughly related to #1021 in the sense that it is regarding Learner Groups and the counts of members therein.  However, this is only about the higher level count on the Learner Groups main page.

The BMB Entire Class group does indeed have 163 members and 0 sub-groups.  All members are at the higher level.

Anatomy Labs (and all of the associated sub-groups) does NOT contain 328 members ...

image

Here are the details of the Anatomy Labs sub-groups ...

image

After adding up all of the members of all of those sub-groups, the total is ...

<H1>164</H1>

This lower screen count seems to reflect this as well.

image

djvoa12:
Fixes #1021 
Fixes #1074

### DIFF
--- a/app/components/learnergroup-header.js
+++ b/app/components/learnergroup-header.js
@@ -1,11 +1,16 @@
 import Ember from 'ember';
 
-export default Ember.Component.extend({
+const { Component } = Ember;
+
+export default Component.extend({
   learnerGroup: null,
+
   actions: {
-    changeTitle: function(newTitle){
-      this.get('learnerGroup').set('title', newTitle);
-      this.get('learnerGroup').save();
-    },
+    changeTitle(newTitle) {
+      const learnerGroup = this.get('learnerGroup');
+
+      learnerGroup.set('title', newTitle);
+      learnerGroup.save();
+    }
   }
 });
--- a/app/templates/components/learnergroup-header.hbs
+++ b/app/templates/components/learnergroup-header.hbs
@@ -1,13 +1,12 @@
 <div class='detail-header'>
   <span class='title'>
     <h2>
-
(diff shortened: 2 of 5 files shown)


jrjohnson:
All the tests taht are failing seem to be legitimate.  They were testing to old way and now they need to be updated to cover the new way of counting.  I put a patch in slack for these changes.

jrjohnson:
I'm seeing some issues with these tests when attempting to clean up deprecations.  I think we might be relying on a buggy behavior for the way records reload in the background.  I'm digging in to see what is up there and it may apply here or we may just want to remove these tests until we have time to sort them out correctly.

jrjohnson:
Yup.  I vote to go ahead an comment these tests out with a todo note.  I will merge this and then we can find a fix later

djvoa12:
I commented out the four assertions and created this [issue](https://github.com/ilios/frontend/issues/1157).

This one is good to go.

jrjohnson:
:+1 reviewed and counted.  The numbers are now correct.

jrjohnson:
:+1:  reviewed and counted.  The numbers are now correct.

jrjohnson:
@homu r+

PR #1119, opened 2015-11-15
all calendar controls have query params

### LINKED ISSUE #1070: School choice on dashboard calendar should be sticky
When you choose a school other than your primary school and then go to an event and click back you will be stuck in your primary school again.  This is dumb.  The school choice needs to be a query param so it is sticky.

djvoa12:
Calendar Controls Query Params:
- [x] `Calendar On` & `Calendar Off`
- [x] `All Events` & `My Schedule`
- [x] `Show Filters` & `Hide Filters`
- [x] `Topic/Detail` & `Course/Type`
- [x] Academic Years `2016-2017, etc.`
- [x] School `Medicine, etc.` 

Acceptance Test

Fixes #1070

### DIFF
--- a/app/routes/dashboard.js
+++ b/app/routes/dashboard.js
@@ -15,12 +15,12 @@ export default Route.extend(AuthenticatedRouteMixin, {
 	},
 
 	model() {
-		const user = this.get('currentUser.user');
+		const user = this.get('currentUser.model');
 		const schools = user.get('schools');
 		const store = this.store;
 		const academicYears = store.findAll('academic-year');
 
-		return hash({ user, schools, academicYears });
+		return hash({ schools, academicYears });
 	},
 
 	setupController() {
--- a/app/routes/dashboard.js
+++ b/app/routes/dashboard.js
@@ -1,11 +1,31 @@
 import Ember from "ember";
 import AuthenticatedRouteMixin from 'simple-auth/mixins/authenticated-route-mixin';
 
-export default Ember.Route.extend(AuthenticatedRouteMixin, {
-	i18n: Ember.inject.service(),
+const { inject, Route, RSVP } = Ember;
+const { service } = inject;
+const { hash } = RSVP;
+
+export default Route.extend(AuthenticatedRouteMixin, {
+	i18n: service(),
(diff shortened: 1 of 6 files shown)


jrjohnson:
@djvoa12 This is GREAT!  You have a couple of JsHint errors in dashboard that need attention.

app/services/current-user.js
```diff
@@ -25,7 +27,8 @@ export default Ember.Service.extend({
         deferred.resolve(null);
       });
     } else {
-      this.get('store').find('user', currentUserId).then(function(user){
+      this.get('store').find('user', currentUserId).then((user) => {
+        this.set('user', user);
```
jrjohnson: Why set this?  Is it being used somewhere?

app/routes/dashboard.js
```diff
@@ -1,11 +1,31 @@
...
 	today: new Date(),
-	setupController: function(){
-    this._super.apply(arguments);
+
+	beforeModel() {
+		return this.get('currentUser.model');
+	},
+
+	model() {
+		const user = this.get('currentUser.user');
```
djvoa12: `user` property is being used right here.

app/services/current-user.js
djvoa12: `current-user` probably needs some refactoring. It should be called `session` or `authManager`. We need to set a `user` or `currentUser` property with user object that the server returns and store it there after a user logs in so we can retrieve any user related info on the spot without making any requests.

app/routes/dashboard.js
jrjohnson: OK - For now this should just be currentUser.model and then take out the set('user' stuff.  I will refactor this whole thing once we can update to ember-simple-auth 1.0 which is waiting on a few dependancies.  It will be much easier to manage the refactor if I can search for currentUser.model everywhere are replace it.

app/services/current-user.js
jrjohnson: Mostly agreed, but we're stuck with this for a few more weeks.

jrjohnson:
:+1: looks great.  @homu r+

PR #1120, opened 2015-11-15
fixed issue #1117

### LINKED ISSUE #1117: Learning Material Description in search results need to show HTML
Currently we're escaping this HTML when searching for a learning material:

lmhtml

djvoa12:
@jrjohnson Can we remove the rule in SCSS linter, where one cannot use `!important`? I need to override the inline styles and I wouldn't have to spend time finding ways around previously defined CSS.

Without override:
screen shot 2015-11-14 at 6 15 29 pm

With override:
screen shot 2015-11-14 at 6 23 21 pm

Fixes #1117

### DIFF
--- a/app/templates/components/learningmaterial-search.hbs
+++ b/app/templates/components/learningmaterial-search.hbs
@@ -20,7 +20,7 @@
     {{#each results as |learningMaterial|}}
       <li {{action 'add' learningMaterial}}>
         <h4 class="learning-material-title">{{learningMaterial.title}}</h4>
-        <span class="learning-material-description">{{learningMaterial.description}}</span>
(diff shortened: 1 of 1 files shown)


jrjohnson:
Closed while trying to comment I guess - sorry.
> I'm sort of OK with this use of !important, but since this rule is pretty specific I'm wondering what we have that is breaking it and maybe we should fix that.

djvoa12:
If you could somehow pre-remove the inline-style that comes with the html, that would work.

screen shot 2015-11-15 at 1 43 05 pm

jrjohnson:
:+1: @homu r+

PR #1129, opened 2015-11-17
re-wrote new learning-material addition

### LINKED ISSUE #1124: When a new LM is being created switch the Add New button to a minus sign
This will look a lot like the pattern for adding courses and sessions.  Choosing a New LM type from the menu opens the create new form, but after that the Add button should turn into a minus sign and clicking it would cancel adding the new LM.  This prevents multiple materials from being added.

djvoa12:
- [x] re-wrote/refactored new add process for `learning-material`
- [x] added collapse button
- [x] modified/added tests

Fixes #1124

### DIFF
--- a/app/components/detail-learning-materials.js
+++ b/app/components/detail-learning-materials.js
@@ -2,7 +2,7 @@ import Ember from 'ember';
 import DS from 'ember-data';
 import { translationMacro as t } from "ember-i18n";
 
-const {computed, inject, isPresent, RSVP} = Ember;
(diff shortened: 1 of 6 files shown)


jrjohnson:
There is no need for creating multiple learning materials.  Its a feature we had early and I think its pretty clearly not very useful compared to a clean simple approach.

djvoa12:
@jrjohnson Ready for review.

jrjohnson:
:+1:

PR #1131, opened 2015-11-17
Update Ember CLI and Other Dependencies

jrjohnson:
New version of ember-cli as well as related updates to ember and the rest of our library ecosystem.  There are some fixes to tests and environment that were required for this updated.


thecoolestguy:
@homu r+

Issue #1000, opened 2015-11-03
Single User Page Needs Details >> Manage Users


dartajax:
This is in the "Find Users" area of the Admin screen.

There is currently a placeholder for the maintenance of the account.  We need to provide the fields to review and maintain the user's account.

`Placeholder for user: #22 David Leonard`

... is the way it currently displays.

PR #1132, opened 2015-11-18
added pagination and search to users route, modified users hbs, started on user hbs

### LINKED ISSUE #998: Need to Add pagination for Manage Users or Endless Scroll
Admin >> Manage Users button.

The resulting list cuts off at approximately twenty records and does not provide a mechanism for loading up the next batch of records.  A button to perform that action needs to be added or we can provide an endless scroll to see all of the records.

### LINKED ISSUE #999: Need to Make Users Clickable on Manage Users
Currently a list displays but there is no way to select a User whose account needs to reviewed or modified.  We need to make the User's record clickable to maintain the account.

The display currently shows the following columns in this order ....

First Name
Last Name 
School

Shouldn't we either ....

* Put the Last Name field first. or 
* Concatenate (Last Name, First Name) in one Name field

djvoa12:
- [x] user-pagination (20 at a time) - query params
- [x] search (max-result of 20) - pagination
- [x] linking `user` route from `users` route
- [x] layout for `users` route, template, controller
- [x] layout for `user` route, template, controller
- [x] translations
- [x] acceptance tests (`users`, `user`, `admin`)

Fixes #998 
Fixes #999 
Fxies #1000


app/controllers/users.js
```diff
@@ -1,47 +1,105 @@
...
+function cleanQuery(query) {
+  return trim(query).replace(/[\-,?~!@#$%&*+\-'="]/, '');
+}
+
+export default Controller.extend({
+  init() {
+    this._super(...arguments);
+
+    // This needs to be removed after user session re-write
+    this.get('currentUser.model').then((user) => {
```
jrjohnson: Promises don't belong in the init method.  The should be handled exactly like any of the other lists with necessary data being fetched in the route.

jrjohnson:
Sorting and page (offset) can be used through the API like this:
http://ilios.dev/api/v1/users?order_by[lastName]=ASC&order_by[firstName]=ASC&limit=10&offset=1

Then offset =11 would give the next ten.

app/controllers/users.js
djvoa12: It's a substitute until the `session` service gets implemented. I shouldn't have to send any requests when that happens and this should go away.

djvoa12:
API gives `firstName` sorting higher priority than `lastName` due to alphabetizing. How can I reverse it?

jrjohnson:
That's not what I'm seeing in some limited testing.  Can you give me the URL ember is generating and I will investigate against the API?

djvoa12:
@jrjohnson Ready for final review.

jrjohnson:
@djvoa12 needs an scss-lint fix

jrjohnson:
@saschaben the last failing test change shouldn't effect testing.  Can you test functionality?

jrjohnson:
I'm seeing some weird formatting when there are learner groups:
screen shot 2015-11-24 at 6 55 38 pm

app/controllers/users.js
jrjohnson: Its not a good temporary solution either.  Having a promise in init follows under the same logic as observers in general.  It makes side effects very difficult to understand.  Any promises needed before the controller is build needs to go into the route.

app/controllers/users.js
```diff
@@ -1,47 +1,139 @@
...
+  }).readOnly(),
+
+  morePages: computed('model', {
+    get() {
+      const { model, limit } = this.getProperties('model', 'limit');
+      return model.length > limit;
+    }
+  }).readOnly(),
+
+  users: computed('model', {
```
jrjohnson: Instead of using the model here and having the complication of a workaround in the route why not just run the query here?  This data doesn't need to be resolved before the controller loads so it makes good sense to just do the work here in the controller instead of splitting half of it into the route.

app/controllers/users.js
```diff
@@ -1,47 +1,139 @@
...
+        return users.slice(0, 20);
+      } else {
+        return users;
+      }
+    }
+  }).readOnly(),
+
+  resultsCount: null,
+
+  searchResults: computed('query', 'offset', {
```
jrjohnson: When you move the basic query out of the model you can combine the users and the searchResults methods and remove the branching logic out of the template.  The combine method can detect if a query is present and calculate the results from that.  If not then it can use the pagination.  This might even allow you to move the query up into the URL as a query param so it is more stable and persistent.

app/models/user.js
```diff
@@ -1,8 +1,9 @@
 import Ember from 'ember';
 import DS from 'ember-data';
 
-const { computed, RSVP } = Ember;
+const { computed, PromiseProxyMixin, RSVP } = Ember;
 const { PromiseArray } = DS;
+const ProxyContent = Ember.Object.extend(PromiseProxyMixin);
```
jrjohnson: Please use DS.PromiseObject for these.  Its the exact same thing, but PromiseObject is what is used everywhere else so it will be easier to maintain over time.

app/models/user.js
```diff
@@ -91,6 +105,16 @@ var User = DS.Model.extend({
       }
       return first + ' ' + last;
   }),
+
+  displayName: computed('firstName', 'middleName', 'lastName', {
```
jrjohnson: displayName can replace fullName.  I don't think we need two ways of printing the users name.

app/styles/components/_admin-dashboard.scss
```diff
@@ -0,0 +1,8 @@
+.admin-container {
```
jrjohnson: The filename should be admin-container if that is the class name.

app/styles/components/_users-pagination-links.scss
```diff
@@ -0,0 +1,14 @@
+.users-pagination {
```
jrjohnson: file name should be user-pagination of else the class name should be user-pagination-links

app/styles/user.scss
```diff
@@ -0,0 +1,41 @@
+.user-container {
```
jrjohnson: same deal.  file name should be user-container.scss

jrjohnson: Also file should be in components.  All sass goes in there.

app/styles/users.scss
```diff
@@ -0,0 +1,18 @@
+.resultslist-header {
```
jrjohnson: this is going to override resultslist-header for every list on the site.  It needs to be underneath another class or the name needs to change.

File is called users.scss so I would ecpect to find a .users {

File should be in components.  Not in styles.

app/templates/components/global-search.hbs
```diff
@@ -2,11 +2,11 @@
   type="search"
   placeholder={{t 'general.globalSearchPlaceholder'}}
   incremental=true
-  onsearch={{action "inputValueChanged"}}
-  onkeyup={{action "inputValueChanged"}}
+  oninput={{action "inputValueChanged"}}
```
jrjohnson: onsearch event is much better and onkeyup provides cross browser compatibility.  Please revert.

app/templates/components/user-list.hbs
```diff
@@ -1,17 +1,25 @@
 <table>
   <thead>
     <tr>
-      <th class='text-left' colspan=1>{{t 'user.firstName'}}</th>
-      <th class='text-left' colspan=1>{{t 'user.lastName'}}</th>
-      <th class='text-left' colspan=1>{{t 'general.school'}}</th>
+      <th class='text-left' colspan=1></th>
+      <th class='text-left' colspan=3>{{t 'user.fullName'}}</th>
+      <th class='text-left' colspan=2>Campus ID</th>
```
jrjohnson: Needs translation.  Or probalby just to use an existing translation.

app/styles/components/_user-list.scss
```diff
@@ -0,0 +1,9 @@
+.user-list {
+  .user-status {
+    color: $ilios-red;
+  }
+
+  .italicize {
```
jrjohnson: Since this is subclassed here it doesn't need to be a generic name.  You should call this class disabled-user-account so it is easy to tell what it does.

app/templates/components/user-list.hbs
```diff
@@ -1,17 +1,25 @@
...
   <tbody>
     {{#each users as |user|}}
-      <tr class="{{if course.showRemoveConfirmation 'confirm-removal'}}">
-        <td class='text-left' colspan=1>{{user.firstName}}</td>
-        <td class='text-left' colspan=1>{{user.lastName}}</td>
-        <td class='text-left' colspan=1>{{user.school.title}}</td>
+      <tr class="{{unless user.enabled 'italicize'}}">
+        <td class='text-left user-status' colspan=1>
+          {{#unless user.enabled}}
+            {{fa-icon 'user-times' title=(unless user.enabled 'disabled')}}
```
jrjohnson: "disabled" needs to be translated.  Also probably doesn't need to be in an unless block since it is inside one that has the same function.

title=(t 'general.disabled')

app/templates/components/users-pagination-links.hbs
```diff
@@ -0,0 +1,31 @@
+<div class="users-pagination">
+  {{#if prevPages}}
+    {{#if arrows}}
+      <a class="prev-arrow" {{action 'getPrevPage'}}>{{fa-icon 'arrow-left'}}</a>
+    {{else}}
+      <a class="prev-button" {{action 'getPrevPage'}}>Prev</a>
```
jrjohnson: "Prev" needs to be translated.

app/templates/components/users-pagination-links.hbs
```diff
@@ -0,0 +1,31 @@
...
+    {{/if}}
+  {{/if}}
+
+  <span>|</span>
+
+  {{#if morePages}}
+    {{#if arrows}}
+      <a class="next-arrow" {{action 'getNextPage'}}>{{fa-icon 'arrow-right'}}</a>
+    {{else}}
+      <a class="next-button" {{action 'getNextPage'}}>Next</a>
```
jrjohnson: Next needs to be translated.

app/templates/components/users-pagination-links.hbs
```diff
@@ -0,0 +1,31 @@
...
+    {{#if arrows}}
+      <a class="prev-arrow" {{action 'getPrevPage'}}>{{fa-icon 'arrow-left'}}</a>
+    {{else}}
+      <a class="prev-button" {{action 'getPrevPage'}}>Prev</a>
+    {{/if}}
+  {{else}}
+    {{#if arrows}}
+      <span>{{fa-icon 'arrow-left'}}</span>
+    {{else}}
+      <span>Prev</span>
```
jrjohnson: again

app/templates/components/users-pagination-links.hbs
```diff
@@ -0,0 +1,31 @@
...
+    {{#if arrows}}
+      <a class="next-arrow" {{action 'getNextPage'}}>{{fa-icon 'arrow-right'}}</a>
+    {{else}}
+      <a class="next-button" {{action 'getNextPage'}}>Next</a>
+    {{/if}}
+  {{else}}
+    {{#if arrows}}
+      <span>{{fa-icon 'arrow-right'}}</span>
+    {{else}}
+      <span>Next</span>
```
jrjohnson: again.

app/templates/user.hbs
```diff
@@ -1,2 +1,71 @@
...
+  {{global-search}}
+
+  <h1 class="user-display-name">{{model.displayName}}</h1>
+  {{#if model.isStudent.content}}
+    <span class="user-is-student">{{t 'general.student'}}</span>
+  {{/if}}
+
+  <div class="user-info">
+    <div class="form-col-12">
+      <label class="form-label">Campus ID:</label>
```
jrjohnson: Campus id should be translated

app/templates/users.hbs
```diff
@@ -4,27 +4,57 @@
...
+            <span class="no-results">{{general.noResultsFound}}</span>
+          {{/if}}
+        {{/if}}
+      {{else}}
+        {{user-list users=users}}
+      {{/if}}
+    </div>
+
+    <div class="pagination-bottom">
+      {{#if query}}
```
jrjohnson: I don't see the difference between the if/else block here.  Am I just tired?

app/templates/components/user-list.hbs
djvoa12: @saschaben said `Campus ID` can stay without one.

jrjohnson: He's wrong event if he is right.  Al strings need to have a translation.  Event it ends up being the same phrase in all languages.

app/templates/users.hbs
djvoa12: Yup. You could use a nap haha. The `users-pagination-links` in the bottom are referencing different computed properties.

`prevResults vs. prevPages` & `moreResults vs. morePages` (due to both route and search being paginated)

jrjohnson: Ah, there it is.  man I really looked at this a few times.

app/templates/components/users-pagination-links.hbs
saschaben: "previous" FR: "précédent" (abbreviated "précéd.")

app/templates/components/user-list.hbs
saschaben: so add "Campus ID" as the trans phrase for Campus IDin all lang files.

app/templates/components/user-list.hbs
saschaben: disabled FR: "désactivée"

app/templates/components/users-pagination-links.hbs
saschaben: next FR: "prochain"

saschaben:
I'm seeing the same weird layout issue as @jrjohnson regarding student groups -- BUT:

1) it is only with students (eg demo student4), not student instructors (demo student instructor1);
2) if you type "demo " into the search bar when the search activates its first call the screen resolves to display the current layout correctly.

saschaben:
Please add a `margin-bottom: 1.5em;` to the class `user-container .user-display-name` to avoid having the name and the search entry form overlap.

djvoa12:
@jrjohnson @saschaben I made a lot of style improvements on the next PR.

app/templates/components/user-list.hbs
djvoa12: Done.

app/templates/components/users-pagination-links.hbs
djvoa12: Done.

app/templates/components/users-pagination-links.hbs
djvoa12: Done.

app/templates/components/users-pagination-links.hbs
djvoa12: Done.

app/templates/components/users-pagination-links.hbs
djvoa12: Done.

app/templates/user.hbs
djvoa12: Done.

app/templates/components/user-list.hbs
djvoa12: Done.

app/templates/components/global-search.hbs
djvoa12: Done.

app/controllers/users.js
djvoa12: I got a work-around.

jrjohnson: Why workaround. Just stick it in the route.

djvoa12: I did shove it in the model. The `init` is no longer there.

app/controllers/users.js
djvoa12: Well, promises that contain data to populate the page upon entering the route should be in the route. Ember also gives me query param support in the route. I only needed to use this here to see if next pages are to be shown. Ideally, if API has `nextPage` property, I would remove this CP. This is a better lay-out than having Controller doing the work.

jrjohnson: Ok we can merge it this way.  I may take a refactor pass at it later.  But since it is working lets move on.

app/controllers/users.js
djvoa12: Since searching for users is a separate task and not route-driven, I kept them separate. I think I can put `query` and `searchPage` as query params. I am not sure if they super necessary, but it's doable.

Query Params:
- page (route-driven)
- query (controller)
- searchPage (controller)

jrjohnson: Nah. I'm thinking refactor.

app/models/user.js
djvoa12: Done.

app/models/user.js
djvoa12: Done.

app/styles/components/_admin-dashboard.scss
djvoa12: Done.

app/styles/components/_users-pagination-links.scss
djvoa12: I took out `<div class="users-pagination">` from `user-pagination-links.hbs` and changed above to `.users-pagination-links`.

I guess I was slightly coding for the future. Components will eventually render only the contents inside it without the wrappers. When that happens, some type of wrapping div needs to inserted (same for lots of other components).

app/styles/user.scss
djvoa12: Done.

app/styles/users.scss
djvoa12: I moved the file to components folder and nested the code under `#users` section.

app/styles/components/_user-list.scss
djvoa12: Done.

app/styles/components/_users-pagination-links.scss
jrjohnson: Yea. That's one if the reasons I wanted this fixed. That way it will be easer to straighten out when we go to angle bracket components.

djvoa12:
@jrjohnson other than the style changes to come in the next PR, I think I addressed all comments and concerns.

jrjohnson:
Looks good.  ping me again when tests are passing and i will test and merge.

saschaben:
I think we may be incorrectly calling secondary cohorts here:

cohorts

This may also be the cause of the student group layout weirdness... @djvoa12 if you've already addressed this in the next PR, please disregard. But if not....let's investigate.

jrjohnson:
@djvoa12 needs a rebase.

jrjohnson:
:+1:

PR #1133, opened 2015-11-18
removed unnatural scrolling related to objectives

### LINKED ISSUE #729: Select Parent Objective for Session objective scrolls to wrong target
If the course details are open, selecting a session parent objective will scroll you to the top of the course objectives, rather than to the opened select parent objective panel for the session,

djvoa12:
For courses and sessions:
- [x] removed scroll for `Add New` button
- [x] removed scroll that occurs when `undo/cancel` button is clicked

Fixes #729

### DIFF
--- a/app/components/detail-objectives.js
+++ b/app/components/detail-objectives.js
@@ -41,7 +41,6 @@ export default Ember.Component.extend({
   actions: {
     manageParents: function(objective){
       objective.get('parents').then((parents) => {
-        scrollTo(".detail-objectives");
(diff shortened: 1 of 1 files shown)


jrjohnson:
I think this is functionality that we need.  We have to orient the users back into the area of the screen that they left when opening the objective manager.

jrjohnson:
oh 729 - I think that is about the target of the scroll not the scroll itself.

djvoa12:
That's correct.

jrjohnson:
:+1: @homu r+

PR #1134, opened 2015-11-18
UI guides user to correct dates

### LINKED ISSUE #1002: Need to Validate that Course End Date > Course Start Date
Currently, there is no problem saving a Course with a Start Date that occurs after the End Date.  We should fix this.

See screenshot for reference.

chumbawumba

djvoa12:
This is not like the validation currently on other input areas. It merely guides the user by putting the dat-picker (for both start and end) in edit mode if the end-date is on or before the start-date. If real validation must be used, then I will likely need to use observers within `inplace-date` component.

Fixes #1002

### DIFF
--- a/app/components/course-overview.js
+++ b/app/components/course-overview.js
@@ -1,8 +1,13 @@
 import Ember from 'ember';
 import DS from 'ember-data';
 
-export default Ember.Component.extend({
-  store: Ember.inject.service(),
+const { Component, inject } = Ember;
+const { service } = inject;
+
+export default Component.extend({
+  store: service(),
+  flashMessages: service(),
+
   editable: true,
   course: null,
   directorsSort: ['lastName', 'firstName'],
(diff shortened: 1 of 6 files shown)


jrjohnson:
Looks great.  Add a flash message to say 'Course start date should be before course end date.' and then everything else can stay the same.  There will just be a visual cue about WHY the other box is opening.

saschaben:
I suspect this pattern is going to confuse users --- we don't actually restrict bad dates, just provide the alert; we also don't give the chance to revert to a previous value (eg, I change start date to a post-end date value, see the flash message, get the edit mode on end date --- but no aid in reverting start date to its original value). This can go forward, but in general we should be actually validating dates so that there is no place where a start date/time can come after an end date/time value across the application.

In the past our date restriction/validation pattern has been to handle this, and to ignore contextually "incorrect" dates -- like having a session which begins/ends outside the range of its course's start and end dates -- since there are frequent situations where that might be desired (don't ask). If we can get to a point where we are doing the same in the new UI, that would be a good thing.

jrjohnson:
After discussion with @saschaben we decided to not validate this field until we can do it in a more conventional way.

PR #1137, opened 2015-11-19
re-styled learning-material search-bar

### LINKED ISSUE #1116: Learning material search on courses and session in miss-aligned
Spacing on the element appears to be breaking floats to the search box shows up a line below the title and add button:

lmalignment

djvoa12:
Used absolute positioning for `detail-title` and `detail-actions` classes

Fixes #1116

### DIFF
--- a/app/styles/components/_detail-learning-materials.scss
+++ b/app/styles/components/_detail-learning-materials.scss
@@ -1,11 +1,23 @@
-.detail-learning-materials {
-  .learning-material-actions {
-    @include span-columns(3 of 12);
-    @include omega;
-    float: right;
+.detail-container {
+  position: relative;
+
+  &.padding-top {
+    padding-top: 4em;
+  }
+
+  .detail-title-absolute {
+    color: $ilios-orange;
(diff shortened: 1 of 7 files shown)


app/styles/components/_detail-learning-materials.scss
```diff
@@ -1,11 +1,23 @@
-.detail-learning-materials {
-  .learning-material-actions {
-    @include span-columns(3 of 12);
-    @include omega;
-    float: right;
+.detail-block {
```
jrjohnson: Why did you un-scope this?  That will cause this style to creep out into the rest of the app.  If that should happen then this belongs in detailView.scss

djvoa12: I changed it to `detail-container`. Since couple of `detail-container`'s child elements are absolutely positioned, `detail-container` needs to have a `relative` position as a parent. It can't apply to the component class.

jrjohnson:
:+1:

jrjohnson:
@homu r+

PR #1154, opened 2015-11-25
Refactor course cohort details

### LINKED ISSUE #560: Program Cohort Picker needs formatting work
here is what I am seeing - 

cohortstuff

### LINKED ISSUE #630: Fix cohort manger to use a buffers
The cohort manager currently interacts directly with the relationship on the course it is assigned to. This is a bug because changes to the cohorts get shared around the application (like to parent objective selector) before they are saved.
Instead we need to use a buffer like the program year competencies manager.

### LINKED ISSUE #847: Remove displayTitleObserver from cohort model
We're currently using an old observer pattern here and it should be a PromiseObject returned from a computed displayTitle property.

### LINKED ISSUE #1144: Unable to add a cohort to a course
Saving course cohorts results in a permissions error for demo_faculty1.

### LINKED ISSUE #1146: Add Cohort Process Needs Refinement
There are a number of ways this could be improved.  This is related to #1039.  Please refer to my comments on that ticket for reference.

When I go to a Course -- see screenshots ..

image

When you look at Jason's Test Course, how would you know there is no Cohort selected yet?

image

The user should not even have the ability to add Sessions if there is not a Cohort attached to the Course. I guess that's a different issue.

Anyway, to add a cohort ...

Click Show Details

image

This is what you get ... See my comments on the image itself.

addcohort

When the user scrolls, they get this ...

image

The years are in order (ascending) which is good but it's confusing when there are multiple schools and programs and then only the year appear.

jrjohnson:
- Pulled list and manager out into separate components
- Using a buffer to track the state so it doesn’t leak out into the app
- Improved listing and sorting of available courts
- better indication of loading state with a spinner

Fixes #1144
Fixes #1146
Fixes #847
Fixes #630
Fixes #560

### DIFF
--- a/app/components/detail-cohort-list.js
+++ b/app/components/detail-cohort-list.js
@@ -1,0 +1,20 @@
+import Ember from 'ember';
+import DS from 'ember-data';
+
+const {computed, RSVP} = Ember;
+const {PromiseArray} = DS;
+
+export default Ember.Component.extend({
+  cohorts: [],
+  sortedCohorts: computed('cohorts.@each.{school,displayTitle}', function(){
+    let defer = RSVP.defer();
+
+    this.get('cohorts').then(cohorts => {
+      defer.resolve(cohorts.sortBy(['school', 'displayTitle']));
+    });
+    
+    return PromiseArray.create({
+      promise: defer.promise
+    });
+  }),
+});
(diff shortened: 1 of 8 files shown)


jrjohnson:
@dartajax as most of the tickets were yours please test at http://565504b971e20a3f980000c6.ilios-frontend.netlify.com/courses/551

dartajax:
:+1: - it looks good to me

PR #1155, opened 2015-11-25
inactive lists are not clickable

### LINKED ISSUE #1136: Director Search Results Display Disabled Looking Records that can Be Selected
Notice below how Julie Diaz' account appears disabled but it can still be selected and added as a Director.  This may not actually be an issue if it has something to do with demo data and whatnot but here goes for now ...

image

Select the disabled looking Julie Diaz account and then ...

She gets added as a Director.

image

djvoa12:
Fixes #1136

### DIFF
--- a/app/components/user-search.js
+++ b/app/components/user-search.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
   store: Ember.inject.service(),
+  classNames: ['user-search'],
   results: [],
   searchTerms: null,
   searching: false,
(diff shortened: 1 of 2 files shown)


jrjohnson:
I don't like this approach.  Using pointer events in the CSS makes it really hard to reason about why click events are not firing.  How about a an if/else block in the template that doesn't add the action or the clickable class to users which cannot be clicked?

djvoa12:
This PR is good to go!

jrjohnson:
:+1: @homu r+

PR #1159, opened 2015-11-26
capitalization cleanup

saschaben:
added title case to all (hopefully) translation phrases used as title
headers, sentence case to all action phrases, and corrected a few
missing translations….

PR #1161, opened 2015-11-26
Revert "capitalization cleanup"

saschaben:
Reverts ilios/frontend#1159


saschaben:
OK. Reverting this in the hopes that it clears the path for the other more important issues, but: let's get the translation file here plugged in asap, as the previous file is a mess.

PR #1162, opened 2015-11-27
Remove our custom test template

jrjohnson:
We're not really using this anymore and the ember one is better anyway


stopfstedt:
:+1:

stopfstedt:
@homu r+

PR #1164, opened 2015-11-30
Fix objective saving

### LINKED ISSUE #1115: Error when adding parent to course objective
https://ilios3-demo.ucsf.edu/courses/697

Tried to add `https://ilios3-demo.ucsf.edu/courses/697` as the 
As demo_faculty1 I attempted to add `Advocate for access to necessary medications, devices and services for all patients (Professionalism)` to the objective
`Describe and discuss the provision of pharmaceutical care by pharmacists in a variety of patient care settings.`

Got a 403 error probably because we are attempting to save both the course and the program year objective and the faculty does't have access to edit the program year objective.  After refreshing the page the part was recorded successfully.

### LINKED ISSUE #1143: Unable to add parent objectives to a course
Adding parent objectives to a course attempts to save both the parent and the child objectives.  This results in a permissions error.  Just save the course objective.

jrjohnson:
We only need to save one side of this relationship.  Saving both runs
into permission errors.

The whole objective area needs a refactor as it uses an old pattern for editing that doesn’t go through a buffer.  However, this fixes #1143 so I’m going to call it a day.

Also fixes #1115

### DIFF
--- a/app/components/detail-objectives.js
+++ b/app/components/detail-objectives.js
@@ -60,57 +60,51 @@ export default Ember.Component.extend({
       });
     },
     save: function(){
-      var self = this;
       if(this.get('isManagingParents')){
         let objective = this.get('mangeParentsObjective');
-        objective.get('parents').then(function(newParents){
-          let oldParents = self.get('initialStateForManageParentsObjective').filter(function(parent){
+        objective.get('parents').then(newParents => {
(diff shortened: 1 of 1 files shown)


saschaben:
:+1:

stopfstedt:
true. true.

PR #1168, opened 2015-11-30
Update to node 5

jrjohnson:
This gives us NPM 3 which should be much better on windows.
 
Also 5 > 0.  That is just a fact.

This can be tested at: http://565c91cdd6865d33c200007a.ilios-frontend.netlify.com/


dartajax:
:+1:

PR #1179, opened 2015-12-01
Fix Shibboleth Logout

jrjohnson:
Shibboleth logout is local so redirecting back to the login page just sends people back to the IDP and ends up logging them in again.  We need to leave them on the ugly shibboleth logout page.

Fixes ilios/ilios#1164


jrjohnson:
This is nearly impossible to test effectively so I'm giving it the :+1: and assigning for code review.

thecoolestguy:
@homu r+

thecoolestguy:
@homu retry

PR #1183, opened 2015-12-02
Limit deletes in the courses area

### LINKED ISSUE #1182: Delete Permission Review
I think we're currently allowing delete option to anyone and that should change.  @saschaben can you review who should be able to delete a course, session, or offering?  Any instructor currently has access to those options.

jrjohnson:
Removed the delete icon for courses, sessions, and offerings when the user is not a Developer or a director in the course, or an instructor in the offering.

Fixes #1182

### DIFF
--- a/app/components/ilios-course-list.js
+++ b/app/components/ilios-course-list.js
@@ -1,15 +1,45 @@
 import Ember from 'ember';
+import DS from 'ember-data';
 
-export default Ember.Component.extend({
+const { computed, inject, RSVP, ObjectProxy, Component } = Ember;
+const { service } = inject;
+const { PromiseObject } = DS;
+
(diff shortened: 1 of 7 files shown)


saschaben:
FYI, our ancient permissions matrix: https://www.dropbox.com/s/7cerjnm5qbfion5/permissions%20matrix%20-%20app.%20B.docx?dl=0

saschaben:
we should probably apply this logic to course level objectives as well...we will map out the specific permissions to course level / session level internal attributes for associated users in another ticket.

saschaben:
seems to work as advertised.
:+1:

stopfstedt:
:+1:

stopfstedt:
@homu r+

PR #1192, opened 2015-12-03
ember watson:convert-ember-data-async-false-relationships

djvoa12:
This task adds an explicit `async: false` options to all `belongsTo` and `hasMany` that either have no options or its options does not contain an explicit async value.


djvoa12:
@jrjohnson good to go

jrjohnson:
This should be async: true.  All of our relationships are async.  It was an uncaught bug that this was incorrect previously.

djvoa12:
@jrjohnson :+1:

PR #1194, opened 2015-12-03
ember watson:convert-ember-data-model-lookups

djvoa12:
This changes the way model lookups happen when working with Ember Data. When using `hasMany` or `belongsTo`, the first argument will become a dasherized string.


djvoa12:
@jrjohnson ready

jrjohnson:
@djvoa12 This seems to be all white space changes?  Is that correct?

djvoa12:
@jrjohnson I think so I just wanted to merge all changes that resulted from all the commands so it won't be caught again for some update in the future, if any.

jrjohnson:
:+1:

PR #1195, opened 2015-12-03
ember watson:convert-resource-router-mapping app/router.js

djvoa12:
Converts `this.resource('user')` to `this.route('user', {resetNamespace: true })` in `app/router.js`


jrjohnson:
:+1:

PR #1196, opened 2015-12-04
Update ember-cli-deploy

### LINKED ISSUE #1187: Old index files are getting deleted from S3 bucket
Only the last 10 index files are kept which is a terrible idea.  We need to update our deploy pipeline to 0.5 and all the plugs we use as well.

jrjohnson:
Updated to v 0.5.1 including plugins.  This version doesn’t limit us to
10 deploys and has many other major improvements.  As part of this move
I also consolidated our assets into a single bucket so it is easy to
CDN and versioned the index files by API version.

Fixes #1187


jrjohnson:
@thecoolestguy after this is merged there are a few changes to the process for deploying the frontend.

jrjohnson:
:+1: no user facing changes.   Just needs code review

stopfstedt:
@homu r+

PR #1199, opened 2015-12-04
Load instructors directly from event

### LINKED ISSUE #1197: Align calendar with new events API data
ilios/ilios#1172 introduces instructor names into userevents and schoolevents we need to update the calendar to work with this data.

jrjohnson:
The new user and school events endpoints return a list of instructors
directly.

Fixes #1197


saschaben:
event display doesn't seem to be showing instructor at all....

jrjohnson:
Yea, my bad.  Can't be tested on demo as it requires a change we just made.

saschaben:
:+1:

thecoolestguy:
@homu r+

PR #1200, opened 2015-12-04
Fix weird time bug with new offerings

### LINKED ISSUE #938: Small Group Generator Sets Time of Sessions to 12:00AM - 12:00AM on Frontend
Pick a session ...

newbug1

Add offering ...

newbug2

Select Small Groups ...

newbug3

Pick Group (ACM)

newbug4
 Create offerings ...

newbug5

Click to edit one of them ... the start end times look fine ... save and close ..

newbug6

Now pick a different one to edit -- 12:00AM - 12:00AM issue.

newbug7

jrjohnson:
Copy the start and end date objects so they are not shared

If we don’t do this then upon initial creation a change to one offering
would effect all of them.

Fixes #938

Can be tested here: http://5661e64bd6865d251100015b.ilios-frontend.netlify.com/

### DIFF
--- a/app/components/session-offerings.js
+++ b/app/components/session-offerings.js
@@ -2,7 +2,7 @@ import Ember from 'ember';
 import layout from '../templates/components/session-offerings';
 import { translationMacro as t } from "ember-i18n";
 
-const { Component, computed, inject, isPresent, RSVP } = Ember;
+const { Component, computed, inject, isPresent, RSVP, copy } = Ember;
 const { service } = inject;
 const { alias, oneWay } = computed;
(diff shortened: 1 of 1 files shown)


dartajax:
That netlify link does not seem to be working correctly.  There is an error on the dashboard and then after clicking Courses and Sessions ...

`Ember Inspector Active
VM305:91 Ember Inspector ($E):  v {routeName: "error", __nextSuper: undefined, __ember1449512579071: null, __ember_meta__: Object}`

dartajax:
:+1:

PR #1208, opened 2015-12-05
refactored global-search

### LINKED ISSUE #1190: Search Results (Global) Do Not Return Correct Data
What this means is ...

When you log in to Ilios and immediately go to Admin and then do a search for an entity that you are confident exists -- in this case, I searched for myself.

See the screen shot below. It orders by Last Name, First Name asc so I am showing the place where Leonard, David should exist.

image

The other search available from ....

image

... seems to work.

image

Now if I go back to the Global Search by clicking Admin as shown, 

image

I am still getting the bad data ...

image

But ...

Eventually it works.  I obtained the correct results by futzing with the search parameter (using Leona, Leon, Leo, Leonar) various thngs like that and eventually the results matched what was returned in the other "Manage Users" search.

djvoa12:
Fixes #1190

### DIFF
--- a/app/components/global-search.js
+++ b/app/components/global-search.js
@@ -1,82 +1,78 @@
 import Ember from 'ember';
 
-const { Component, inject, run, computed } = Ember;
+const { Component, computed, inject, isEmpty, run } = Ember;
 const { service } = inject;
 const { debounce } = run;
-const { or, notEmpty } = computed;
+const { trim } = Ember.$;
+const { PromiseObject } = DS;
+
+function cleanQuery(query) {
+  return trim(query).replace(/[\-,?~!@#$%&*+\-'="]/, '');
+}
 
 export default Component.extend({
   store: service(),
   i18n: service(),
+
   classNames: ['global-search'],
-  results: [],
-  searching: false,
+
+  query: null,
+  delay: 500,
+
   showMoreInputPrompt: false,
-  showNoResultsMessage: false,
-  currentlySearchingForTerm: false,
-  hasResults: notEmpty('results'),
-  showList: or('searching', 'showMoreInputPrompt', 'showNoResultsMessage', 'hasResults'),
+
+  revisedQuery: computed('query', {
+    get() {
+      const query = this.get('query');
+      const revisedQuery = cleanQuery(query);
+      const validate = this.checkQueryLength(revisedQuery);
+
+      if (validate) {
+        this.set('showMoreInputPrompt', false);
+      } else {
+        this.set('showMoreInputPrompt', true);
+      }
+
+      return revisedQuery;
+    }
+  }).readOnly(),
+
+  _updateQuery(value) {
+    this.set('query', value);
+  },
+
+  checkQueryLength(query) {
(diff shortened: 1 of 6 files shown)


djvoa12:
@jrjohnson I need a code review and your thoughts on an issue. If I type in `leonard` in the search (under demo3 proxy), I don't get any results. Backend issue?

jrjohnson:
The names on demo are scrambled so I'm guessing there is no Leonard there.  I will take a look at the code (probably tomorrow morning)

app/components/global-search.js
```diff
@@ -1,82 +1,78 @@
...
 
-const { Component, inject, run, computed } = Ember;
+const { Component, computed, inject, isEmpty, run } = Ember;
 const { service } = inject;
 const { debounce } = run;
-const { or, notEmpty } = computed;
+const { trim } = Ember.$;
+const { PromiseObject } = DS;
+
+function cleanQuery(query) {
```
jrjohnson: Splitting this up here is cleaner, but since it is only used once it should just go in the revisedQuery CP.  Otherwise you have to go looking for it when checking to see what it does.

We're not using this anywhere else, is it necessary?  Or just something you added for good measure?

app/components/global-search.js
```diff
@@ -1,82 +1,78 @@
...
+
+      return revisedQuery;
+    }
+  }).readOnly(),
+
+  _updateQuery(value) {
+    this.set('query', value);
+  },
+
+  checkQueryLength(query) {
```
jrjohnson: Same thing here, this is a single use function that required me to go looking aroudn for what it did.  Since what it does is super simple it can just be part of the CP.

app/components/global-search.js
```diff
@@ -1,82 +1,78 @@
...
-      } else if(noWhiteSpaceTerm.length < 3){
-        this.setProperties({
-          results: [],
-          showMoreInputPrompt: true,
+
+      return revisedQuery;
+    }
+  }).readOnly(),
+
+  _updateQuery(value) {
```
jrjohnson: We're not using this convention of creating private methods with an _ prefix.  This should just be updateQuery.

jrjohnson:
This looks good, couple of organizational things to clean up.  I didn't test the functionality, just reviewed the code.

app/components/global-search.js
djvoa12: It's actually being used in another place so I am going to put the function in `utils`.

app/components/global-search.js
djvoa12: Done.

app/components/global-search.js
djvoa12: Done.

jrjohnson:
I tested this and it seems to be working and the code looks good.  I want to wait to merge it until we can test against stage where the names are a little easier to validate and where the original issues were showing up.  Stage will need to be updated first to deal with new config location.

jrjohnson:
:+1: I tested against demo - testing against stage is way harder than it should be.

jrjohnson:
@homu r+

PR #1211, opened 2015-12-07
Hit the right endpoint for token refresh

### LINKED ISSUE #1210: Token not being refreshed at the correct url
We're still using the old /auth/refresh instead of /auth/token when our local token expires and we need to check in the with API for a new one.

jrjohnson:
Fixes #1210


thecoolestguy:
@homu r+

PR #1212, opened 2015-12-07
ember watson:replace-needs-with-injection app

djvoa12:
Convert `needs` declarations the individual properties using the new `Ember.inject.controller()` feature. Also convert any uses of the `controllers` hash to use the newly defined properties.


jrjohnson:
:+1: @homu r+

PR #1213, opened 2015-12-07
Improved Session List

### LINKED ISSUE #810: ILM label should display in session list
currently, ILMs are displayed in identical manner to normal sessions; instead, we should be labeling them in the "first offering" column with the due date prepended by "ILM: Due "

jrjohnson:
Made a couple of improvements:
- Add a loading indicator to session list
- Improved loading of session list first offering time and session type
- Display ILM prefix in session first offering column (fixes #810)

### DIFF
--- a/app/components/ilios-sessions-list.js
+++ b/app/components/ilios-sessions-list.js
@@ -113,6 +113,7 @@ export default Component.extend({
 
   saved: false,
   savedSession: null,
+  isSaving: null,
 
   actions: {
     toggleEditor() {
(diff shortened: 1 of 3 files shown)


saschaben:
not ready for testing, throws undefined errors on attempt to load login page when proxied, throws multiple 404 errors when not proxied.

saschaben:
- [ ] the __ILM: Due By__ label is pulling in incorrect timestamps along with date (for current date, it uses currentTime; for other dates, using 12am). either no timestamp at all should be displayed in this release, or 5pm should be displayed.

- [ ] the initial loading indicator for the list makes it feel like it is actually taking longer to load....I think this is just a psychological quirk, but, let's consider. the internal loaders and loading rocks.

jrjohnson:
@saschaben ready for review again

saschaben:
initial load much better with count.

Am getting some significant choppy lag when creating a new session and saving -- the list reload seems to interrupt the save session transaction and return.

jrjohnson:
@saschaben I see the delay you're talking about.  My preference would be to merge this and make that a separate bug as its kind of sideways related.  I can fix it here though if you don't want to merge without it.

jrjohnson:
@saschaben - fixed the saving of a session - give it another look.

saschaben:
good enough for government work

:+1:

thecoolestguy:
@homu r+

PR #1215, opened 2015-12-07
Add ember-inflector to dependancies

jrjohnson:
We’re using this in our reporting code now so we need to actively
include it.


thecoolestguy:
@home r+

thecoolestguy:
@homu r+

PR #1216, opened 2015-12-07
Better learner group list

### LINKED ISSUE #991: Program (when not selected) should default to “Select a Program …”.
On Learner Groups page, initial selection of program/cohort should have a default as above. Otherwise, we will always load the wrong program....

jrjohnson:
Improved the select lists to be regular selects which load a little bit better.

The program has to be selected each time until we can force a default selection based on user preferences.

Fixes #991


saschaben:
:+1:

thecoolestguy:
@homu r+

thecoolestguy:
@homu retry

PR #1217, opened 2015-12-07
ember watson:find-overloaded-cps app

djvoa12:
Helps you locate all the places where your source may trigger the "Using the same function as getter and setter" deprecation.


jrjohnson:
:+1: yup, looks like dead code

PR #1218, opened 2015-12-08
 removed Ember.HTMLBars.makeBoundHelper deprecation

djvoa12:
Deprecation Warning:
- Using `Ember.HTMLBars.makeBoundHelper` is deprecated. Please refactor to using `Ember.Helper` or `Ember.Helper.helper`.

### DIFF
--- a/app/helpers/is-equal.js
+++ b/app/helpers/is-equal.js
@@ -1,11 +1,27 @@
 import Ember from 'ember';
 
-export function isEqual(params) {
-  if(params.length < 2){
-    return false;
-  }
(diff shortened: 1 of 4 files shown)


jrjohnson:
These should follow the pattern in https://github.com/ilios/frontend/blob/master/app/helpers/is-in.js which uses and observer to ensure the helper updates when the value updates.  This is considered good use of an observer.

djvoa12:
@jrjohnson need code review.

jrjohnson:
:+1: looks good

PR #1221, opened 2015-12-08
removed @each dependent key

djvoa12:
Deprecation Warning:
- /Depending on arrays using a dependent key/


jrjohnson:
:+1: looks good

Issue #1201, opened 2015-12-04
My Reports Issue - Sessions / Courses Associated with [Instructor Name] Not Returning Correct Results


dartajax:
I searched to do a report on myself (an Instructor in Ilios) and here is what I get.

![image](https://cloud.githubusercontent.com/assets/7493331/11599261/9fda63f8-9a7a-11e5-9ca7-0a7bb235e2b3.png)

That's it .. no Leonard.  And when you put in Leonard, you get 0 results.  That is not correct.

![image](https://cloud.githubusercontent.com/assets/7493331/11599281/c4a983e4-9a7a-11e5-964d-90fdf120d01f.png)

![image](https://cloud.githubusercontent.com/assets/7493331/11599218/65152028-9a7a-11e5-9365-c293472c06aa.png)

dartajax:
I was testing this with last names 'Bruer' and also 'Graff' -- both of these are Pharmacy instructors and are attached to quite a few sessions and courses - no results returned.

PR #1222, opened 2015-12-08
Make all Users available as report instructor

jrjohnson:
The instructor role isn’t important here and makes it confusing when
the person you are looking for cannot be found in the list.

Refs #1201


jrjohnson:
@saschaben please confirm that this change is OK.  It would now be possible to Run a report for all courses instructed by a student, but I don't see the harm in that.  Especially if they are an instructor without the necessary role.

saschaben:
:+1:

thecoolestguy:
@homu r+

PR #1226, opened 2015-12-08
Handle user middle name correctly

### LINKED ISSUE #1225: User model fails when user does not have a middle name
https://github.com/ilios/frontend/blob/master/app/models/user.js#L105 needs to handle null values correctly.

jrjohnson:
Don’t capitalize it automatically
Don’t break when it is missing.

Fixes #1225


saschaben:
:+1:

thecoolestguy:
@homu r+

PR #1231, opened 2015-12-09
Fix adding new file learning material

### LINKED ISSUE #1230: Unable to add new file learning material
The request goes through but the file itself is not saved.

jrjohnson:
We were not sending enough information to attach the file to the record.
This sends the fileHash with new learning materials

Fixes #1230

### DIFF
--- a/app/serializers/learning-material.js
+++ b/app/serializers/learning-material.js
@@ -1,9 +1,19 @@
+import Ember from 'ember';
 import DS from 'ember-data';
+
+const { get } = Ember;
 
 export default DS.RESTSerializer.extend({
   isNewSerializerAPI: true,
   serialize: function(snapshot, options) {
     var json = this._super(snapshot, options);
+
+    //When POSTing new file learningMaterials we need to include the file hash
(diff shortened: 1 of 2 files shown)


jrjohnson:
Should be testable against demo.

saschaben:
actually, not testable against demo, as demo doesn't allow the upload of files. needs to be tested on stage for validation.

jrjohnson:
@saschaben I just validated as working against the demo API.  Seems to allow the creation of new files.  Is that something we need to fix / prevent?

jrjohnson:
This should be tested with a small file or else wait a few seconds after uploading before clicking save.  Separate ticket #1232 for handling the delay correctly.

saschaben:
Looks good now. still throwing a 500 error when attempting to download from the calendar event detail page though (but not from the course management page).

jrjohnson:
:+1: from @saschaben. Can you create another issue for the detail download?

saschaben:
can't recreate. nm

thecoolestguy:
@homu r+

PR #1233, opened 2015-12-09
Improved user search

### LINKED ISSUE #1228: user search returns limited list of values
when you run a user search, the results list is truncated to only a handful of values, rather than returning all corresponding results. example: login as demo_dev1, go to admin, and search for gutierrez. Nicholas will not show up, since "N" comes after the point of list truncation.

jrjohnson:
Increased the query limit to 100 and added a count to the top of the results.  The count should let the know they need to add more terms without us explicitly saying that, but we can add that later if necessary.  For now this fixes #1228.

### DIFF
--- a/app/components/global-search.js
+++ b/app/components/global-search.js
@@ -48,7 +48,8 @@ export default Component.extend({
         const searchResults = store.query('user', {
           q,
           'order_by[lastName]': 'ASC',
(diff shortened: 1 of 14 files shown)


jrjohnson:
This is now ready for review and should work again demo.

saschaben:
:+1:

PR #1240, opened 2015-12-10
Improve handling of learning material file uploads

### LINKED ISSUE #1232: Race condition uploading files
Currently the file uploader starts working when a file is selected, but with a large file or a slow internet connection it doesn't finished quickly.  If the user clicks save on the learning material before the uploader finishes then they end up saving junk.  
Once #1227 is merged we can address this in a ways - 
1) Require a file be uploaded before allowing a save the same as we do for other validations
2) Add a progress bar to the file upload so that users will have an indication when it is done
3) Even is save is pressed block action until we get back a valid file hash

jrjohnson:
- [x] If the file has not finished uploading don’t allow the LM to be saved
- [x] When the file is being uploaded show the progress
- [x] When the file is done uploading show a check mark

Fixes #1232

### DIFF
--- a/app/components/file-upload.js
+++ b/app/components/file-upload.js
@@ -9,13 +9,20 @@ export default FileField.extend({
   filesDidChange(files) {
     const uploadUrl = this.get('url');
     const uploader = Uploader.create({ url: uploadUrl });
-
+    
+    this.sendAction('startUploading');
+    
     uploader.on('didUpload', (e) => {
       this.sendAction('finishedUploading', e);
+    });
+    
+    uploader.on('progress', (e) => {
+      this.sendAction('setUploadPercentage', e.percent);
     });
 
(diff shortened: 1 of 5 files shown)


jrjohnson:

saschaben:
getting "bad request" error when attempting to upload the following file: https://www.dropbox.com/s/w4t3yhn6nqz0fk3/CDLG_Presentation_12-19-2012.pptx?dl=0

jrjohnson:
I'm guessing this as an issue with file size on demo - we do need to handle this better I created this ticket in the API to help with that: 
https://github.com/ilios/ilios/issues/1196

We also should validate file size before starting the upload, that ticket is now here:
https://github.com/ilios/frontend/issues/1245

saschaben:
:+1:

saschaben:
code review and merge please.

PR #1250, opened 2015-12-14
implemented feature switch functionality

djvoa12:
`IliosFeatures` property's value (from `config/environment`) is an object which can contain a list of features that should be turned on/off based on their own unique situations.


jrjohnson:
:+1:

PR #1258, opened 2015-12-20
Fix some ILM Timezone Madness

jrjohnson:
ILM Sessions are coming from the API with a UTC time, but that isn’t
really true.  They have no real timestamp.  This forces them to be saved and to appear at 5pm on the correct date.


jrjohnson:
@saschaben this can be tested against demo.

saschaben:
:+1:

saschaben:
:+1:

stopfstedt:
LGTM :+1:

PR #1275, opened 2015-12-22
Add some skew to our queries for calendar events

jrjohnson:
Looking a few days ahead and behind loads a little bit more data than we need but insures we don’t get into any timezone trouble.

Adding 6 days to the start and end of a month allows us to populate
days that appear on the calendar like the 30th before the 1st.

### DIFF
--- a/app/components/dashboard-calendar.js
+++ b/app/components/dashboard-calendar.js
@@ -44,10 +44,17 @@ export default Component.extend({
     this.set('selectedCourses', []);
   }),
   fromTimeStamp: computed('selectedDate', 'selectedView', function(){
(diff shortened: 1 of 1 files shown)


saschaben:
do we need to do a similar skew for week loads (1 + week + 1) and days (1 + day + 1) as well, to avoid loss on sat/sun events or midnight events?

jrjohnson:
The skew is on everything. 1 day front and back if days and weeks. 5 days front and backing r months.

stopfstedt:
partial fix. 
events will now show correctly across months boundaries now in the _weekly view_.
date skew does not seem to affect the _month view_, perhaps an issue with the calendar itself and not with the query range.
in any case - this is good to be merged in.

stopfstedt:
:+1:

PR #1288, opened 2016-01-01
Fix broken publish and schedule functionality

jrjohnson:
The publish event was never getting saved so nothing was getting
published.

### DIFF
--- a/app/mixins/publishable.js
+++ b/app/mixins/publishable.js
@@ -1,6 +1,6 @@
 import Ember from 'ember';
 
-const { RSVP, inject, computed } = Ember;
+const { inject, computed } = Ember;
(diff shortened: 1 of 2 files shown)


saschaben:
This looks to be solid so far from the UX side; testing proxied against demo means not being able to verify the data in the publish_event table vs. the course/session tables, but functionally this appears to be on it.

saschaben:
:+1:

saschaben:
@jrjohnson if this is ready now with the test modification for merge, let me know. (You still have it tagged as 'in progress')

jrjohnson:
Waffle does that tagging.  It's ready as far as I am concerned.

PR #1289, opened 2016-01-04
Update ember-simple-auth version

jrjohnson:
With the release of ember-simple-auth 1.0 several changes are required.  This is a critical update on our path to Ember 2.0 as well as allowing for a simplification and paves the way for new features of the currentUser service.

Outstanding issues after testing:
- [x] Shibboleth login not setting the token correctly so subsequent requests (like user events) don't get an auth header.
- [x] requests to /errors are not authenticated (possibly the same as the shibboleth issue)


jrjohnson:
This PR is currently deployed to https://ilios-stage.ucsf.edu/ so the shibboleth part can be tested there.

stopfstedt:
login works. 

logout works, although using the browser's back button from the SSO logout page will log you back. 

i also noticed that the application throws errors after logging out, before redirecting to the SSO logout page.

selection_071

jrjohnson:
Back button logout seems to be an existing issue in production.  Must be hanging on to a cookie.  I will investigate.

jrjohnson:
I created a new issues #1304 for the back button problem.  Will investigate separately.

PR #1294, opened 2016-01-06
Match publishing with the API

jrjohnson:
Publish event is removed from the API in ilios/ilios#1212 so things are just published now.

This should be merged with the corresponding API ticket.


stopfstedt:
this makes things so much simpler - my favourite kind of code change.

LGTM :+1:

PR #1303, opened 2016-01-07
Scheduled events on the calendar

jrjohnson:
Events published as TBD are coming out of the API with way less info now.   This handles them better.

### DIFF
--- a/app/components/dashboard-calendar.js
+++ b/app/components/dashboard-calendar.js
@@ -72,13 +72,22 @@ export default Component.extend({
             let sessionEvents = {};
             let promises = [];
             events.forEach(event => {
-              promises.pushObject(this.get('schoolEvents').getSessionForEvent(event).then(session => {
-                let sid = session.get('id');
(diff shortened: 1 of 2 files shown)


jrjohnson:
@saschaben please test the scheduled events on a calendar as a student.  They should be un-clickable and contain very little information.  Small Group scheduled events should appear correctly in month, week, and day calendars.

jrjohnson:
@stopfstedt can you take a look at this?  It requires using HEAD for the API so @saschaben can't test it easily.

stopfstedt:
reviewed. 
i'm confirming that grouping scheduled events that fall into the same time slot are displayed as one event in week/month view. _this part works_.

however, unclickability has not been achieved. i can still click through to the details page of what appears the first of these deflated, scheduled events.

jrjohnson:
Try an npn install to update the calendar dependency

stopfstedt:
that fixes that. all good now, thanks!

PR #1306, opened 2016-01-08
fixes regression in calendar filtering due to blanked out scheduled events.

stopfstedt:
schedules events don't carry information about their origin, hence attempting to load their owning session/course data fails and breaks the filtered calendar.

omitting them during filtering fixes this regression.


jrjohnson:
This would probably make sense to refactor into a computed property that just returned events with info. But this is a good solution for now. When tests pass I will merge.

jrjohnson:
Need rebase

PR #1309, opened 2016-01-09
Lock travis node version to 5.1

jrjohnson:
Some problems building with the latest node version.

This is speculative, might work, might not.  We’ll see.


jrjohnson:
Nope

PR #1324, opened 2016-01-13
display owner rather than author in learning material detail list

gboushey:
Fixes #1258.

Changed the value from lmProxy.learningMaterial.originalAuthor to  lmProxy.learningMaterial.owningUser.fullName.

Added user fields to fixtures to test in tests/acceptance/course/session/learningmaterials-test.js and updated acceptance tests.  

note - I think this issue may also be present in print-course.hbs


jrjohnson:
@gboushey The failing test is due to a very similar set of tests in tests/acceptance/course/learningmaterials-test.js They look the same, but one tests learning materials attached to a session and the other tests them attached to a course.

PR #1325, opened 2016-01-15
Flag disabled learner group users

### LINKED ISSUE #1281: Adding Learner Group to ILM Sessions Raises Error For SOP User
The save does get completed upon refreshing the screen however.

I did not have the same issue.

The user has user_role values of 1,2,3 and did not have any entries in the "permission" table.

### LINKED ISSUE #1282: Learner Group Management Displays Inactive Accounts
image

image

In this case, it is clear which one is the correct one since there are bogus email accounts for the two wrong accounts but they also are disabled if you look them up in the user table.

jrjohnson:
Makes it possible to correctly identify disabled users.

Fixes #1282


saschaben:
:+1:

PR #1332, opened 2016-01-17
Use ember's pushObjects function to extract array

jrjohnson:
Doing it with the forEach in the loop causes a jshint error, this
should be equivalent.

Fixes problem with current test failures.  Offering manager should be tested to ensure nothing weird results.


stopfstedt:
LGTM :+1:

PR #1334, opened 2016-01-18
Clean up deprecation workflow config

jrjohnson:
Added comments for where remaining deprecations are coming from.   So this will be easier to handle.


saschaben:
:+1:  LGTM. Still getting 2 deprecations from the calendar on load:
```
1. Using `Ember.HTMLBars.makeBoundHelper` is deprecated. Please refactor to using `Ember.Helper` or `Ember.Helper.helper`.
Trace in the console
2. Usage of Ember.computed.any is deprecated, use `Ember.computed.or` instead.
````

PR #1340, opened 2016-01-19
Display correct learning material owner in course print view

### LINKED ISSUE #1339: author is displayed in print course, should be owner
In the table for learning material detail, the original author is displayed where the owner should be displayed.  

Similar error to #1258.

gboushey:
owner field in print summary should display owner name rather than original author name.  no tests currently exist for this page.

fixes #1339.

### DIFF
--- a/app/templates/components/print-course.hbs
+++ b/app/templates/components/print-course.hbs
@@ -232,7 +232,7 @@
                     {{lm.learningMaterial.title}}
                   </td>
                   <td class='text-center'>{{lm.learningMaterial.type}}</td>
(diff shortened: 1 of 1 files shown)


stopfstedt:
LGMT :+1:

stopfstedt:
on second, inspection - this only catches session LMs. course LMs also need to have their displayed owner corrected.

@gboushey - same file but further up needs the same fix.

https://github.com/gboushey/frontend/blob/incorrect-field-1339/app/templates/components/print-course.hbs#L125

PR #1344, opened 2016-01-20
Fix for navigation delay

### LINKED ISSUE #1331: navigation from page has significant delays
When navigating away from certain pages, there is a major delay in action. In order of slowness:

1. from the dashboard to anywhere when calendar filters are OPEN.
2. From Session to anywhere when Offering edit is OPEN.
3. From Session to anywhere.
4. From course when course details are fully expanded.

Items 3 & 4 are not significantly ugly in normal use, but items 1 & 2 are. These may be one of the primary causes of current difficulties with performance.

jrjohnson:
The dashboard delay appears to be caused by garbage collection on the large number of checkbox component we had on the page.  I replaced them with plain input elements which seems to be much faster.

Fixes #1331

### DIFF
--- a/app/components/boolean-check.js
+++ b/app/components/boolean-check.js
@@ -7,9 +7,5 @@ export default Component.extend({
 
   tagName: 'span',
 
-  value: false,
-
-  click() {
-    this.sendAction();
-  }
+  value: false
 });
--- a/app/components/dashboard-calendar.js
+++ b/app/components/dashboard-calendar.js
@@ -64,7 +64,7 @@ export default Component.extend({
         promise: this.get('userEvents').getEvents(this.get('fromTimeStamp'), this.get('toTimeStamp'))
       });
     } else {
-      let deferred = Ember.RSVP.defer();
+      let deferred = RSVP.defer();
       this.get('selectedSchool').then(school => {
         this.get('schoolEvents').getEvents(school.get('id'), this.get('fromTimeStamp'), this.get('toTimeStamp')).then(events => {
           if(this.get('selectedView') === 'day'){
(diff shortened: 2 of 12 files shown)


jrjohnson:
@saschaben while this isn't done yet, can you give it a quick test to ensure that you are seeing the same performance that I am.  I can now navigate from the dashboard with filters open and zero delay.

saschaben:
@jrjohnson I am seeing what you are seeing: great improvement! If we can apply similar cleanup to the other areas of delay, we should be in the homestretch.

saschaben:
(I am however getting an error when I collapse the calendar ("hide calendar") with filters open -- it's throwing that `Cannot read property 'then' of undefined` error. uncertain if this is just a PR branch issue, but let's be sure not to regress this cockroach back into the code.)

jrjohnson:
It looks like we may be running into a solved chrome bug:
https://bugs.chromium.org/p/v8/issues/detail?id=2935#c27

I'm testing in some other platforms.  Also speed will improve with Ember 2.0 as well.

jrjohnson:
I think the dashboard is as far as this PR is going to take us.  Navigating away from a session is going to require some structural changes to that page.

@saschaben I can't reproduce your bug.  I just rebased this branch so if you can check it out again and run `nomless` see if you still see the behavior.

saschaben:
The bug is consistent after rebase and `nomless`. to replicate:

proxy to demo.
goto dashboard.
open filters.
switch to "all events".
select topic filter "anatomy"
click event at 1/20 3pm to view detail page.
return to dashboard.
hide calendar.

Error page is returned `Cannot read property 'then' of undefined`.

jrjohnson:
@saschaben Ok - I think I fixed this.  Give it another try.

saschaben:
huzzah :+1:

jrjohnson:
@stopfstedt tests have finally passed, this is ready for review

stopfstedt:
code looks good, click-testing now.

PR #1348, opened 2016-01-21
Prevent errors from throwing errors

jrjohnson:
When we had a problem posting to the /errors API we got into a bad loop
of repeated exponential errors.  Prevent that.  Also added a time which
will only allow one error to be posted every second.


stopfstedt:
reviewed and tested. LGTM :+1:

PR #1349, opened 2016-01-22
load course/session LMs by filtering on owning course.

### LINKED ISSUE #1342: Draft Status Learning Materials Are Not Being Hidden From Learners
1.

See the second learning material has been saved as Draft -- Logged in as me.
image

2.

Spoofing a student ...

image

stopfstedt:
fixes #1342 

corresponds to ilios/ilios#1272.

### DIFF
--- a/app/components/ilios-event.js
+++ b/app/components/ilios-event.js
@@ -127,7 +127,7 @@ export default Component.extend({
       promise: defer.promise
     });
   }),
-  courseLearningMaterials: computed('i18n.locale', function() {
+  courseLearningMaterials: computed('i18n.locale', 'session.course', function() {
     let defer = Ember.RSVP.defer();
 
     this.get('thesession').then(session => {
@@ -221,7 +221,7 @@ export default Component.extend({
     });
   }),
 
-  sessionLearningMaterials: computed('i18n.locale', function() {
+  sessionLearningMaterials: computed('i18n.locale', 'session', function() {
     let defer = Ember.RSVP.defer();
 
     this.get('thesession').then(session => {
--- a/app/components/ilios-event.js
+++ b/app/components/ilios-event.js
@@ -127,49 +127,50 @@ export default Component.extend({
       promise: defer.promise
     });
   }),
-  courseLearningMaterials: computed(
-    'i18n.locale',
-    'offering.session.course.learningMaterials.@each.learningMaterial.[title,absoluteFileUri]',
-    function(){
-      let defer = RSVP.defer();
+  courseLearningMaterials: computed('i18n.locale', function() {
+    let defer = Ember.RSVP.defer();
 
-      this.get('thesession').then(session => {
-        session.get('course').then(course => {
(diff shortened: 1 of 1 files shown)


stopfstedt:
the corresponding backend fix introduces an interesting regression in the course print view, resulting in a 403 when requesting session/course LM data for in-draft materials as student.

we have to load LMs there too as a list, filtered by course/session id.

stopfstedt:
**BLOCKED BY #1350**

jrjohnson:
I think this can actually be merged at any time as it doesn't require the API changes.  I'm going to respect the DO NOT MERGE label for now, but this looks ready to merge to me.

app/components/ilios-event.js
```diff
@@ -127,49 +127,50 @@ export default Component.extend({
       promise: defer.promise
     });
   }),
-  courseLearningMaterials: computed(
-    'i18n.locale',
-    'offering.session.course.learningMaterials.@each.learningMaterial.[title,absoluteFileUri]',
-    function(){
-      let defer = RSVP.defer();
+  courseLearningMaterials: computed('i18n.locale', function() {
```
jrjohnson: This computed property should have a dependency on `thesession.course`

app/components/ilios-event.js
```diff
@@ -219,47 +220,49 @@ export default Component.extend({
...
-    'offering.session.learningMaterials.@each.learningMaterial.[title,absoluteFileUri]',
-    function(){
-      let defer = RSVP.defer();
 
-      this.get('thesession').then(session => {
-        session.get('learningMaterials').then(sessionLearningMaterials => {
-          let promises = [];
-          let mappedLearningMaterials = [];
-          sessionLearningMaterials.forEach(sessionLearningMaterial => {
+  sessionLearningMaterials: computed('i18n.locale', function() {
```
jrjohnson: This computed property should be dependent on `thesession`

app/components/ilios-event.js
stopfstedt: yes, i did that originally. however, this will cause the same XHR for retrieving course LMs multiple times.
since its a computed function, I suspect this method is firing off every time `thesession.course` changes. 
any idea as how to overcome this?

jrjohnson:
Interesting. I think that is what we were running into yesterday. We probably need a course computed property which would change less often. Let's just leave this as is. I have it in my list to cleanup.

jrjohnson:
Oh. It need a if(course) block. Otherwise it will fire with a blank course and pull a thousand LMs.

stopfstedt:
changed it to be depending on `session` and `session.course` instead of `thesession`, `thesession.course`. that seems to be in line with other computed properties and gets rid of the redundant requests.

stopfstedt:
@jrjohnson  please review once more. thanks!

jrjohnson:
Ok I created #1355 as a reminder to clean this up.  Since this is working and in line with the rest of the component I'm good to merge.  Just waiting on tests.

PR #1352, opened 2016-01-22
Remove draft learning materials from print view

### LINKED ISSUE #1350: don't show in-draft LMs in course print view
course/session-LMs need to be loaded differently for course print summary, see comment section in #1349 regarding the 403 triggering regression.

jrjohnson:
Students only have access to those learning materials that are final or revised.  By querying for them manually we ensure that we only get the ones the user has access to.

Fixes #1350


stopfstedt:
:+1: LGTM

PR #1356, opened 2016-01-25
Load a limited set of sessions on the course view

### LINKED ISSUE #1317: truncate session list to improve load time
instead of loading all sessions for a course, limit initial load to 50 sessions, and provide a "load more sessions" button at the bottom of the list to load the remainder. Button should be labeled "Show XX More Sessions"

And now that we have column sorting, initial sort on load should be date ascending if at all possible.

sessionlist

jrjohnson:
Instead of loading all sessions load only 50 of them.  This speeds up rendering of the sessions list on the course page substantially.

Fixes #1317

WIP: Still needs some translations

### DIFF
--- a/app/components/pagedlist-controls.js
+++ b/app/components/pagedlist-controls.js
@@ -1,0 +1,73 @@
+import Ember from 'ember';
+
+const { computed } = Ember;
+const { lte } = computed;
+
+export default Ember.Component.extend({
+  classNames: ['pagedlist-controls'],
+  tagName: 'div',
+  offset: null,
+  limit: null,
+  total: null,
+  start: computed('offset', function(){
+    return parseInt(this.get('offset')) + 1;
+  }),
+  end: computed('offset', 'limit', function(){
+    const total = this.get('total');
+    let end = parseInt(this.get('offset')) + parseInt(this.get('limit'));
+    if(end > total) {
+      end = total;
+    }
+
+    return end;
+  }),
+  offsetOptions: computed('total', function(){
+    const total = this.get('total');
+    const available = [10, 25, 50, 100, 200, 400, 1000];
+    let options = available.filter(option => {
+      return option < total;
+    });
+    options.pushObject(available[options.length]);
+
+    return options;
+  }),
+  lastPage: computed('total', 'limit', 'offset', function(){
+    const total = parseInt(this.get('total'));
+    const limit = parseInt(this.get('limit'));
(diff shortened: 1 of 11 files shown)


saschaben:
This looks pretty good! a few notes:

1. Per #1317 ,  Can we provide the initial load with a sort by first offering date, rather than title? This is pretty important.
2. The show more tab should have a 1-pixel `<HR>` to hang off of, same as the course details tab:
tab1
vs. 
tab2
3. On return from session detail, is there any chance of returning to the fully loaded session list, rather than back at the truncated list? This could be a little confusing if you are in the details of a session that isn't in the first 50 and then return to not see it....
4. The current list load is a little choppy: is there a way to not unload the initial 50 as we load the full list, or do we have to requery the whole thing to make this work? This is not a super big deal, but would be nice.

jrjohnson:
Going to attempt to do this as ann Mainstream scroll. All the sessions will be loaded right away but only 30 or so will actually display. Sorting would be on the entire list.

jrjohnson:
@saschaben please take another look at this. I went with a 'search results' type of paged interface which we kind of talked about and discarded, however I think it works as an interim solution. I'm hoping it can work for us until we have time to figure out infinite scrolling.  Probably needs some styling improvements, but I wanted to get some input before I went to far down the road.

jrjohnson:
Of note - final solution will be implemented with https://github.com/emberjs/ember-collection which is still alpha software, but should be stable fairly soon.  That provides infinite scroll rendering that we won't have to create on our own.

saschaben:
Hmm. I was skeptical of this approach, but it may work. I think that the sorting presents some opportunity in this presentation for user confusion, but I think maybe before we decide, this is one we get in front of some power users and see what they think of this approach. It certainly resolves the load time issue.

app/components/pagedlist-controls.js
```diff
@@ -0,0 +1,73 @@
+import Ember from 'ember';
+
+const { computed } = Ember;
+const { lte } = computed;
+
+export default Ember.Component.extend({
+  classNames: ['pagedlist-controls'],
+  tagName: 'div',
+  offset: null,
```
stopfstedt: perhaps provide sensitive defaults for offset, total and limit.

stopfstedt:
overall :+1:

app/components/pagedlist-controls.js
stopfstedt: ^ scratched after further discussion.

PR #1371, opened 2016-02-02
Collapse Objectives

### LINKED ISSUE #1360: Collapse course and session objectives by default
The objectives list is very long and should be collapsed by default.  A URL parameter should be used to persist the state.  showCourseObjectives, showSessionsObjectives and should be false by default.

Mock of collapsed objectives:
collapsed objectives

jrjohnson:
The objective list is very long, this collapses it by default into a much smaller space.

Fixes #1360

### DIFF
--- a/app/components/big-text.js
+++ b/app/components/big-text.js
@@ -23,8 +23,9 @@ export default Component.extend({
     }
   }),
   cleanText: computed('text', function(){
+    const text = this.get('text') || '';
     //strip any possible HTML out of the text
-    return this.get('text').replace(/(<([^>]+)>)/ig,"");
+    return text.replace(/(<([^>]+)>)/ig,"");
   }),
   displayText: computed('cleanText', 'totalLength', 'length', 'expanded', function(){
     let cleanText = this.get('cleanText');
--- a/app/components/collapsed-objectives.js
+++ b/app/components/collapsed-objectives.js
@@ -1,0 +1,20 @@
+import Ember from 'ember';
+
+const { Component, computed } = Ember;
+const { alias, sort } = computed;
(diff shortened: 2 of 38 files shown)


jrjohnson:
@saschaben please review - slightly different from the mock as I left out the summary on the left.  That's more of a style challenge than a code one - I just don't know how to fit everything in if we use the left margin for a summary.

I was also thinking about limiting the display to the first 5 objectives with a show more box at the bottom.  I don't know if that is worth doing or event a good idea - thoughts?

saschaben:
Good start, but still a little challenged.

Styling / UX issues:
 - layout when collapsed is very confusing: it's hard to identify what I am looking at. Individual objectives visually run together. And it still takes up too much space.

If we are going to show the objective text when collapsed:
- [ ] it needs to be truncated to one line (< 100 chars) with closing ellipsis.
- [ ] there needs to be some better differentiation between displayed objectives.
- [ ] we still need a summary -- if not in the left margin, then above with the title and collapse control.

Alternately we could just show the summary, and collapse completely. The "show more" option is possible, but probably not the right approach. If they need to see the objectives, they need to see the entire set.

Let's review options.

jrjohnson:

saschaben:
Looks great. I think we got this one right. The collapsers now work quite cleanly, I like what you've done.

final cleanup items....
1. missing translation: `general.summary` (FR: 'résumé')
2. please correct my bad translation for  `'objectiveCount': 'other'` to `'Il y a {{count}} objectifs'`

stopfstedt:
click tested and code reviewed. LGTM :+1:

PR #1380, opened 2016-02-04
give more breathing room to pagination controls.

### LINKED ISSUE #1370: clean up session listing controls
here's what we have:

as-is

here' what would look a little better, adding 0.75em padding to the `clickable is-clickable` element:
to-be

stopfstedt:
fixes #1370.


saschaben:
Looks good on UI side, but we should provide a spacer for when there are fewer than the default "per page" number of sessions listed:

good:

goodspace

needs spacer:
needspace

PR #1385, opened 2016-02-05
Fix user profile

jrjohnson:
Don't destroy cohorts list when displaying students 
We were mutating the original list of cohorts and removing the primary cohort.  Thus when a user was saved they got removed from that cohort, which is obviously bad.

Also updated the display of cohort information to be a bit cleaner.


thecoolestguy:
@homu r+

PR #1387, opened 2016-02-08
added recurring events functionality

### LINKED ISSUE #1326: Replace "Recurring Event" option for offerings
This is functionality which was dropped in the conversion due to it being buggy as sin in previous releases; unfortunately it appears to be a critical need for the users.

Recurring events allow for:
1. the creation of multiples of a given offering, 
2. the cloned offerings take identical associations as the original which spawns them
3. events can be set to occur on a given day or days of the week, for a specified number of events
4. or a specified number of weeks
5. Recurring events do not need to be chained or tracked as a set, but if it were possible to allow for that (as very low hanging fruit), I wouldn't cry.

djvoa12:
WIP

@jrjohnson need review

Latest commit notes:
1) Reduces the number of weeks recurring by one (eg. 3 weeks is 3 total weeks)
2) Refactored offering creation (single, single recurring, small group, small group recurring)
3) Added an observer to always add first user picked day on recurring days

Need to verify:
- Creation on all 4 levels, as specified above, works nicely
- Couldn't validate that default instructors and instructorGroups get added properly. Need to validate this.

Issue:
- One of the latest PR merges creates a validation conflict. Need to rethink the approach as validating user input for number of weeks vs. room is different (based on `recurringEvents` property).

Note: Need tests after validating above.

Fixes #1326

### DIFF
--- a/app/components/offering-editor.js
+++ b/app/components/offering-editor.js
@@ -4,7 +4,7 @@ import moment from 'moment';
 import ValidationError from 'ilios/mixins/validation-error';
 import EmberValidations from 'ember-validations';
 
-const { Component, computed, isEmpty, isPresent, ObjectProxy, RSVP } = Ember;
+const { Component, computed, isEmpty, isPresent, ObjectProxy, observer, RSVP } = Ember;
 const { alias, notEmpty } = computed;
 const { all, Promise } = RSVP;
 const { PromiseArray } = DS;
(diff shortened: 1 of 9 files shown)


jrjohnson:
As discussed in the PR that added room validations I'm ok with dropping ember-validators here (and probably everywhere) and just writing the validations as custom properties on the component.

jrjohnson:
Replaced by #1440

PR #1388, opened 2016-02-08
Update to Ember 2!

jrjohnson:
Collection of changes which result in us being on the latest ember-cli and the latest ember release.

Should be tested by several people before we merge to ensure there are no edge regressions or weirdness.

To ease testing this code is now available on https://ilios-stage.ucsf.edu/


dartajax:
I notice that on stage (Ember 2.0) and not on demo that when you are doing a search on session (for example "vas", and then you decide to widen the search so you hit the backspace key thinking your cursor still had the focus in the search box.  What happens is the same as hitting the backspace button on the browser since the focus is no longer in the search box.  This happens on stage but not on demo.  On demo, the focus stays where it should continue to be.

dartajax:
The toggles don't work really well especially when turning them to the off position.

see the screen shots below 


image

now click the Independent Learning toggle ...

It is inconsistent but possible to find yourself in a situation like the one below where clearly it is no longer an ILM because the hours field and due by field are missing but it appears in the yes position.

image

jrjohnson:
@dartajax good catch.  I addressed it in the latest version.  I'm going to pass this off to @stopfstedt now for code review and merging, unless you want more time to play with it on stage.

stopfstedt:
LGTM :+1:

PR #1391, opened 2016-02-08
fixes broken associations in my-report generator.

### LINKED ISSUE #1390: my reports: associating with "anything" and "instructor" broken
1. select any value in "associated with" dropdown, then select "anything" option. an error occurs. the error screen is displayed.
2. select "instructor" as value from the "associated with" dropdown. an silent error occurs. no autocomplete/search box for instructor name is shown.

stopfstedt:
fixes #1390

### DIFF
--- a/app/templates/components/new-myreport.hbs
+++ b/app/templates/components/new-myreport.hbs
@@ -28,7 +28,7 @@
     <p>
       {{t 'dashboard.associatedWith'}}
       <select onchange={{action "changePrepositionalObject" value="target.value"}}>
-        <option value=null selected={{is-equal null currentPrepositionalObject}}>
+        <option value='' selected={{is-equal null currentPrepositionalObject}}>
           {{t 'general.anything'}}
         </option>
         {{#each prepositionalObjectList as |option|}}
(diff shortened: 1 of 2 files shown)


app/templates/components/new-myreport.hbs
```diff
@@ -28,7 +28,7 @@
     <p>
       {{t 'dashboard.associatedWith'}}
       <select onchange={{action "changePrepositionalObject" value="target.value"}}>
-        <option value=null selected={{is-equal null currentPrepositionalObject}}>
+        <option value='' selected={{is-equal null currentPrepositionalObject}}>
```
jrjohnson: I feel weird about this change - mainly because of the comparison to `null` in the selected attribute.  Is this part of the fix?

stopfstedt: `value=null` came through as string value "null". which to me feels even _weirderer_ than evaluating the truthiness of a blank string.

jrjohnson: Ha, ok - yea, thats worse.

PR #1395, opened 2016-02-09
don't show collapsible section titles in display mode and on empty objectives list.

### LINKED ISSUE #1393: Objective Arrow Should Not Be Visible When Adding MeSH to Session Objective
see screen shots 

the arrow is visible indicating maybe a drop-down menu or more information 

see #1392 as well

image

However when it is clicked, the entire process of adding MeSH is aborted and it is confusing as a user to know what just happened.

stopfstedt:
fixes #1393


saschaben:
LGTM :+1:

PR #1396, opened 2016-02-09
Update session first offering when offerings are updated

### LINKED ISSUE #1276: Hard Refresh Should Not Be Needed After Offering Date/Time Change
It has been reported and we have successfully repeated that when you change a date or time of an offering, the date/time of the offering does not refresh to reflect the change in the session list.

jrjohnson:
Computed property needed to track the offering start times or it
wouldn’t update correctly.  Maybe a little bit of overkill here with
the tracking, but I don’t see any harm in being thorough.

Fixes #1276

### DIFF
--- a/app/components/ilios-sessions-list.js
+++ b/app/components/ilios-sessions-list.js
@@ -65,7 +65,7 @@ export default Component.extend({
     this.set('debouncedFilter', this.get('filter'));
   },
 
-  proxiedSessions: computed('sessions.[]', function() {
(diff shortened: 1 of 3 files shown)


jrjohnson:
@dartajax good catch.  I addressed it in the latest version.  I'm going to pass this off to @stopfstedt now for code review and merging, unless you want more time to play with it on stage.

stopfstedt:
LGTM :+1:

PR #1397, opened 2016-02-09
set max length of offering room input to 60 char.

stopfstedt:
refs #1398

### DIFF
--- a/app/components/offering-editor.js
+++ b/app/components/offering-editor.js
@@ -50,7 +50,7 @@ export default Component.extend(EmberValidations, ValidationError, {
 
   validations: {
     room : {
-      length: {maximum: 60, allowBlank: true}
+      length: {maximum: 60, allowBlank: true, messages: { tooLong: "offerings.errors.roomTooLong" }}
     }
   },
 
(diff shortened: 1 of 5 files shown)


stopfstedt:
selection_094

stopfstedt:
@jrjohnson please review and let me know if i'm on the right track.

not very happy right now with the inability to manually interpolate values into flash messages. 
also, the validation-error to flash message assignment feels kinda kludgy. `Ember.keys()` on validation errors object does not recurse into nested objects, therefore i had to alias nested component props to pull them up.

thoughts/ideas?

jrjohnson:
Hmmm.... I don't love this and I'm not sure we are really getting anything from the validators library that we couldn't easily do on our own.  In fact this component actually has a couple of other validations happening that are not using the validators - they just spit out flash messages, which we could the translate more effectively.  I'm wondering if we should just pull out the small implementations we have of this and go with writing custom case specific validations and messages in each case.  That would probably give us a better UX for errors and I think the code would possible by cleaner as well.

As a side note I'm not super happy with the maintenance of ember-validators by dockyard.  They may be too busy to keep it up to date.

saschaben:
FR: `Nom de l'emplacement est trop long (max. 60 caractères)`

stopfstedt:
ok. thanks for the feedback. i hope this is 'good enough' for now. moving on...

jrjohnson:
Yea, I'm ok with that.  @saschaben can you validate the UX and merge - the code looks good.

Issue #1318, opened 2016-01-13
Add Taxonomies to extend topic tags


saschaben:
See Ilios/Ilios#1257.

Extend the current topic manager to provide for additional complex taxonomies of tagging.

saschaben:
DUE BY 4 MARCH 2016

stopfstedt:
regarding: **linking taxonomies across schools.** 

in Ilios2, the scope for selectable topics was set by the _currently set school_ (via the "school switcher") in the user's session. 
That concept of current school has gone away in Ilios3.
We need an alternative means of determining which taxonomies are eligible for assignment to courses/sessions in Ilios3.

Here is what I remember from previous conversations.

1. taxonomies belonging to the course's owning school will _always_ be available.
2. if a cohort from another school has been linked to the given course, all taxonomies from that cohort's owning school will become available for selection as well.

Example:

1. Course A is owned by School of Medicine. All taxonomies (vocabs/terms) owned by School of Medicine are eligible for assignment to this course.
2. Course B is owned by School of Medicine, and has Cohorts from School of Pharmacy and School of Dentistry assigned to it. Therefore, all taxonomies owned by these three schools can be assigned to this course.

@saschaben please confirm/debunk/correct my assumptions stated above. Thank you.

saschaben:
@stopfstedt please see my notes above.

stopfstedt:
thanks for the feedback. exposure of taxonomies to students via calendar events is a different, but related can of worms, good to have some of the access permissions penned down here.

PR #1404, opened 2016-02-12
added DS models and Mirage support for vocabs and terms.

stopfstedt:
refs #1318


jrjohnson:
I'm 100% ok with merging this (after conflict gets resolved), but just a warning -  it may not actually work.  Our fixtures have been a bit broken for a while and I've been waiting on mirage 0.2.0 to get released to move the entire system over to factories and the new ORM.

We WILL need the factories (even if we get rid of the fixtures eventually).

stopfstedt:
@jrjohnson thanks for the review and warning. rebased against master, this should be good to go now.

stopfstedt:
actually, i take that back. still getting some test errors...WIP.

PR #1406, opened 2016-02-12
Try out the new code climate engines

jrjohnson:
New analysis from code climate… hopefully it works and doesn’t make us look terrible.

We Should get some analysis in the PR.... lets see...


jrjohnson:
Never mind.... apparently we have to pay for that :(  oh well - I guess merge it if it seems legit.

stopfstedt:
seems legit.

PR #1410, opened 2016-02-12
Add a new user by form

jrjohnson:
This exposes a form to create a new user from the user list.  Currently hidden behind feature flag so it will only show up in development.

Saving will only work against the latest ilio/ilios#3.6 release


saschaben:
This looks great. A couple issues noted:

- We are loading in new users from this form as `disabled`. Not sure that is the correct default.
- A new user account works fine -- but I am able to log in with a disabled newly created account. This is a bug, although it may just be exposing something pre-existent. A disabled user is one which is not provided __access__ to the system. BUT: as an entity is still visible within the system, and potentially available for management and association to curricular data. We have this a little backward right now.

config/environment.js
```diff
@@ -8,13 +8,13 @@ module.exports = function(environment) {
...
-      'font-src': "'self'",
-      'connect-src': "'self'",
-      'img-src': "'self' data:",
-      'style-src': "'self'",
-      'media-src': "'self'"
+      'default-src': ["'none'"],
+      'script-src':  ["'self'"],
+      'font-src':    ["'self'"],
+      'connect-src': ["'self'"],
+      'img-src':     ["'self', data:"],
```
stopfstedt: just double-checking- is this correct?

jrjohnson: Pushed an updated version with this fixed.

PR #1412, opened 2016-02-17
Deploy a JSON configuration Object

jrjohnson:
Redis is a bust, it is completely insecure and making it secure is a
huge pain.  Instead lets ship the JSON file to S3 and pull it down over
SSL.   Placed all of the index assets into a single bucket with a
prefix that allows it to be versioned and separated by environment.

This is a simplified deployment strategy that makes it possible for the API server to pull the configuration down and build an index file which can be customized.


jrjohnson:
@thecoolestguy I assigned this to you because it is a change in our AWS setup.  The really important thing going on is that a bunch of index:$VERSION.json files are going to get dumped into an AWS bucket.  I have separated these into several environments production/staging/development and then hard coded an API version identified (currently 1.1) this will let us keep the backend up to date with the latest compatible frontend I hope.  It needs a second pair of eyes on the process though to make sure I'm not making any logical errors in here somewhere.

PR #1414, opened 2016-02-18
Manage Pending User Updates

jrjohnson:
Process for managing users who’s accounts were caught in the sync process and could not be updated.

@saschaben - this is a first draft so we have something to work with - review needed.


stopfstedt:
dig.

PR #1427, opened 2016-02-22
Added an option to course printing to see unpublished sessions

### LINKED ISSUE #1419: Provide better course print option
right now, course print is:

 A. available only ON published courses, and only FOR published course information. This is correct for accessing print summary via the calendar event detail page.

We still need to 

 B. provide the ability to print summary from within non-published courses on the course management page. This is for two purposes: 
  - one is to "see what the students see", and provide the SAME print view as item `A` above; 
  - the second is to "see everything I need to review my course", and act as a pre-flight checklist of sorts, which requires the display of both published and unpublished course info. Both need to be printable.

The original wireframes are a good place to start mining for ideas. https://www.dropbox.com/s/fnj9fa19g0ngr64/Ilios%202%20Courses.pdf?dl=0

We should fix the first issue (provide broader access to the print view to correct its absence/allow for view of unpublished data from the course management page only), and review options for the second (pre-flight review).

jrjohnson:
This allows privileged users to print the course with all unpublished
sessions.  The default is still to show only published sessions.

Unpublished courses and sessions cannot be viewed by unprivileged users
even if they want to.

Fixes #1419

### DIFF
--- a/app/components/print-course.js
+++ b/app/components/print-course.js
@@ -8,6 +8,7 @@ const { service } = inject;
 export default Component.extend({
   store: service(),
   course: null,
+  includeUnpublishedSessions: false,
   tagName: 'section',
   classNames: ['printable course'],
   sortTitle: ['title'],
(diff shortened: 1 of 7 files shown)


saschaben:
this seems a good approach for now as far as I can tell / test.

PR #1441, opened 2016-02-25
Fix program year objectives picker

### LINKED ISSUE #1439: cannot add objective to program year
log into ilios. then pick "programs" from the menu. click through to any program year in any program.
then click the add ("+") button in the objectives section.
the application fails and an error is shown.

here the beginning of the error stack trace from current production env (URLs redacted) :

```
There is 1 error
this.attrs.expand is not a function
TypeError: this.attrs.expand is not a function
    at e.default.a.extend.actions.toggleNewObjectiveEditor (assets/ilios-6868f04cedd4dab99d8fec4940bad663.js:2:14111)
```

stopfstedt:
fixes #1439.

### DIFF
--- a/app/components/programyear-details.js
+++ b/app/components/programyear-details.js
@@ -3,4 +3,14 @@ import Ember from 'ember';
 const { Component } = Ember;
 
 export default Component.extend({
+
+  program: null,
+  programYear: null,
+  pyObjectiveDetails: false,
+
+  actions: {
+    togglePyObjectiveDetails(){
+      this.sendAction('togglePyObjectiveDetails');
+    }
+  }
 });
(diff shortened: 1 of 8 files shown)


jrjohnson:
Code looks good - just needs tests that line up with the new collapsed state by setting the URL to force the open.

app/components/programyear-details.js
```diff
@@ -3,4 +3,14 @@ import Ember from 'ember';
 const { Component } = Ember;
 
 export default Component.extend({
+
+  program: null,
+  programYear: null,
+  pyObjectiveDetails: false,
+
+  actions: {
+    togglePyObjectiveDetails(){
```
jrjohnson: This doesn't seem to be getting called anywhere and looks redundant.

saschaben:
well, function wise this makes me happy. so let's scrub code, get tests and merge.

PR #1473, opened 2016-03-07
Code Cleanup: @each

### LINKED ISSUE #1462: Cleanup our use of @each
Lots of warning like this in our tests:
WARNING: Dependent keys containing @each only work one level deep. You cannot use nested forms like todos.@each.owner.name or todos.@each.owner.@each.name. Please create an intermediary computed property.

Looks like some cleanup is in order in several locations.

jrjohnson:
We were using @each incorrectly in several computed properties.  Ember 2.4 was throwing warnings about this even in cases where they were probably working.

Fixes #1462


thecoolestguy:
@homu r+

PR #1495, opened 2016-03-16
generator for learner subgroups

### LINKED ISSUE #690: Auto Generate New Learner Sub Groups
There needs to be the functionality on the Learner Groups page where the user can click Add New (for 1 new sub group) and the name automatically is generated with the following naming convention ...

[Parent Group Name][Maximum Sub Group # + 1]

In other words, in you were creating a sub group inside Anatomy (and there already exist Anatomy 1 and Anatomy 2), your new group would automatically be named Anatomy 3 and this name would appear in the title field for the save event to be performed.

Also, it would be helpful if there were a drop-down menu where the user could specify the quantity of sub-groups to be created automatically.  Maybe we should have 1 be default and the user could change it.  

The names would be generated using the same logic.  If 4 New Sub Groups were to be created - assuming we just created Anatomy 3, the following four groups would be generated.

Anatomy 4
Anatomy 5
Anatomy 6
Anatomy 7

stopfstedt:
fixes #690.


saschaben:
quick clickthru shows this off well. Notes: 
- [ ] we are missing translations
- [ ] the close of the sub-group manager after the creation of new group sets is pretty herky-jerky: we should probably just put in a spinner here so users don't try to move around on a frozen screen, or recode the destruction of the manager element for a more graceful exit....

otherwise, awesome sauce.

saschaben:
A thought on the page rendering stuff: what if we were to put in a collapser for the sub-group display div -- so when you generate your groups, the display is by default collapsed -- wondering if this would either provide better UX, help with whatever goop is going on in the DOM, neither, or both...?

stopfstedt:
- numbering offset and padding are in
- tests are in.

ok, this is ready for review (again).

saschaben:
this is looking really good. I'm wondering, now that we have the numerator plugged, if we should reduce the max number of group we allow to be generated at a time, until we address the bulk operations issues noted earlier (and until we introduce -- if we do -- the random populating of groups / creation of groups by dividing group populations by a specific number), to keep the user experience in control. 12 seems to be the magic number....let's review.

stopfstedt:
fixed padding bug and added progress overlay. 
this is ready for re-review by @saschaben  and code-review by @jrjohnson .

saschaben:
@saschaben says :+1:

PR #1511, opened 2016-03-24
Include courses from next year on dashboard

### LINKED ISSUE #1490: My Courses dashboard doesn't include courses from next year
https://github.com/ilios/frontend/blob/master/app/services/current-user.js#L268 needs to include last year, this year, and next year.  Which Right now would by 2014, 2015, 2016.

jrjohnson:
Fixes #1490


stopfstedt:
:+1:

PR #1532, opened 2016-03-31
Competency collapser and summary in program year and course details

### LINKED ISSUE #1455: add collapser to competency pane
similar to objectives and taxonomies, competencies should be collapsible/collapsed by default.
based on @saschaben's feedback in #1418 

> One thing --- if we can sneak in a collapser for the Competency section on program and course, the user experience would be much improved. Right now, because of the typical length of the competency list displayed, the page jumps significantly away from the focus area of Terms after saving in the manager. Collapsing competencies would reduce the page height of the display block, and in turn reduce the distance of the visual herkyjerky jump.

stopfstedt:
fixes #1455 

course details:
selection_136

program year details:

selection_137

### DIFF
--- a/app/components/collapsed-competencies.js
+++ b/app/components/collapsed-competencies.js
@@ -1,7 +1,52 @@
 import Ember from 'ember';
+import DS from 'ember-data';
 
-const { Component } = Ember;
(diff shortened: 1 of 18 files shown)


saschaben:
Likey

PR #1533, opened 2016-03-31
rm hardwired 1000 records limit when reporting on mesh terms and instructors.

stopfstedt:
fixes #1533.


stopfstedt:
_...RUCTORS_

PR #1540, opened 2016-04-01
Add students with a cohort

### LINKED ISSUE #1402: Expose UI for creating / adding users
From the user list in the admin panel new users should be able to be added to Ilios.  If the backend is using a directory then a search should take place to add user by directory info, otherwise a straight form can be used to create a new user in the system.

Refs #684

jrjohnson:
For both directory and non-directory schools students can now be added as new users with a Primary Cohort.

Since it is now complete this activates add new user feature for general use.

Fixes #1402


jrjohnson:
@saschaben this is ready for review - I just need to update the translations and tests will pass.

saschaben:
looks good overall, but doesn't appear to be correctly pulling cohorts for selection/assignment: the list is verry truncated (I suspect somewhere we have a `limit=20` set somewhere), and does not include most recent cohorts or all programs.

cohort list for selection should include (in an ideal world):
- All available cohorts (ie program years) for all schools beginning with the previous year (so, if it is 2016, the earliest cohort available for any program/school would be "Class of 2015");
- cohorts should have *some* sort order other than `cohort_id` -- alpha would be acceptable, alpha and grouped by school/program even better.
-the add submission needs a spinner and possibly a modal overlay --- it currently hangs a little after clicking the done button, and allows for a double submission (which appears to create one record with a valid authentication table input, and one without).

I'm assuming we will get around to dealing with validation for duplicate username/pw issues, duplicate ID values?

PR #1545, opened 2016-04-04
added universal locator to course overview.

### LINKED ISSUE #1519: Add Unique Course ID to Course Details
In the past, we provided a unique-ish ID to allow for specific courses to be identified both internally and in the UI: 
course manager 2016-03-29 09-23-33

In order to expedite course identification between years and terms, and for tracking ability, we should do the same now. A similar hash pattern and location and presentation. And even better if we make it a searchable field....

stopfstedt:
fixes #1519 

selection_138


saschaben:
:+1:  LGTM

PR #1553, opened 2016-04-07
Update Froala Editor to v2

### LINKED ISSUE #1336: Update ember-froala version
We can now update to froala v2 which is better.  We should.

jrjohnson:
Major update of the WYSIWYG editor which had some API changes that needed to be accounted for.

WIP: The editor now works, but we need to check to ensure it still has the options and behavior we expect.

Fixes #1336

### DIFF
--- a/app/components/big-text.js
+++ b/app/components/big-text.js
@@ -23,7 +23,12 @@ export default Component.extend({
     }
   }),
   cleanText: computed('text', function(){
-    const text = this.get('text') || '';
+    let text = this.get('text') || '';
+
+    if ((typeof text === 'object' && 'toString' in text || typeof text === 'number')){
+      text = text.toString();
+    }
+
     //strip any possible HTML out of the text
     return text.replace(/(<([^>]+)>)/ig,"");
   }),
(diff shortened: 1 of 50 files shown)


saschaben:
Looks good. two things:

1. now that we are correctly handling deletion of content, we need to add validation for objective title ("objective title cannot be blank")
2. code view is awesome, but there doesn't seem to be a way to turn it off once you turn it on.

jrjohnson:

saschaben:
👍

app/components/big-text.js
```diff
@@ -23,7 +23,12 @@ export default Component.extend({
     }
   }),
   cleanText: computed('text', function(){
-    const text = this.get('text') || '';
+    let text = this.get('text') || '';
+
+    if ((typeof text === 'object' && 'toString' in text || typeof text === 'number')){
```
stopfstedt: possible to user `Ember.typeOf()` here instead of rolling a manual type detection? http://emberjs.com/api/#method_typeOf

jrjohnson: Woot!  Man I googled that crap forever yesterday.  I couldn't figure out how to look into the prototype chain and see what kind of Object something was.  
Thanks.

stopfstedt: LGTM otherwise.

jrjohnson:
Updated and ready to merge (hopefully).

PR #1560, opened 2016-04-08
Default calendar filter to type instead of detail

### LINKED ISSUE #1505: Make calendar filter default setting "course/type" instead of "detail"
Right now the filters default to "[topic]/detail" when turned on; they should instead default to "course/type".

jrjohnson:
Fixes #1505


saschaben:
👍

PR #1564, opened 2016-04-09
Disable the password validator when it is not needed

### LINKED ISSUE #1561: Save Event not Processing on Add User From Directory on Stage
https://ilios-stage.ucsf.edu/users?addUser=true&search=loomis

image

image

After clicking the Done button, nothing happens - there is no event associated with the button click.

jrjohnson:
When creating a new user for a shibboleth or ldap install a password is not required.

Fixes #1561


saschaben:
👍

PR #1579, opened 2016-04-12
added secondary cohort picker to user profile form

### LINKED ISSUE #1556: Edit users secondary cohort
On the user profile screen it should be possible to add / remove secondary cohorts for a user.

UI: 
Put an edit pencil next to the secondary cohorts list.
Present the cohort picker from the course page when it is clicked
Exclude the users primary cohort from the list so it can't be removed accidentally or changed at all.

Discussion:
@saschaben what restrictions should be placed on removing a secondary cohort?  Do we:
1) not allow removal if student is in learner groups?
2) remove student from all learner groups when they are removed from a secondary cohort?
3) remove students from secondary cohort and not worry about learner groups at all?

Refs #684

stopfstedt:
fixes #1556

### DIFF
--- a/app/components/detail-cohort-manager.js
+++ b/app/components/detail-cohort-manager.js
@@ -18,45 +18,13 @@ export default Component.extend({
   selectedCohorts: [],
   sortBy: ['title'],
   sortedCohorts: sort('selectedCohorts', 'sortBy'),
-  useableCohorts: computed({
-    get() {
-      let defer = RSVP.defer();
-      this.get('currentUser.model').then(user => {
-        user.get('schools').then(schools => {
-          RSVP.all(schools.mapBy('programs')).then(programsArrays => {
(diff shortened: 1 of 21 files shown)


stopfstedt:
@saschaben this can be click tested already. integration tests are still pending ATM.

jrjohnson:
@saschaben can you click test this?

saschaben:
functionally this is spot on. The presentation viz the layout of the form is a little wonky though: we have some alignment issues (left/right), and I think we should better be using a form-data field wrapped in an if statement to display the value of "unassigned" (until there is an assignment) rather than having it post in the same space below the label as we wind up putting the cohort info....

jrjohnson:
@stopfstedt - Travis Says `not ok 104 Firefox 31.0 - Acceptance: Session - Mesh Terms (itgc): manage mesh`

PR #1580, opened 2016-04-13
Ensure courses can be added when none exist

### LINKED ISSUE #1457: course management page breaks if no academic years can be loaded from backend
this manifests itself on fresh installs of Ilios. kind of a chicken/egg situation - in order to create a course, the system needs to have at least on course already in order to load academic years for selection.

the solution is to provide a default in the frontend in such cases.

jrjohnson:
Academic years are not real things, they come from the `year` on a
course.  When no courses have been created there are no academic years in the system and new courses can not be created.  Instead of relying on
the API for academic year this builds the years based on the two years
before and after the current year.

Fixes #1457

### DIFF
--- a/app/components/new-course.js
+++ b/app/components/new-course.js
@@ -1,8 +1,9 @@
 import Ember from 'ember';
+import moment from 'moment';
 import { validator, buildValidations } from 'ember-cp-validations';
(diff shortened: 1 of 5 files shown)


saschaben:
Looks good as far as I can tell; not sure quite how to test for an empty set of courses, since I am not sure how to load in a clean and empty version of the db locally.

Issue #1506, opened 2016-03-23
Hide the Lower Panes When Multi-Edit in Learner Groups Is Selected, and refactor groups management


dartajax:
Currently when you switch on the Multi-Edit functionality in Learner Group Maintenance, it only activates the top pane to be able to multi-select.  The label should change as well to something like "Sub-Group Assignments" since once you select multi-edit, all of the sub group assignments appear in the top pane for maintenance.

top pane activated containing all of the functionality of the lower panes ...

![image](https://cloud.githubusercontent.com/assets/7493331/13993289/f8e72978-f0dc-11e5-8c21-08eac5211db9.png)

lower panes not needed anymore ...

![image](https://cloud.githubusercontent.com/assets/7493331/13993319/1792416e-f0dd-11e5-8e8a-7b8bb24386fa.png)

dartajax:
This mostly applies in the "Cohort Members NOT assigned to ..." area since there may be an entire cohort down there that needs to be added to groups and / or sub-groups.

saschaben:
Additionally:

- [ ] add breadcrumbing
- [ ] allow sorting of names
- [ ] default to multi-edit
- [ ] add search
- [ ] provide more constrained list for multi-edit "select all"

jrjohnson:
@saschaben / @dartajax  if we default to multi edit in the top box then are the bottom panes redundant?  Should they be changed instead to be lists of the users in those places with links to the groups they are in as opposed to being the place where edits can be made?

jrjohnson:
What does the more constrained select all look like?  I'm assuming just people in the tree excluding those in the cohort?

jrjohnson:
Do we need to be able to sort names in multi-edit mode?  Or is that just in the bottom two boxes?

saschaben:
@jrjohnson Great questions for my vague definitions! Here are a few initial answers, leading to (of course) many more questions:

 > _if we default to multi edit in the top box then are the bottom panes redundant?_ 
 >> I'm not sure about this one; on the one hand, yes, they become redundant, but if they become only lists, will they serve a value purpose? Hence my thought of just collapsing them to clean up the contextual view of the page, and worry about functionality later....?

 > _What does the more constrained select all look like? I'm assuming just people in the tree excluding those in the cohort?_
 >> That is correct.

 > _Does add search mean add filtering to the list of users? Which list (all three?)_
 >> That is correct. And I think yes, on all lists.

[...]

PR #1588, opened 2016-04-14
Learner Group Inprovements

jrjohnson:
Addressed a few simple pieces of #1506 before tackling a larger refactoring.

Added breadcrumbs to the top of the page.

Added a title attribute with email and campusId in it to make it easier
to figure out which user was which when there are duplicate names in a
cohort.

### DIFF
--- a/app/styles/components/_breadcrumbs.scss
+++ b/app/styles/components/_breadcrumbs.scss
@@ -11,6 +11,7 @@
 
   display: inline-block;
   margin-bottom: $base-spacing;
+  margin-top: $base-spacing / 2;
   text-align: left;
 
   span {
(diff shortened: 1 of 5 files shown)


jrjohnson:
@saschaben please take a look.  Since addressing the large set of issues in learner groups is going to take a bit more time I figured we should get these things in for this weeks release.

saschaben:
@jrjohnson good call. I think this looks fine for now -- we can pump up styling with the larger revamp. two quickies though: 
1) If it is possible to quickly plug in a `0.5em margin-left` value for the `.breadcrumb` style div, that would make it a bit more visually balanced until then, but if it mucks up our other placements in the app then it can stay as is.
2) Again, if possible: Let's add a pipe with some spacing as delimiter in the tooltip to help clean that up: `Yakko.Warner@example.edu  |  xxx666872` rather than `Yakko.Warner@example.edu xxx666872`

That should do it for a first step we can release on friday.

jrjohnson:
I can add the margin but it will move the one on vocabularies as well.  Is that OK?

saschaben:
Just validated that the margin shift won't adversely affect display in vocabs. go for it.

PR #1597, opened 2016-04-14
renamed ext id to course id.

### LINKED ISSUE #805: Change External ID label to Course ID (like in current Ilios version)
I noticed while testing Courses and Sessions in the new version of Ilios (in development) that the label "External ID" confused me. I was wondering what's so external about it? It may be a database field name but it seems that to the users, it would make more sense to label it as "Course ID" like we have in Ilios 2.4.x. This is just an easily veto-able suggestion that I wanted to throw out there.

stopfstedt:
fixes #805

### DIFF
--- a/app/locales/es/translations.js
+++ b/app/locales/es/translations.js
@@ -254,7 +254,7 @@ export default {
     'noSessions': 'No hay ningún sesión en este curso',
     'noCohorts': 'No hay ningún clase de la graduación en este curso',
     'educationalYear': 'Año Escolar',
-    'externalId': 'Identificación del curso',
+    'externalId': 'Identificación del Curso',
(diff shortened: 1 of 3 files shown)


app/locales/es/translations.js
```diff
@@ -254,7 +254,7 @@ export default {
     'noSessions': 'No hay ningún sesión en este curso',
     'noCohorts': 'No hay ningún clase de la graduación en este curso',
     'educationalYear': 'Año Escolar',
-    'externalId': 'Identificación externo',
+    'externalId': 'Identificación del curso',
```
dartajax: Can you please capitalize the words 'Curso' and 'Externo'?

stopfstedt: handled.

PR #1599, opened 2016-04-15
Allow new learner groups to be created full

### LINKED ISSUE #1591: allow learner groups to be populated on creation -- or not.
For top level groups, they should allow for creation either as empty groups, OR with selected cohort members.

For sub groups, should allow for creation either as empty groups, OR with selected group tree members.

jrjohnson:
Allows the entire cohort to be placed into a new group instead of just
creating them empty.

Fixes #1591

### DIFF
--- a/app/components/new-learnergroup-multiple.js
+++ b/app/components/new-learnergroup-multiple.js
@@ -30,9 +30,5 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
         }
       });
     },
-
-    cancel() {
-      this.sendAction('cancel');
-    },
   }
 });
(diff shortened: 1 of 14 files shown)


jrjohnson:
@saschaben can you give this a look?

saschaben:
this looks good. we'll still at some point have to address the same functionality at the sub-group level, but that gets into splitting / randomization stuff, and so is really a different ticket for another day.
👍

saschaben:
one thing: can we get the action spinner on the `Done` button when the transaction is fired?

jrjohnson:
Spinner added, tests passing, ready for review.

stopfstedt:
I approve of this.

PR #1623, opened 2016-04-19
pass down and check for mycourses flag before adding delete button in…

### LINKED ISSUE #1621: Delete Button Not Needed In My Courses - Delete Fails Anyway
See the screen shot.

There is no reason anybody should be able to delete a Course to which they are explicitly attached.  

See below.  Plus it bombs out when you try.

image

image

image

stopfstedt:
… courses list.

fixes #1621

### DIFF
--- a/app/components/ilios-course-list.js
+++ b/app/components/ilios-course-list.js
@@ -6,38 +6,47 @@ const { service } = inject;
 const { PromiseObject } = DS;
 
 const CourseProxy = ObjectProxy.extend({
-  userCoursesOnly: false,
+  isUserCourse: false,
   content: null,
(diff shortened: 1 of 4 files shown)


jrjohnson:
I don't think this is the right solution to the problem as it only solves the issue if the user happens to be looking at the course using the myCourses view.  Modifying https://github.com/ilios/frontend/blob/master/app/components/ilios-course-list.js#L12 to return false if the course is in a state where it cannot be removed is a more permanent solution.

stopfstedt:
valid point @jrjohnson. i made the appropriate changes.

stopfstedt:
obsolete.

PR #1647, opened 2016-04-27
Improve Learner Group display

### LINKED ISSUE #1506: Hide the Lower Panes When Multi-Edit in Learner Groups Is Selected, and refactor groups management

### LINKED ISSUE #1697: Learner Group Check All Stays Checked After Operation Has Been Completed
A state such as shown below doesn't really fit our processes.

image

I had used the Check All functionality recently but after using it, shouldn't it be de-selected?

### LINKED ISSUE #1698: Modal Dialog Takes Too Long To Appear in Learner Groups
This only is evident when you move a whole bunch of students into another group or sub-group.

Find a full learner group (or sub group).  In this example there are 166 students included in the move from Full Class Group 04 to Full Class Group 06.

After clicking the button as shown below..
image

I will try and capture the screen post-button-click to demonstrate if possible.

The button has been clicked but the modal dialog box should have popped up to inform the user (me) that the transfer of all these learners is taking place.

image

jrjohnson:
Substantial re-write of the single learner group view.

Fixes #1506
Fixes #1697
Fixes #1698


jrjohnson:
@saschaben can you take a look at what is here now?  Couple of questions for sure:
1. What should be editable here?  Right now nothing is - is that a good idea?  I have no idea.
2. I changed the headers to be group members / cohort people, does that make sense?
3. How do the tables for group members look?  Its a little different from the mock.

Also any general feedback welcome at this point as I'm going to be duplicating a lot of the structure on the edit view.

saschaben:
@jrjohnson here's feedback:

1. I think we go with the no-edits view as is now, and provide a manager control to enter edit mode as we do elsewhere in the system.
2. not sure this display is intuitive enough for our users: maybe return to a text label a la v2:
 
> group members: _n_
> total cohort members: _nx_

or even just "_n_ members of _nx_ total cohort members"

3. The tables look OK: we can probably make them prettier, but for now I think the display is alright. This is the first instance (I think) where we have sorted headers for a table in a scolling box, which is a little odd when the header scrolls out of view, but unless making a static header is bog simple I don't want to spend time trying to get it put in.

also: the style on the group members box has a 0px top border; since the header is not static, it should be matching the other borders (2px) to make an actual box.

all in all this looks GREAT.

jrjohnson:
Notes from group review:

- [ ] default location and instructors should be editable on the summary
- [ ] sub group table should not appear if there are no groups
- [ ] change Members in header to 'Total Members in Cohort: $COUNT'
- [ ] add members from the cohort wth a plus sign to add them one at a time
- [ ] Create an add all button with some kind of confirmation message to ensure it wasn't clicked accidentally


Simplified workflow for managing users
	- When in manage just have two buttons (add to group and remove to cohort)
	- List the users current group which is sortable
	- The current group should be a link which will take the admin to that group when clicked.

jrjohnson:
@saschaben this is ready for you to take another look at.  It is mostly un-styled (buttons are kind of all over the place) and some of the translations need work - but I want to make sure that the interactions are correct before getting into that.

saschaben:
OK. This is close to where we need it. Overall I think the functions and transaction are tight. A few notes to address:

1: in manage mode for a subgroup, the list goes from just the subgroup in read-only to the unsorted jumble of everyone in the tree in manage mode. If there is any way to list the current subgroup members first --- or at worst open the manage mode sorted by group --- that would be easier for users to deal with.

2. Did we decide on no "select all" option checkbox? or is not having it on either the manage users box or the "not in group" box an oversight?

2. the +/- management control is nice, but a little confusing at first. I don't have any issues with leaving it in and getting feedback on any possible changes from users, but we should make sure that we document really well.

3. clicking the "remove to cohort" button on a user in a subgroup works great -- but it fires the modal panel twice (I'm assuming for the two operations of 'move to top of tree' and THEN 'move out of tree'); because the progress bar restarts, this is a little confusing. If it is re-workable, cool. If not, document the hell out of it for users.

4. with the column headers in manage mode, we may want to change the label of "group name" to "(sub)group name", to help distinguish that they are looking at the top level of a tree of subgroups....

5. Finally, it looks like the count numbers at the top right (eg "2/156") are not refreshing properly after actions moving users in and out of subgroups or the tree.

In all mostly very minor usability issues. This is a HUGE improvement!

jrjohnson:
@saschaben take another look!

saschaben:
looking good! The only thing I see is a nit: the top border of the scroll boxes for the tables is not there: |_|

If it's an easy css tweak, cool. otherwise, I think we are good.

saschaben:
👍

stopfstedt:
LGTM :+1:

PR #1650, opened 2016-04-27
Fix bad translation key in title

### LINKED ISSUE #1584: Missing Translation -- Add New User - User Exists
image

The callout on mouse-over states ...

`Missing translation: admin.gotToUser`

It should probably explain that the user already exists in Ilios or something like that, no?

jrjohnson:
Fixes #1584


dartajax:
I think this one is good to merge, no?

PR #1653, opened 2016-04-27
Improved session sorting

### LINKED ISSUE #1444: Session Sort / Filter Options Need to Maintain Their Values / State
To reproduce ...

* Pull up a Course 
* Filter the Session List as shown below 

image

* Review one of the Sessions 

image

* Return to the list of Sessions

image

* The filter has been cleared and everything has been reset to default values.

image

### LINKED ISSUE #1571: First Offering Sort On Sessions Is Not Working On Initial Click
To reproduce ...

https://ilios-stage.ucsf.edu/courses/812

image

Click on First Offering sort ascending button.

Screen refreshes and returns this ...

image

Clearly, the sort order is not actually correct.  You can refer to the squiggly lines for out of order items.

image

Click it again as shown to reverse the sort order to First Offering Descending.

This looks correct.

image

Click the First Offering sort again to revert it to ascending.

This looks correct as well now - even with only the first 25 loaded.

image

Advance to upcoming pages to verify the sort is still correct.

jrjohnson:
Pre-load session details for sorting This slows down the initial load, but it is the only way to get sorting
to work for first offering, group count, and offering count. Fixes #1571

Make session sorting sticky
Used url query parameters to make the choice for session sorting sticky.
Which allows the user to go back and forth to individual sessions and
maintain their place. Fixes #1444

### DIFF
--- a/app/components/ilios-sessions-list.js
+++ b/app/components/ilios-sessions-list.js
@@ -6,7 +6,7 @@ const { Component, computed, inject, observer, RSVP, run, ObjectProxy } = Ember;
 const { PromiseArray, PromiseObject } = DS;
 const { service } = inject;
 const { debounce } = run;
-const { sort, not, alias } = computed;
+const { sort, not, alias, collect } = computed;
 const { Promise } = RSVP;
 
 const SessionProxy = ObjectProxy.extend({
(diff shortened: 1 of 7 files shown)


jrjohnson:
@saschaben please review speed tradeoff for loading busy courses.  Do we need to do something drastic and complicated or is this sufficiently fast?

saschaben:
It definitely is a hit in performance for load time, but I think we can manage with it, for now. We'll have to watch for problems on slow networks though...

stopfstedt:
LGTM

PR #1671, opened 2016-05-05
Modernize user search

jrjohnson:
While working on #1506 I'm running into issues with the user-search component so I took a detour and touched it up.


stopfstedt:
click tested and code reviewed.

:+1:

PR #1721, opened 2016-05-24
Make session filter name a query parameter

### LINKED ISSUE #1675: Filters Need to Maintain State on the Session List
It is nice that the sort order remains but the filtering does not.

Filter for exams and select one to review as shown below...

image

Return to the session list as shown below 

image

Now the screen reverts and loses the filter was applied.

image

This is a good one to fix as far as user experience is concerned.  I have heard about this frequently from users.

jrjohnson:
This results in the filter being sticky while navigating between
sessions.

Fixes #1675

### DIFF
--- a/app/components/ilios-sessions-list.js
+++ b/app/components/ilios-sessions-list.js
@@ -1,11 +1,9 @@
 import Ember from 'ember';
 import DS from 'ember-data';
-import { translationMacro as t } from "ember-i18n";
 
-const { Component, computed, inject, observer, RSVP, run, ObjectProxy } = Ember;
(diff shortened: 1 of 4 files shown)


jrjohnson:
@dartajax please review and make sure this meets muster for #1675

dartajax:
Unless I did something wrong, which is definitely possible, I am not seeing the session filters remaining as search parameters.  The session list resets.

dartajax:
I am passing this on to @stopfstedt for code review.  Functionally, it seems fine to me.

stopfstedt:
click tested and code reviewed. :+1:

PR #1726, opened 2016-05-24
Correctly identify ilmSessions which can be published

### LINKED ISSUE #1658: Published Sessions Appearing In "Sessions Incomplete: cannot publish"
The session shown below is already published.

image

image

image

jrjohnson:
We were depending on a key which doesn't exist.  Depend on
ilmSession.dueDate to ensure that publishable ILMs show up that way.

Fixes #1658


jrjohnson:
@dartajax this is ready to review to ensure you don't see any other weirdness with what is supposed to show up where in publish all.

saschaben:
works for me 👍

PR #1727, opened 2016-05-25
Sortable course list

### LINKED ISSUE #1262: Sort Order Selection Needed on Courses
This is related to #98 -- going to break it out into separate issues.

For Course, should allow sort by:
Course | School | Level | Start Date | End Date | Status

The values selected should be sticky and as such should retain the user's choice during the current session with Ilios.

jrjohnson:
Courses can be sorted by title, level, date, and status.

Fixes #1262

### DIFF
--- a/app/components/ilios-course-list.js
+++ b/app/components/ilios-course-list.js
@@ -10,16 +10,21 @@ const CourseProxy = ObjectProxy.extend({
   content: null,
   currentUser: null,
   showRemoveConfirmation: false,
+  i18n: null,
   status: computed('content.isPublished', 'content.isScheduled', function(){
+    const i18n = this.get('i18n');
     let course = this.get('content');
+    let translation = 'general.';
     if (course.get('isScheduled')) {
-      return 's';
-    }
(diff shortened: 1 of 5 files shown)


app/components/ilios-course-list.js
```diff
@@ -4,11 +4,23 @@ import DS from 'ember-data';
...
+const { collect, sort } = computed;
 
 const CourseProxy = ObjectProxy.extend({
   content: null,
   currentUser: null,
   showRemoveConfirmation: false,
+  status: computed('content.isPublished', 'content.isScheduled', function(){
+    let course = this.get('content');
+    if (course.get('isScheduled')) {
+      return 's';
```
stopfstedt: since this is supposed to be an alphabetical sorting, it would be better if we used the translatable values for 'scheduled', 'published' and 'not published here'.

jrjohnson: Good call!  Will change.

jrjohnson:
Requested changes made

stopfstedt:
i accept this.

PR #1728, opened 2016-05-25
Better course director management

### LINKED ISSUE #545: Manage Directors Redo
Currently managing directors is the only place where we have no save / undo - we need to have that.  Follow the pattern currently in the topic manager.

Needs a mock which indicates how the list is displayed when not in edit mode and where the 'Manage Directors' button goes.

### LINKED ISSUE #1630: Too Easy To Remove Director From Course - Follow Usage Pattern of Other Fields
I am pasting in Manage MeSH just as an example.  It seems it is too easy to accidentally click on the Director label and have the association with that course for the director removed with no confirmation or save event required.

image

jrjohnson:
Placed the course directors inside of a manager like learner groups.

Had to fix the style of the course overview component which necessitated some test cleanup.  On the plus side it is responsive now.

Fixes #1630
Fixes #545

### DIFF
--- a/app/templates/components/course-director-manager.hbs
+++ b/app/templates/components/course-director-manager.hbs
@@ -1,0 +1,20 @@
+<div class='detail-actions'>
+  <button class='bigadd' {{action (perform saveChanges)}}>
+    {{fa-icon (if saveChanges.isRunning 'spinner' 'check') spin=(if saveChanges.isRunning true false)}}
+  </button>
+  <button class='bigcancel' {{action this.attrs.close}}>{{fa-icon 'undo'}}</button>
+</div>
+<div class='detail-content'>
+  <ul class='removable-list tag-list'>
+    {{#each (sort-by 'lastName' directors) as |user|}}
+      <li {{action 'removeDirector' user}}>
+        {{user.fullName}} | {{fa-icon 'remove'}}
+      </li>
+    {{/each}}
+  </ul>
+  {{user-search
+    addUser='addDirector'
+    roles='1'
+    currentlyActiveUsers=directors
+  }}
+</div>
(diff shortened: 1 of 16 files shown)


jrjohnson:
There is a minor test to fix, but @saschaben can you take a look at this and see if it will work for managing the directors.  I'm also not sure where to put the save buttons when in management mode.  Are they Ok there?

saschaben:
I think this works well. The buttons are fine where they are. The display mode is a little hard to read when there are a lot of course directors selected (the whole delimited text list for items like this is troublesome to me), but unless we can slap on a new style (badge without "x" delete control, more non-breaking space between names, something boneheadedly simple like that) without any trouble, let's leave as is and create a followup styling enhancement ticket.

jrjohnson:
Lets merge this and address badges elsewhere.

saschaben:
👍

stopfstedt:
needs to be rebased.

app/templates/components/course-director-manager.hbs
```diff
@@ -0,0 +1,20 @@
...
+  <button class='bigadd' {{action (perform saveChanges)}}>
+    {{fa-icon (if saveChanges.isRunning 'spinner' 'check') spin=(if saveChanges.isRunning true false)}}
+  </button>
+  <button class='bigcancel' {{action this.attrs.close}}>{{fa-icon 'undo'}}</button>
+</div>
+<div class='detail-content'>
+  <ul class='removable-list tag-list'>
+    {{#each (sort-by 'lastName' directors) as |user|}}
+      <li {{action 'removeDirector' user}}>
+        {{user.fullName}} | {{fa-icon 'remove'}}
```
stopfstedt: for consistency reasons, i suggest removing the pipe separator.

jrjohnson:
Rebased and fixed

PR #1732, opened 2016-05-27
Prevent mistake when bulk editing learner groups

### LINKED ISSUE #1724: Make Action Buttons Disabled or Invisible When Users are Using Check Boxes on Learner Groups
See the screenshots below ...

image

4 learners selected ...
But one learner "Aaron Anderson" clicked for removal with the red "-".

image

The 4 selected users are no longer selected and the one learner gets removed.

image

jrjohnson:
Once a user is bulk managing learner groups by choosing one of the checkboxes disabled the ability to move a single learner.

Fixes #1724


dartajax:
I assigned @stopfstedt for code review.  Functional testing passed as far as I am concerned.
:+1:

PR #1741, opened 2016-05-31
added 'detail-block' class to course overview component.

### LINKED ISSUE #1379: Add quick copy functionality for LM file links
once a learning material is saved, we should provide the permalink that is generated so that faculty can plug it in as a link in the CLE, etc.

perma

### LINKED ISSUE #1739: pixel divider line absent in course details "show details"
looks like out update to directors caused us to lose our line. let's put it back.

linea1

for comparison, the "hide details" view:

linea2

stopfstedt:
fixes #1739

before

selection_011

after

selection_009


jrjohnson:
Actually we are now hoping to avoid classes that apply too many places and will eventually get rid of all the detail-* stuff. This should be added to the course overview directly https://github.com/ilios/frontend/blob/master/app/styles/newcomponents/course-overview.scss

saschaben:
👍

jrjohnson:
@stopfstedt a lot of this is a reversion of the progress made in https://github.com/ilios/frontend/pull/1728 instead it should simply add back in the style to https://github.com/ilios/frontend/blob/master/app/styles/newcomponents/course-overview.scss and not re-add detail-block with is a pretty bloated class that I want to be completely rid of in the next 2 months.

stopfstedt:
this will need to be addressed as part of a more substantial refactoring.

PR #1743, opened 2016-05-31
expose additional LM props in search results

### LINKED ISSUE #1572: Learning Materials Need More Detail
We need to expose as much information as possible regarding Learning Materials when being attached to Sessions and Courses.

Here is a sample search with results.

image

Notice that there is not information about when the file was uploaded or saved, who the author of the document is,  In fact, the second and third audio file would appear identical to the user.  We should provide all of the information that we can to make this easier to know which document to select.

stopfstedt:
for consistency reasons, i added the upload date to LM details view as well.

this fixes #1572


saschaben:
This looks great. should be a big aid to the users.

:+1:

PR #1744, opened 2016-06-01
made language keys configurable.

stopfstedt:
i'd like to reuse this component for the Curriculum Inventory UI.
there, the verbiage is slightly different, so i'd like to switch out the language keys.

selection_012


jrjohnson:
Ok - fine by me.

PR #1748, opened 2016-06-03
Improved learner group selection

### LINKED ISSUE #1611: ILM Learner Group Badges do not show cohort data
the badges after group selection in the ILM group manager do not show the top level group/cohort name info. They should.

Modify to use the sub-text format introduced in MeSH and again in the secondary cohort picker to remedy this

jrjohnson:
Selection of learner groups for ILM sessions is improved:
- More descriptive badges for selected groups
- Separate cohorts for learner group selection
- Improved filtering of available groups

Todo:
- [x] Loading indicators for selected and available groups
- [x] Better spacing between multiple cohorts
- [x] Try and fit more available groups onto the screen to scroll less
- [x] Return bottom border
-  [x] Deal with different title button locations from other blocks

- [x] reduce the line height from 1.5 to 1.25 or even 1.15
- [x] non-manage view display, with badges minus "X" for listing when not in manage mode
- [x] Re-build group list display like https://github.com/ilios/frontend/pull/1748#issuecomment-223717039
- [x] Make initial view collapsed
- [x] Add learner count to each group badge
- [x] Highlighting top level groups and removing the X action
- [x] Sort groups with tree included so they are ordered with most top level stuff first

Fixes #1611


jrjohnson:
@saschaben please review learner group management for ILM sessions.

saschaben:
OK. Looks pretty darn good. here are the things we need to sort out.

1. Parent group labels. 99% of the time, a parent with a set of children (eg, "anatomy labs") will be an empty group, with all users enrolled in the various small groups. If we pull over (ie "add) the parent, we wind up with a ghost group, which fouls up reporting, and for offerings fouls up the generation of small groups by adding a ghost offering. In the ILM manager now, we are pulling over parent groups, and allowing them to be the master key for addition and removal of all children. This may cause some mayhem. I should've caught this yesterday. I'm not sure that in practice what we have here is as user friendly as it seemed in theory. At the moment this will be OK for ILMs, but if we want to use the same control in offerings we need to rethink this. Let's discuss.

2. legibility. the badges look great, but the subtext is too often getting cut off. can we reduce the line height from 1.5 to 1.25 or even 1.15?

3. non-manage view. The listing sucks right now. We should be following the form used by the vocabulary display, with badges minus "X" for listing when not in manage mode:

groups

jrjohnson:
Notes from meeting:
- Group badges by top level group @saschaben - can you mock this?  Or take a whiteboard photo of what this would look like.

saschaben:
Here's a sorta idea. I've added the "remove all groups link along with each grid grouping, but we could just as easily use the popup "remove all, or just this one?" option voiced by Jasonh.

non-managed would be identical butfor the "x" on the badges (and well you may ask, what's a 'butfor'?)

I've included a variant for the edge case of including a top level group where the top level has members, but I'm not sure of that one. welcome input.

gridbyname

jrjohnson:
@saschaben please take a look at new design.  I'm not sure about interactions for adding and removing the top level group without removing its children so lets discuss.  I also wasn't sure if I should highlight the top level group as you did - or if that yellow was just for emphasis.

saschaben:
@jrjohnson yes lets discuss. the highlighting was an attempt to differentiate the top level group badge, since it is different in both function and meaning from the others in this presentation. not sure if that is the correct way to go though.

jrjohnson:
@saschaben please review again - probably still needs some test cleanup and translations, but it is ready for inspection.

jrjohnson:
@dartajax / @saschaben this is ready for final review.

saschaben:
👍  this looks great to me.

dartajax:
I like it. 

:+1:

dartajax:
This might work for the selection of Instructors / Instructor Groups as well?

stopfstedt:
LGTM

PR #1786, opened 2016-06-15
Bulk addition of users

### LINKED ISSUE #1557: Bulk addition of users
Allow users to upload a tab separated list of users which are added to Ilios.

The list should be parsed for validity in the browser and then presented to the user as a table to confirm.
User then chooses primary school and students primary cohort (like in new user process)
Once the user confirms the TSV file request should be sent to /application/bulkUserCreation endpoint where it is parsed and returns JSON response with success indications and failure messages.

Show success / failure status to the user and give them an option to download a TSV file which includes the new userId as well as error messages.

Screen should include a sample file of tab separated user values.

Requires ilios/ilios#1375

jrjohnson:
Users can now be added by uploading a CSV file.

Using a nice CSV parser that auto detects delimiters and line endings so we should probably get away with most files.

Created a sample file which can be downloaded.

User input is validated before being submitted.

ToDo:
- [x] Improve styling of form
- [x] Handle students (currently we’re ignore cohorts)
- [x] Disable the save button if there are no valid users
- [x] Handle username / password os else remove them from the form
- [x] Handle uploads for directory schools (probably just a username)
- [x] add max-height to users list to contain it
- [x] don't attempt to created authentication if no username is present in spreadsheet
- [x] handle duplicate usernames

- [x] Validate file extension as csv or txt

Fixes #1557


saschaben:
Cool -- was just looking at the parser you've selected, seems very solid. We should take note and make sure to tell as best we can however his identified issues with csv files created from excel that don't encode properly. The workaround or two that are noted in his FAQ should cover us, but let's not forget it. We are certainly going to have users who generate files with special characters but who aren't locally defaulting to UTF-8 encoding....

jrjohnson:
@saschaben please test again to validate the new process of putting up a saving indicator while creating new accounts.

saschaben:
getting `Ember Data Request POST /api/v1/authentications returned a 500 Payload (Empty Content-Type) [object Object]` when attempting to upload.

jrjohnson:
Looks like you attempted to add a username that was already there.  I'm not really sure how to handle that.  Maybe pre-fetch a list of all usernames and then validate against them?  That could be a pretty big list though.

saschaben:
Got it. success with a new file of 16 records, complete in .... 7 seconds. That's not quite the five records per second we are targeting, but it's better than a poke in the eye.
we need to go over how to provide users with the necessary onscreen feedback for action, failure, success, and error to cover the basic scenarios (like a dupe record or dupe key field). We should provide users with feedback on:

number of records in the bulk file
number of records OK to load
number of records NOT OK to load (and what fields require correction)
number of records accepted on load
number and identification of records failed on load
reason for failed record loads: duplication of record, duplication of key field, malformed field, missing data

jrjohnson:
Depends on https://github.com/ilios/ilios/pull/1462 - can be tested when that is merged.

jrjohnson:
Ready for review.

app/components/bulk-new-users.js
```diff
@@ -0,0 +1,278 @@
...
+      let authentication = false;
+      if (isPresent(userInput.get('username'))) {
+        authentication = store.createRecord('authentication', userInput.getProperties(
+          'username',
+          'password'
+        ));
+        authentication.set('user', user);
+      }
+
+      let rhett =  {user, userInput};
```
stopfstedt: yeah!

jrjohnson: Keeping it alive!

stopfstedt:
code reviewed. LGTM.

PR #1802, opened 2016-06-17
More intuitive initial learner group management

### LINKED ISSUE #1773: Bad auto collapse when managing learning groups
Ilm Session learner group manager collapses on save when used with no groups.
Steps:
1 - Create a new ILM session
2 - Click manage learner groups
3- add some groups and save
4 - instead of returning to the un-collapsed view it will default back to the collapsed view
5 - click the arrow to un-collapse
6 - now we are weirldy in an editing state

### LINKED ISSUE #1803: Specify "none" when no selected learner groups.
When no groups are selected the header is too big.

@saschaben is removing the header the right choice or should be instead add "none" below it?

screen shot 2016-06-17 at 1 31 22 pm

jrjohnson:
When managing an empty ILM the learner groups should initially be
uncollapsible, then when a group is added it should transition to
non-collapsed.

Fixes #1773
Fixes #1803

### DIFF
--- a/app/components/detail-learnergroups.js
+++ b/app/components/detail-learnergroups.js
@@ -34,8 +34,12 @@ export default Component.extend({
     let subject = this.get('subject');
     let learnerGroups = this.get('learnerGroups');
     subject.set('learnerGroups', learnerGroups);
-    yield subject.save();
-    this.get('setIsManaging')(false);
+    try {
+      yield subject.save();
+    } finally {
+      this.get('setIsManaging')(false);
+      this.get('expand')();
+    }
   }),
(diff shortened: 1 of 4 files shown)


jrjohnson:
@saschaben / @dartajax  ready for review.

saschaben:
works for me. BUT: any chance while we're in here that we can suppress the monstrously large "Selected Learner Groups" when the count = 0?

jrjohnson:
Not before 1:30pm so I would rather fix this bug and then do that separately.

saschaben:
👍

stopfstedt:
LGTM :+1:

PR #1828, opened 2016-06-29
Pikaday v2 upgrade

stopfstedt:
refs #1827

### DIFF
--- a/app/templates/components/course-rollover.hbs
+++ b/app/templates/components/course-rollover.hbs
@@ -30,7 +30,14 @@
     {{#liquid-if expandAdvancedOptions class='vertical'}}
       <div class='item'>
         <label>{{t "general.startDate"}}:</label>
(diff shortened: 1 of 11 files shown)


app/templates/components/course-rollover.hbs
```diff
@@ -30,7 +30,14 @@
...
       <div class='item'>
         <label>{{t "general.startDate"}}:</label>
-        {{pikaday-input size=10 value=startDate maxDate=maxDate minDate=minDate format='YYYY-MM-DD'}}
+        {{pikaday-input
+          size=10
+          value=startDate
+          maxDate=maxDate
+          minDate=minDate
+          format='YYYY-MM-DD'
+          onSelection=(action 'selectStartDate')
```
jrjohnson: Alternatively this can be done without creating an action method on the component with 
`onSelection=(action mut startDate)`

Its valid either way.

jrjohnson:
Looks good.  I'm going to wait to merge until post release since we're very close to having that tagged.

PR #1834, opened 2016-06-30
Add title to course rollover

### LINKED ISSUE #1822: Need to provide ability to roll course over into same (current) year
for a course that might be given multiple times in a year, the UI needs to provide the ability to add it to the same year as the parent course.

This can be handled by appending a value to the title, like `(2)` or something similar, or providing the ability to edit the name on rollover.

Note that when a course is rolled over into the same academic year as the parent, it KEEPS all relationships to program year: course objective > program objective links are cloned along with everything else.

jrjohnson:
Course title can be sent with rollover to override the original.  This allows for title changes and also rolling over courses to the same year.

Fixes #1822

### DIFF
--- a/app/components/course-rollover.js
+++ b/app/components/course-rollover.js
@@ -156,9 +156,6 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
   },
 
   actions: {
-    selectStartDate(selectedDate) {
-      this.set('startDate', selectedDate);
-    },
     changeTitle(newTitle){
       this.set('title', newTitle);
       this.get('loadUnavailableYears').perform();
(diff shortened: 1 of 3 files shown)


stopfstedt:
seems to fail on validation.

app/components/course-rollover.js
```diff
@@ -32,9 +51,16 @@ export default Component.extend({
...
   skipOfferings: false,
+  title: null,
 
   save: task(function * (){
     yield timeout(10);
+    this.send('addErrorDisplaysFor', ['title', 'selectedYear']);
+    let {validations} = yield this.validate();
+
+    if (validations.get('isInvalid')) {
+      return;
```
stopfstedt: borks out here.

saschaben:
on testing it doesn't seem to be sending the new title value. it resends the old title value....

PR #1850, opened 2016-07-06
changed height to min-height.

stopfstedt:
minor CSS fix.

observe!

with hardwired line height, bad:

selection_030

with min-height, better:

selection_031


jrjohnson:
💯 🔧

PR #1853, opened 2016-07-07
removed max-date constraint on date picker for rollover start date

stopfstedt:
changing the academic year selection will result in a max date recalculation. however, the date picker does not pick up on that, it will continue to operate on the originally set min/max date ranges.

the same applies to the min date of that range, the recalculated date does not get applied in the date picker. the base-line requirement is that the rollover  date cannot come before the course's original start date, which can be enforced. updated the date picker component to take the course start date as min date instead.

### DIFF
--- a/app/templates/components/course-rollover.hbs
+++ b/app/templates/components/course-rollover.hbs
@@ -51,7 +51,7 @@
         {{pikaday-input
           size=10
           value=startDate
-          minDate=startDate
+          minDate=course.startDate
           format='YYYY-MM-DD'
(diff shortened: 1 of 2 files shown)


app/templates/components/course-rollover.hbs
```diff
@@ -51,8 +51,7 @@
         {{pikaday-input
           size=10
           value=startDate
-          maxDate=maxDate
-          minDate=minDate
+          minDate=startDate
```
jrjohnson: I think this should be `course.startDate`

PR #1855, opened 2016-07-08
Don't show new learner group form when there are no cohorts

### LINKED ISSUE #1846: Don't allow new learner groups to be created when there is no cohort
The learner group list currently allows new learner groups to be created even when no cohort has been selected.  This should be prevented as it raises an ugly error.
no cohort

jrjohnson:
This prevents attempting to create an invalid learner group.

@saschaben do we need to do something here besides hiding the button to make it clear WHY this isn’t possible?

Fixes #1846


stopfstedt:
reviewed. LGTM :+1:

Issue #1311, opened 2016-01-11
Offerings group selector is buggy


saschaben:
When selecting groups for offering assignment:

- the "groups available for selection" list does not visually match the "groups already selected" list.
   - group trees given as name values on the "available" list are the complete (correct) trees, those on the left ("selected") are not, as they are missing their parent tree name values:

![item1](https://cloud.githubusercontent.com/assets/1753431/12245713/1911a234-b85f-11e5-8fe9-21562a2bc1df.png)

- On initial load of the page, the "groups available" list does not exclude the groups already selected and  displayed in the selected list, making for both user confusion and an inability to de-select groups for edit.

![ilios 2016-01-11 12-23-01](https://cloud.githubusercontent.com/assets/1753431/12245622/9145c20e-b85e-11e5-8b7d-6dc25c87383e.png)

....but after a single click in the right side "available" list, the view corrects itself, and editing is possible:

![afterclick](https://cloud.githubusercontent.com/assets/1753431/12245670/d687b926-b85e-11e5-8116-3b9dccb830ba.png)

saschaben:
one option is to replace this pattern with that in the ILM:

![ilm-lg](https://cloud.githubusercontent.com/assets/1753431/14619031/64f7f740-056a-11e6-889b-f7fc115ffdbe.png)

But here as well, we are missing the top level cohort identification info on our badges: this is noted in #1611 and should be addressed along with this ticket.

jrjohnson:
Needs #1775 first to make it easier to cleanup display.

saschaben:
refs also #866

PR #1869, opened 2016-07-15
Rewrite Offering creation and editing

jrjohnson:
Improved user experience by:
- using the same learner group manager as we use for ILM learner groups
- Switch to duration for selecting end date instead of a date
- Validate more consistently with warning instead of flash messages

Improved developer experience by:
- Removing a lot of duplicate code between offering editor and offering manager
- Brining offering-form up to our current thinking on component design
- Adding more integration tests.

WIP:
- [x] Test creating new offering
- [x] Test editing offering
- [x] Test creating recurring offering
- [x] Test creating small group offering
- [x] Test creating recurring small group offering
- [x] Validate new look and experience.
- [x] Needs translations

@saschaben / @dartajax this is now ready to be battle tested creating all kinds of offerings and editing existing ones.  Please make sure I didn't miss any existing functionality and that everything works as you would expect.

@stopfstedt this code should be ready for review.  Functionally compete and tests are added.  Sorry for the giant PR.


saschaben:
So far so good with testing functionally; <strike>have not yet reviewed as member of assigned student group</strike>.
some layout tweaks however would be good if we can do them:
newoffering

saschaben:
just a thought, POV from others would be helpful:

offgroupsmgr

jrjohnson:
@saschaben I made these adjustments with two notes:

- I Agree that the alignment of the selected learner groups was weird with that padding.  I put it back flush against the edge.
- All of our inputs are now 2.5em.  I favor consistency on this, can you take a look at the new user form and see if we should adjust those back down to 2em as well?

saschaben:
V. Cool. 'Last Mile' comments and response:

- 2.5em inputs: I'm fine with consistency...but our search/filter inputs are actually 1.6em in height. either we need to reduce everything to 1.6em (aaggh no no no), or increase the search input fields to match.

- the recurring event "weeks" input needs the same padding between in and its label as did duration:
recurpadding

- what do you think about flipping recurring and location, for when "offering" is selected as opposed to "small group"? Right now, with location on the left, it makes the "recurring" block jump to the right when selected; if we flipped them, then Location would just magically appear....maybe:
reversepositions

- finally along the same lines as item #2 above, should we put Instructor above the group picker, so that it appears in the same real estate as location when flipped on? It already pushes everything down so I'm not sure that is an issue, especially if we fix the bug that keeps search results from closing after selection.
instructors

jrjohnson:
Ready for review.  @saschaben - can you check the boxes you have tested up top?

saschaben:
👍

app/components/offering-form.js
```diff
@@ -0,0 +1,432 @@
...
+    let offerings = [];
+    offerings.push({startDate, endDate, room, learnerGroups, instructorGroups, instructors});
+    if (!makeRecurring) {
+      return offerings;
+    }
+
+    const userPickedDay = moment(startDate).day();
+    //convert strings to numbers use parseFloat because parseInt takes a second
+    //argument and gets thrown off by map sending that argument as the counter
+    const recurringDayInts = recurringDays.map(parseFloat).sort();
```
stopfstedt: gross. but acceptable with the code comment above. thanks!

saschaben:
fixes #1311

PR #1884, opened 2016-07-22
account for domains that have not been persisted to the server.

### LINKED ISSUE #1874: Error on adding new competencies
I think this may be the culprit -- switcheroo of (domain,title) to (title,domain)

https://github.com/ilios/frontend/commit/b746e28b79346b1489e430b0af04f49a38994531#diff-ddb61f6d0e67e223917a7c915be035bfL33

stopfstedt:
fixes #1874

### DIFF
--- a/app/components/school-competencies-manager.js
+++ b/app/components/school-competencies-manager.js
@@ -12,7 +12,15 @@ export default Component.extend({
     }
     let domains = competencies.filterBy('isDomain');
     let objs = domains.uniq().map(domain => {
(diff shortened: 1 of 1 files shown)


saschaben:
Looks good. Numeric count tho is now confusing, as it counts both domains and subdomains. We should have a count list of ONLY domains (top level), or ONLY subdomain (bottom level), but not the combined total of both as the displayed count.

saschaben:
👍

PR #1896, opened 2016-07-27
Copy sessions into another course

### LINKED ISSUE #1509: Functionality Needed For Session Copy Into Same Or Different Course
We need to make it possible to copy a session (with or without the offerings) into (by default) the same course or into a different course of the user's choosing.

It would copy all of the attributes of the original session (MeSH, Vocabulary, etc.).  If copied into the same course, a "(2)" would be appended to the session title.  It would then be up to the user to update all of the details of the cloned session.

jrjohnson:
Copies the session with no offerings, but all other metadata, into
another course.

Fixes #1509

WIP: 
- [x] Translations

@saschaben this is ready for review, especially re the placement of the copy link.

### DIFF
--- a/app/components/session-copy.js
+++ b/app/components/session-copy.js
@@ -122,7 +122,8 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
       filters: {
         year,
         school
-      }
+      },
+      limit: 10000
     });
 
     if (isEmpty(this.get('selectedCourse'))) {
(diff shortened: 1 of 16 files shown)


saschaben:
target course list is not complete: is truncated at 20 course values. should list ALL courses for the selected year, including draft courses.

jrjohnson:
Session objective are not copying correctly and carrying mesh descriptors.  Still need to work on this.

jrjohnson:
@saschaben ready for another review.  Should be copying session objectives correctly now including their mesh terms, but not their linked course objectives.

saschaben:
👍

stopfstedt:
LGTM

PR #1908, opened 2016-08-01
make program objectives without competencies selectable as parents.

### LINKED ISSUE #1905: Cannot add parent objective to course objective if the parent program objective is not mapped to a competency
see also #1696.

A program competency should be available for selection as a parent regardless of whether it has been mapped or not to a higher level parent. 

Mapping from one level to the next higher level cannot be dependent on the mapping of other levels.

stopfstedt:
fixes #1905

### DIFF
--- a/app/components/course-objective-manager.js
+++ b/app/components/course-objective-manager.js
@@ -1,7 +1,8 @@
 import Ember from 'ember';
 import DS from 'ember-data';
 
-const { Component, computed, observer, on } = Ember;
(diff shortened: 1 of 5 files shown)


stopfstedt:
@saschaben please review and let me know if this hits the mark. thanks!

saschaben:
This should work just fine. The "no domain" category works for me.

stopfstedt:
ok, thanks. will complete translations and add tests.

thecoolestguy:
@homu r+

PR #1924, opened 2016-08-05
Curriculum Inventory Reporting UI - iteration 2

stopfstedt:
adds full CRUD for sequence blocks.

### DIFF
--- a/app/components/curriculum-inventory-report-details.js
+++ b/app/components/curriculum-inventory-report-details.js
@@ -1,9 +1,36 @@
 import Ember from 'ember';
 
-const { Component, computed } = Ember;
+const { Component, computed, inject } = Ember;
 const { alias } = computed;
+const { service } = inject;
 
 export default Component.extend({
+  store: service(),
   report: null,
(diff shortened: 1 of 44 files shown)


jrjohnson:
Testing notes:

- [ ] Filters school breaks in medium screen sizes
- [ ] Change finalize message color from delete red to warning yellow
- [ ] If start date is set in a block end date should be validated
- [ ] saving ad edited report blows up because the duration is empty and it shouldn't be
- [ ] is selective label should be in the edit for a sequence block
- [ ] need a way to delete a course selection without deleting the entire block - needs https://github.com/DockYard/ember-one-way-controls/blob/master/docs/one-way-select.md#adding-a-blank-or-prompt-option
- [ ] change optional label in required dropdown to "optional (elective)"
- [ ] move course selection to the top of the block editor
- [ ] move count as one offering checkbox to the leftmost column
- [ ] make default sort on session list be the title
- [ ] as a future enhancement go to on/off toggles for the session count once selector

PR #1938, opened 2016-08-08
wait a full second on user bulk upload tests.

stopfstedt:
feeble attempt to "sit out" some of the [intermittent test failures](https://travis-ci.org/ilios/frontend/builds/150777866#L962) we're seeing.


stopfstedt:
seems like the source of these troubles, the `BackBurner` class, will be removed from Ember-CLI after v2.8.0 http://emberjs.com/deprecations/v2.x/#toc_ember-backburner

which is soon, v.2.8.0-beta2 is already out.

stopfstedt:
Travis green-lit the first run of this PR. re-running it once more.

PR #1940, opened 2016-08-10
Specify prototype extension explicitly

jrjohnson:
We were getting nasty deprecation notices about Ember extending date,
and we haven't used the function extensions in a while so those are both turned off.  We still need Array and String though and likely will for a
long time.


stopfstedt:
I accept that.

PR #1945, opened 2016-08-11
Improved handling of competency domains

### LINKED ISSUE #1696: Cannot link to or select single-level competency
If a competency has no sub-domains (ie if it is only single level, not 2-level), it is inaccessible from the selector for program year competency:
comp1
Similarly, if a competency has subdomains, the top-level domain is unselectable for linking -- even if it is added to a program year stripped of sub-domains:
comp2

comp3

Since the bridges curriculum does not have competency subdomains, in order to map correctly this will need to be addressed.

jrjohnson:
This set of changes allows a program year to be associated with top level competencies (domains).  They can also be associated with program year objectives.

Fixes #1696


jrjohnson:
@saschaben please test.

saschaben:
- [] make sure that only selected competencies/domains load for mapping selection

saschaben:
blocked by #1947. Waiting for merge to test and validate.

saschaben:
Looking good functionally. only issue I see is with the handlling of the objective parent selector: if you click on any domain, the highlighting and checkbox are activated as expected -- but if you click directly on the checkbox, it activates the highlighting, but not the checkbox itself.

This is a visual UX issue only -- functionally the select works regardless.

jrjohnson:
Issues with checkboxes fixed, assigning to @stopfstedt  for code review

saschaben:
We eventually will have to deal with the "challenged" refreshing of available competencies for selection when both managers are open and there is no intervening page refresh, but that is for another day.
👍

jrjohnson:
@saschaben I wasn't able to get that to happen again, thought maybe it was my imagination.  Are you seeing it here?

saschaben:
@jrjohnson it only occurs with objectives that are already hooked up to a parent; if you modify the competency set while the objective manager is also open and then try to change a parent to a newly added competency, this pops up.  It's not a common use pattern and should not block this pr.

stopfstedt:
LGTM

PR #1947, opened 2016-08-11
Remove ui-select and use a standard select element

### LINKED ISSUE #1946: ui-select is gone, we need to remove it
The ui-select addon has been removed by its creator, we have to replace or remove it immediately so NPM will run.

jrjohnson:
The ui-select component is gone we were only using it to pick the start
date for a program year which can be done with a select list just as
well.

Fixes #1946


stopfstedt:
functionality-wise this looks good to me. however, some of the styles are off. 

1. 'add' button placement.
2. 'saved successfully' message after creating new programyear


selection_010

jrjohnson:
I would have bet anything that we had center placed add buttons on other lists.  I would have been 100% wrong I guess.  I can't find a single one.  Maybe they're just like that in my head.... will fix.  Good catch.

jrjohnson:
@stopfstedt I fixed the alignment of the expand button, however the saved section matches the one in courses and instructor groups so I left it as is to deal with later.

saschaben:
clicktests are AOK. when tests pass this looks good to merge. I will add a separate ticket regarding the save message alignment.

jrjohnson:
Tests are filing in some crazy random fashion.  Really annoying, investigating.....

PR #1949, opened 2016-08-12
Handle mesh descriptors correctly in reports

### LINKED ISSUE #1928: My Reports allows NULL payload with MeSH report generation
If a user searches a mesh term but does not actually select it, and then clicks the save button on MyReports, it sends a NULL value as the prepositional object, and generates an illegal report which then will lock the user out from loading their dashboard until the report is cleared:

mesherror

Solution: "SAVE" should be inactive until an actual selection is made from the mesh returns.

jrjohnson:
A default value was being set here and it shouldn't have been, we need
to wait for the user to search first before knowing the mesh term.

Fixes #1928


saschaben:
👍

PR #1961, opened 2016-08-16
When small groups are created use the defaults

### LINKED ISSUE #1435: Default Location and Instructors Not Working Properly
The first thing I am doing here is setting the Learner Group to have a default instructor and default location.

image

Then I went and added this Learner Group to a Session Offering.  There is no problem here.  The correct default Instructor and Location were picked up from the Session Offering.

https://ilios-stage.ucsf.edu/courses/802/sessions/23602

image

So, I deleted the default Instructor and Location and created a new Session Offering with Anatomy Lab 1 only attached and got the following ...

image

Shouldn't it pick up the default Location and Instructors from the Learner Group assignment that I showed at the top of the ticket?  I don't think this is the exact bug that was reported but I wanted to point it out anyway as I try and root out to see if this is the same issue.

jrjohnson:
For small group offerings we use the default instructors, instructor
groups, and locations specified for each learner group when creating the offerings.

Fixes #1435

### DIFF
--- a/app/components/offering-form.js
+++ b/app/components/offering-form.js
@@ -233,7 +233,7 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
 
     return offerings;
   }),
-  makeSmallGroupOfferingObjects(offerings){
+  makeSmallGroupOfferingObjects: task(function * (offerings){
     const smallGroupMode = this.get('smallGroupMode');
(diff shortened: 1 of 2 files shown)


saschaben:
Click test looks good. 👍

stopfstedt:
LGTM

PR #1976, opened 2016-08-17
Correctly identify selectable competencies

### LINKED ISSUE #1975: naked competency still unlinkable
see below: competencies with children remain unlinkable to program objectives:
unselectable

The issue is we are not distinguishing between a competency domain with children where the children are included in the selected competencies for a program:

__`Ex. #1`__

> Medical Knowledge
> > - [ ] Problem-Solving and Diagnosis
> > - [ ] Knowledge for Practice
> > - [ ] Treatment
> > - [ ] Inquiry and Discovery

and a domain with children, but in a program where only the domain parent level is selected:

__`Ex. #2`__

> - [ ] Medical Knowledge

In this case, `Medical Knowledge` in the MD program -- where it is included with all its subdomains as in ex. `#1` -- would be un-selectable while all its children would be selectable; in ex. `#2`, `Medical Knowledge` in the Bridges program is 'naked', and should be selectable as it has no subdomains associated with it _in this program_.

jrjohnson:
It doesn't matter how many children a domain has, what matters is how
many of the children have been attached to the program year we are
working with.  If none have been attached then the domain is linkable.

Fixes #1975

### DIFF
--- a/app/components/objective-manage-competency.js
+++ b/app/components/objective-manage-competency.js
@@ -81,15 +81,18 @@ export default Component.extend({
 
   domainsWithNoChildren: computed('domains.[]', function(){
     return new Promise(resolve => {
-      this.get('domains').then(domains => {
-        filter(domains.toArray(), (competency => {
-          return new Promise(resolve => {
-            competency.get('children').then(children => {
-              resolve(children.length === 0);
+      this.get('competencies').then(competencies => {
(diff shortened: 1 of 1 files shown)


saschaben:
Totally fixes 1975, including the model rocket that got caught in the eucalyptus tree out by UCI and never came down. 👍

PR #1984, opened 2016-08-17
reassigned col spans to make room for action icons.

### LINKED ISSUE #1978: Action Icons on CI Not Arranged Well
They should be arranged in a line, non?

image

stopfstedt:
fixes #1978


saschaben:
please add a `margin-left:7px` to the fa-download icon (or a `margin-right:7px` to the trash icon) to even out the spacing of the icons in line. Thinking the download is a better target since we don't use it elsewhere.

iconpadding

PR #1985, opened 2016-08-17
replaced eyeball icon from edit links in finalized reports.

stopfstedt:
as discussed, use the edit icon everywhere - even if the linked-to entity cannot be edited.


saschaben:
LGTM 👍

PR #1996, opened 2016-08-18
Add ember exam

jrjohnson:
Ember exam makes it possible to split tests and run them in parallel.  This has two advantages:
1) Each travis build is faster (taking total time from 14minutes to 8 minutes)
2) When there are errors it is easier to match criteria and run a smaller local set of tests to fix them.

It turns out there were some problems here that had to be worked out:
1 - testgroups just got in the way, they are removed, the bulk of the file changes are just removing those.
2 - We had some tests that would sporadically fail when run in isolation this way.  I fixed those and that is the rest of the file changes. The next step is probably to add randomization to try and identify these ordering issues.

I attempted to keep commits separate so hopefully it is easy to follow the changes I had to make to get this working.


stopfstedt:
:+1:

PR #2037, opened 2016-08-25
Don't magically assign students to cohorts

### LINKED ISSUE #2010: Admin Console Cohort Selected Even Though Not Visible In Drop-Down
Select some Learners.  Notice that no Cohort has been selected.  Click Save.

image

The Save event appears to have occurred correctly.  In fact it did ... see below.

It did actually put the Learners into the Primary_Cohort of 14 - `Class of 2011 Masters Entry Program in Nursing`

That Cohort primary_cohort_id of 14 does exist but it has never been published so the save event is being performed on a cohort that does not appear in the drop-down.  It should return an error that no cohort has been selected.

Could it be that the drop-down contains a value but nothing is being displayed because the Program Year has not been published?  That is the case with this one.  

See below -- the fake student `Beverly Andrews` will be used in this example.

Currently she has no primary cohort.

image

Go to the admn console.

image

Select her record and click save with the apparently empty drop-down for cohort. 
image

The save event processes.  Pull her up in the Admin Console.  There she is with a Cohort that could never have been selected since 

a.  the drop-down appears to be empty
b.  the program year has never been published (which is why it didn't "appear" in the drop-down.

image

jrjohnson:
When no eligible cohorts are found we should NOT just choose the last
cohort in the school and dump students into it.  In fact we should do
nothing.

Fixes #2010

### DIFF
--- a/app/components/assign-students.js
+++ b/app/components/assign-students.js
@@ -27,24 +27,20 @@ export default Component.extend({
   savedUserIds: [],
   isSaving: false,
 
-  bestSelectedCohort: computed('school.cohorts.[]', 'primaryCohortId', {
-    get(){
-      return new Promise(resolve => {
-        const primaryCohortId = this.get('primaryCohortId');
-        let school = this.get('school');
-        school.get('cohorts').then(cohorts => {
-          if (primaryCohortId) {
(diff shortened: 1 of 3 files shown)


saschaben:
LGTM

dartajax:
It looks good to me as well.

PR #2048, opened 2016-09-01
Fix some Jumpy Cursor Issues

### LINKED ISSUE #2045: Jumpy Search Cursor
after the upstream fix to the search component, we still have jumpy cursor issue in on place in the system

jumpy_cursor

jrjohnson:
Fixes #2045

### DIFF
--- a/app/components/offering-form.js
+++ b/app/components/offering-form.js
@@ -45,7 +45,7 @@ const Validations = buildValidations({
       allowString: true,
       integer: true,
       gte: 0,
-      lte: 60
+      lte: 59
     })
   ],
   learnerGroups: {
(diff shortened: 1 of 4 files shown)


jrjohnson:
@dartajax please review.

dartajax:
In the screen shot below 

image

I have my cursor after the zero (minutes).  I would like to "delete" this zero and type in "15".  When I hit the backspace key on my keyboard, nothing happens.  The same is true if I use the arrow key (left arrow) to put the cursor in front of (before) the zero character.  When I now try to delete the zero by using the delete key on my keyboard, same as with backspace, nothing happens.  I am aware that if I decide to type in "15", regardless of where my cursor is, that it will accept my entry of "15".  It is a bit confusing though as a user.

dartajax:
Other than what I pointed out above, I am not noticing any jumpy behavior.

jrjohnson:
Good feedback @dartajax this is going to require some additional refactoring of this form.  Can you create that as a separate ticket.

jrjohnson:
I have reviewed my own code here and it LGTM 👍

PR #2056, opened 2016-09-06
Fix school permissions issues for users in different schools

### LINKED ISSUE #2055: school permissions in profile list incorrectly
When an admin user from school A goes to manage a user from school B in the profile page, the list of schools for permissions gets munged:

1) instead of showing a list of all schools other than the profiled user's primary school, it shows a list of all schools other than the current admin user's primary school.

schools1

2) in manage mode, it displays a doubled, incorrect listing for the primary school of the current admin user, and (now) no listing of the profiled user's primary school.

schools2

jrjohnson:
When the authenticated user was not in the same primary school as the
user they were modifying permissions got weird.  This changes the tests
to check for this condition and also modified the CPs to grab the
correct lists of schools.

Fixes #2055

### DIFF
--- a/app/components/user-profile-schools.js
+++ b/app/components/user-profile-schools.js
@@ -91,14 +91,9 @@ export default Component.extend({
     return new Promise(resolve => {
       currentUser.get('model').then(authUser => {
         authUser.get('schools').then(schools => {
-          authUser.get('school').then(primarySchool => {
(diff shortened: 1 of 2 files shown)


saschaben:
Clicktest passes.

thecoolestguy:
@homu r+

PR #2063, opened 2016-09-07
Fix cohort assignment

### LINKED ISSUE #2057: Admin Console Students Requiring Assignment Is Not Working Correctly
There are two different but possibly related issues here.

# Issue One #

image

After clicking Save as shown above, it appears the student has been added to the Cohort correctly.

image

However, refresh the page, and ...

The student is back, indicating that the save event was not processed.  The same can be verified by looking at the user record and noticing that there is no primary_cohort_id value assigned.  Also, there is no entry in the user_x_cohort table, which should also have been processed by the Save event.

image

# Issue Two #

Click the drop-down as shown below.  Try to select a different value (Class of 2019 for example)

image

It never actually loads those values up correctly.  Endless load icon received.

image

jrjohnson:
Fixes two different issues which were preventing cohort assignment from working.

Fixes #2057

### DIFF
--- a/app/components/assign-students.js
+++ b/app/components/assign-students.js
@@ -77,7 +77,8 @@ export default Component.extend({
     for(let i = 0; i < cohorts.length; i++){
       let cohort = cohorts[i];
       let obj = {
-        id: cohort.get('id')
+        id: cohort.get('id'),
+        model: cohort
       };
       let programYear = yield cohort.get('programYear');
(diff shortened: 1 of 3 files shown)


jrjohnson:
@dartajax please test and assign to @thecoolestguy for code review if it passes your click testing.

dartajax:
:+1:

thecoolestguy:
@homu r+

Issue #1781, opened 2016-06-13
Window Sizing Causes Flyout Menu to Overlay App


dartajax:
It is possible to resize the window so that the flyout menu is activated but covers up portions of the application.  See below.

![image](https://cloud.githubusercontent.com/assets/7493331/16026105/bee14cb0-3180-11e6-8e1c-75a8de224581.png)

dartajax:
This window size ...

![image](https://cloud.githubusercontent.com/assets/7493331/16026166/3b0fa49e-3181-11e6-8588-0b5a69214b3d.png)

saschaben:
The current flyout becomes static "out" at about 1470px width; it should not do so until about 1570px width.

PR #2064, opened 2016-09-08
correct menu sizing and transforms

saschaben:
update slide out menu to avoid overlay when in interstitial sizing


saschaben:
@thecoolestguy please review my css mods here. I think it should be solid.

saschaben:
BTW, I believe this is not going to function properly in Safari 9.1, and will need a user test in IE.

saschaben:
fixes #1781

dartajax:
Seems to work fine in IE 11.

PR #2081, opened 2016-09-13
When new courses are deleted remove them correctly

### LINKED ISSUE #1763: Frontend Rendering of Course Delete Fails for Newly Added Course
This is also probably true in other areas of the application.  I will put the results of other places this fail occurs in upcoming ticket updates.  

NOTE:  This does NOT happen in Sessions.  Maybe that behavior could be a good model for areas that do have failure.

To reproduce ...

image

image

image

image

At first nothing seems to happen.

Click the Yes button again..

This error occurs.

image

Hit Refresh (F5) and notice that the Course has indeed been deleted.

jrjohnson:
We were not clearing deleted courses out of the newCourses array, this
led to them still being displayed even when they had been deleted.

Fixes #1763

### DIFF
--- a/app/controllers/courses.js
+++ b/app/controllers/courses.js
@@ -173,7 +173,11 @@ export default Controller.extend({
   actions: {
     removeCourse: function(course){
       course.deleteRecord();
-      course.save();
+      course.save().then(()=>{
+        if (this.get('newCourses').includes(course)) {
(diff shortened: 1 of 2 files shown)


dartajax:
LGTM :+1:

thecoolestguy:
@homu r+

PR #2082, opened 2016-09-13
Clone objectives when creating a new program year

### LINKED ISSUE #2074: Program Year Objectives are not properly cloned on rollover
Currently, program year objectives are not being cloned on rollover. Instead, a new link record in program_year_x_objective is being created to the pre-existing objective. This is entirely incorrect, since if an objective then is modified in a given year, all past objectives will modify with it.

BAD!

On rollover, program year objectives should be CLONED. A new objective record should be created, as well as a new record in program_year_x_objective, along with any competency linkage and/or mesh linkage that exists.

We also need to verify that this is not occurring anywhere else.

OBJECTIVES MUST BE UNIQUE IN EACH USE. We must approach this like we do learning materials. there can be a mother record, but each instance of use MUST be unique, modifiable, and discrete.

jrjohnson:
We were incorrectly linking to the existing objectives, we needed to
create new ones and attach them to the new program year.  Otherwise we will have problems with the audit trail when objective titles are changed in a new program year.

Fixes #2074

### DIFF
--- a/app/components/new-programyear.js
+++ b/app/components/new-programyear.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import { task } from 'ember-concurrency';
 
 const { Component, isEmpty } = Ember;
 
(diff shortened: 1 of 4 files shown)


saschaben:
Is there any way to guarantee (or even just predict) the order in which the cloned objectives will be processed? since it is random now, it generates a list which (ordered by id) is in a completely different order than the parent, and hard to validate/parse.

Many programs and courses want to list the objectives in a specific order (we have another ticket for allowing this; we should discuss options to control listing behavior here and elsewhere where we clone.

saschaben:
👍

thecoolestguy:
@homu r+

PR #2101, opened 2016-09-16
Handle users with no authentication record correctly

### LINKED ISSUE #2086: Admin Console Blows Up When Retrieving Accounts with no authentication record
This may be because the records are missing the corresponding entry in the authentication table.

https://ilios3-demo.ucsf.edu/users/1374

This user returns the following error ..

image

I don't have access to the backend to verify whether this particular user has a record in the authentication table or not.  However, with the time I returned this error in production, the authentication table record was indeed missing.

jrjohnson:
We were not doing a good job of dealing with this situation.

Fixes #2086

WIP:
- [x] Needs translations

### DIFF
--- a/app/components/user-profile-bio.js
+++ b/app/components/user-profile-bio.js
@@ -50,9 +50,13 @@ const Validations = buildValidations({
     }),
   ],
   username: {
+    descriptionKey: 'general.username',
     validators: [
       validator('length', {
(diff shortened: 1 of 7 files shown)


dartajax:
LGTM :+1:

saschaben:
👍

PR #2112, opened 2016-09-18
Use isoWeekday for rolling over course

### LINKED ISSUE #1871: Default date for rollover not correct on Sunday
`Integration | Component | course rollover: rollover course with new start date`
`Integration | Component | course rollover: rollover course prohibit non-matching day-of-week date selection`

Are failing on Sunday Morning - this failure appears to be legitimate.  If I attempt to rollover a course that is already created into the current year the rollover start date is calculated incorrectly.

dates

jrjohnson:
Set the and week both using the iso standard to avoid conflicts between
the standards.  Specifically Sunday is 7 in the ISO standard and 0 in
javascript.

Fixes #1871


saschaben:
LGTM

PR #2120, opened 2016-09-20
Deal with empty HTML correctly

### LINKED ISSUE #1438: Session Description Allows Saving of Blank Space and Empty Strings
There is some strange behavior with editing session description.

Opening this one in favor of #841 which I will be closing.

If a session description has been saved, it is not possible to delete everything and save that.  See below.

Session has description -- Click on it to edit

image

image

Delete the session description and click Done.

image

What happens is ...

Old title still there

image

The bogus workaround -- put a blank space in there using the keyboard space bar.

image

click done and 

..

image

It's gone but at least with 3.x, it is possible to click to edit even though the blank space is there.  We should allow the save to revert the value of the field back to NULL.  This should have happened above where the description was deleted and the Done button was clicked.

### LINKED ISSUE #2072: Objective titles are being turned into empty 0-length strings
somewhere in the process of either conversion from v2 to v3, or with the implementation of Froala, or in the rollover process, or a combination of those issues, objective titles are being allowed which are empty strings.

This requires a data remediation, and determination of proximate cause.

jrjohnson:
In Session Description, Objective title, learning material description, and anywhere else HTML is being saved we need to account for a blank set of input.  If a user just enters a space or some new lines this is technically blank, but the display result is the same.

Fixes #2072
Fixes #1438

### DIFF
--- a/app/components/editable-field.js
+++ b/app/components/editable-field.js
@@ -7,7 +7,7 @@ export default Component.extend({
   value: null,
   isEditing: false,
   isSaving: false,
-  isSaveDisabled: true,
+  isSaveDisabled: false,
   renderHtml: false,
   classNames: ['editinplace'],
   clickPrompt: null,
--- a/app/components/editable-field.js
+++ b/app/components/editable-field.js
@@ -1,15 +1,24 @@
 import Ember from 'ember';
 import { timeout, task } from 'ember-concurrency';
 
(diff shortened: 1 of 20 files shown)


saschaben:
LGTM 👍

thecoolestguy:
@homu r+

PR #2127, opened 2016-09-20
Copy session objectives in the order they were created

### LINKED ISSUE #2089: preserve objective order when copying sessions
When a session is copied the objectives have to be saved in order so their original sort order is preserved.

jrjohnson:
Since we don't have explicit ordering of objectives the creation date is
the implicit sort order.  This is based on ID and should be preserved
when copying a session.

Fixes #2089

### DIFF
--- a/app/components/session-copy.js
+++ b/app/components/session-copy.js
@@ -86,10 +86,12 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
 
     // save the session first to fill out relationships with the session id
     yield session.save();
+    yield all(toSave.invoke('save'));
 
     //parse objectives last because it is a many2many relationship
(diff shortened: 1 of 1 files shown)


saschaben:
Looks good so far. Am running into some issues with learning materials failing and returning a 0 payload, but I think that is due to the setup of LMs on Demo.

saschaben:
_FAIL_: session copy is not correctly posting to the selected target courses when the target course is from another academic year than that of the parent. Instead it only will post to the first listed course from the list of courses available in the parent course academic year.

jrjohnson:
Ok - new ticket will be addressed separately.  Let's go ahead and review / merge this.

thecoolestguy:
@homu r+

PR #2129, opened 2016-09-21
Use tasks to prevent multiple user additions

### LINKED ISSUE #2059: New User Creation allows for Multiple Entries
when adding a new user from the directory or from the UI form, The interface allows for repeated submit actions, causing multiple duplicate records to be generated. Only the first record receives an authentication record, leaving the other dupes unusable and cluttering the system.

The input action should be constrained to allow ONLY ONE TRANSACTION per request.

jrjohnson:
Before this change it was possible to double or quintuple click the add
user button and create a bunch of bad accounts.  Using a dropable task
this is no longer possible.  Only the first click will be counted.

Fixes #2059

### DIFF
--- a/app/components/new-directory-user.js
+++ b/app/components/new-directory-user.js
@@ -1,5 +1,6 @@
 import Ember from 'ember';
 import { validator, buildValidations } from 'ember-cp-validations';
+import { task } from 'ember-concurrency';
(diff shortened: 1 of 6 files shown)


saschaben:
LGTM 👍

thecoolestguy:
@homu r+

PR #2136, opened 2016-09-22
Reset the selected course when changing years

### LINKED ISSUE #2130: Session copy fails to point to correct target
session copy is not correctly posting to the selected target courses when the target course is:
  <strike> a) from another academic year than that of the parent, and</strike>
   b) when only one course is available in the select list, or 
   c) when the desired course is the first listed in the select dropdown.
Instead it will post to the first listed course from the list of courses available in the parent course academic year.

See also #2127

jrjohnson:
If we don't reset the course then loading the new years doesn't change
the default course selected.  If a user then clicks done without
changing the course selection an invalid course is chosen.

Fixes #2130

### DIFF
--- a/app/components/session-copy.js
+++ b/app/components/session-copy.js
@@ -152,6 +152,7 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
 
   actions: {
     changeSelectedYear(newYear){
+      this.set('selectedCourse', null);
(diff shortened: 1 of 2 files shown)


saschaben:
👍

thecoolestguy:
@homu r+

PR #2194, opened 2016-10-04
rework of learning materials search

### LINKED ISSUE #1872: What is up with the Learning Material Search Input?
It doesn't match our standard input fields; it doesn't match our non-matching search inputs.

Should at the very least match other fields in height.

stopfstedt:
fixes #1872 

this is much more in line with our other search dialogs now, see for example MeSH search.

### DIFF
--- a/app/components/learningmaterial-search.js
+++ b/app/components/learningmaterial-search.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import { task } from 'ember-concurrency';
 
 const { Component, inject } = Ember;
 const { service } = inject;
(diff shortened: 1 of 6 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 626c70596c11f99fcf8c1bfa4294ddb95162d63e

http://deploy-preview-2194.ilios-frontend.netlify.com

saschaben:
@stopfstedt is it possible to have new LMs that are selected get added to the top of the LM list? In other words, reverse the default display order from oldest > newest to newest > oldest? It will be easier for users to see that the things they have selected are added that way.

saschaben:
user functional clicktest passes muster. 👍

app/components/learningmaterial-search.js
```diff
@@ -1,99 +1,101 @@
...
-        showClearButton: false,
-      });
-    },
-
-    inputValueChanged() {
-      let input = this.$('input')[0];
-      let searchTerms = input.value;
-
-      if (this.get('currentlySearchingForTerm') === searchTerms) {
+    search: function(query){
```
jrjohnson: This should be converted to a task to mitigate annoyances that happen when a promise returns after a component is removed from the dom.

app/components/learningmaterial-search.js
```diff
@@ -1,99 +1,101 @@
...
+      var self = this;
+      if (Ember.$.trim(query) === '') {
+        self.set('searchReturned', false);
```
jrjohnson: Any reason not to use `this` here?

app/components/learningmaterial-search.js
```diff
@@ -1,99 +1,101 @@
...
-      }
-
+      this.set('searchReturned', false);
       this.set('searching', true);
-
-      debounce(this, function() {
-        this.send('search', searchTerms);
-      }, 300);
+      this.set('query', query);
+      var currentMaterials = this.get('currentMaterials');
```
jrjohnson: prefer `let` over `var`

jrjohnson: In this case it actually looks like `const` might be the right choice though.

app/components/learningmaterial-search.js
```diff
@@ -1,99 +1,101 @@
...
+      this.get('store').query('learningMaterial', {
+        q: query,
+        limit: this.get('searchResultsPerPage') + 1,
+        'order_by[title]': 'ASC',
+      }).then(function(lms){
+        let results = lms.map(function(lm){
+          return ProxiedMaterials.create({
```
jrjohnson: Instead of doing this as a proxy can we just use a helper in the template?
Like: `{{contains material currentMaterials}}`

app/components/learningmaterial-search.js
```diff
@@ -1,99 +1,101 @@
...
-      debounce(this, function() {
-        this.send('search', searchTerms);
-      }, 300);
+      this.set('query', query);
+      var currentMaterials = this.get('currentMaterials');
+      this.get('store').query('learningMaterial', {
+        q: query,
+        limit: this.get('searchResultsPerPage') + 1,
+        'order_by[title]': 'ASC',
+      }).then(function(lms){
```
jrjohnson: This should be done as an arrow function:
`}).then(lms => {`

app/components/learningmaterial-search.js
```diff
@@ -1,99 +1,101 @@
...
-
-    search(searchTerms) {
-      const store = this.get('store');
-
-      this.set('searching', true);
-      store.query('learningMaterial', { q: searchTerms }).then((learningMaterials) => {
-        let results = learningMaterials.filter((learningMaterial) => {
-          return !this.get('currentMaterials').contains(learningMaterial);
+    searchMore: function() {
+      var self = this;
```
jrjohnson: use arrow functions instead of `self`

app/components/learningmaterial-search.js
```diff
@@ -1,99 +1,101 @@
...
     },
-
-    search(searchTerms) {
-      const store = this.get('store');
-
-      this.set('searching', true);
-      store.query('learningMaterial', { q: searchTerms }).then((learningMaterials) => {
-        let results = learningMaterials.filter((learningMaterial) => {
-          return !this.get('currentMaterials').contains(learningMaterial);
+    searchMore: function() {
```
jrjohnson: should be written as `searchMore()`

app/components/learningmaterial-search.js
```diff
@@ -1,99 +1,101 @@
...
+      var self = this;
+      var currentMaterials = this.get('currentMaterials');
```
jrjohnson: Prefer `const` over `var` here

app/components/learningmaterial-search.js
```diff
@@ -1,99 +1,101 @@
...
+      var query = this.get('query');
+      this.get('store').query('learningMaterial', {
+        q: query,
+        limit: this.get('searchResultsPerPage') + 1,
+        offset: this.get('searchPage') * this.get('searchResultsPerPage'),
+        'order_by[title]': 'ASC',
+      }).then(function(lms){
```
jrjohnson: Arrow function

app/components/learningmaterial-search.js
```diff
@@ -1,99 +1,101 @@
...
+        let results = lms.map(function(lm){
+          return ProxiedMaterials.create({
```
jrjohnson: Use helper in template instead of proxy for this.

app/components/learningmaterial-search.js
```diff
@@ -1,99 +1,101 @@
...
+      this.set('searchReturned', false);
+      this.set('searching', false);
+      this.set('searchPage', 0);
+      this.set('hasMoreSearchResults', false);
+      this.set('query', '');
+    },
 
-    add(lm) {
-      this.sendAction('add', lm);
+    add(proxy) {
```
jrjohnson: Add should be a task which yields the result of sendAction('add') and queues up repeated clicks to ensure we don't try and add the same thing twice.

saschaben:
👍

PR #2201, opened 2016-10-06
Handle multiple users moving around in learner groups better

### LINKED ISSUE #2172: Multiple clicks to remove a single user from a group do not queue correctly
If you open the manager (for example https://ilios3-demo.ucsf.edu/learnergroups/6311?edit=true)
and start clicking minus signs they will all start saving.  However as each one saves it tramples on the work of the save before it.  So each save may add or remove different users.  Instead multiple clicks should be queued up and run non-concurrently.

jrjohnson:
We were killing the spinners to often and updating the saved data out of
order.  Instead we should finish each transaction in queue correctly and
keep the list of users being managed open while the component exists.

Fixes #2172

### DIFF
--- a/app/components/learnergroup-cohort-user-manager.js
+++ b/app/components/learnergroup-cohort-user-manager.js
@@ -4,7 +4,7 @@ import { task, timeout } from 'ember-concurrency';
 const { Component, computed, isEmpty } = Ember;
 
 export default Component.extend({
-  didReceiveAttrs(){
+  init(){
     this._super(...arguments);
     this.set('usersBeingMoved', []);
     this.set('selectedUsers', []);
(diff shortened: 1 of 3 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 68e30e62ecc72b6e985fd51d3982893d2cb8a9ae

http://deploy-preview-2201.ilios-frontend.netlify.com

saschaben:
@jrjohnson spinners appear to be turning off before moves are complete when moving multiple users --- it's far more noticeable when adding from the user list `Cohort Members NOT assigned ` than when removing them from the active group.

jrjohnson:
@saschaben please test again.  Should be working for both boxes now.

saschaben:
huzzah 👍

thecoolestguy:
@homu r+

PR #2203, opened 2016-10-07
lock curriculum inventory download button while d/l is in process

stopfstedt:
requires PR ilios/ilios#1613.

a spinner is shown on the d/l button once the request has started, but no response has been received yet.
once the actual file download has started, the d/l button gets unlocked and the spinner disappears.

this works, _i think_. this can't be tested successfully whilst proxying to the backend b/c of cross domain cookie access restrictions.

### DIFF
--- a/app/components/curriculum-inventory-report-header.js
+++ b/app/components/curriculum-inventory-report-header.js
@@ -23,13 +23,21 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
     this._super(...arguments);
     this.set('reportName', this.get('report.name'));
     this.set('isFinalized', this.get('report.isFinalized'));
+
+    // check if the d/l link points at the same domain as the current page is on.
+    let parser = document.createElement('a');
+    parser.href = this.get('report.absoluteFileUri');
+    let backendDomain = parser.hostname;
+    let frontendDomain = window.location.hostname;
+    this.set('downloadFromSameDomain', (backendDomain === frontendDomain));
   },
   classNames: ['curriculum-inventory-report-header'],
   report: null,
   reportName: null,
   publishTarget: alias('report'),
   isFinalized: false,
-  isDownloading:false,
+  isDownloading: false,
+  downloadFromSameDomain: false,
 
   downloadReport: task(function * (report){
     let anchor = document.createElement('a');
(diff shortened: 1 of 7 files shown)


jrjohnson:
Deploy preview ready!

Built with commit ed7f9949f9032955a254984c12e70395750eed5b

http://deploy-preview-2203.ilios-frontend.netlify.com

jrjohnson:
There is no guarantee that the frontend and API will be on the same domain. Does that make a difference for this solution?

stopfstedt:
@jrjohnson yes this makes a difference. i suppose i can disable the download status checker code if the download link URL and the frontend are not on the same domain.

stopfstedt:
ok, i put a check in to ensure that front and backend are served up from the same domain before providing the download status tracker.

stopfstedt:
Fallback: show a  flash message if front and backend are not served up from the same domain:

selection_014

app/components/curriculum-inventory-report-header.js
```diff
@@ -17,16 +20,54 @@ const Validations = buildValidations({
...
     this._super(...arguments);
     this.set('reportName', this.get('report.name'));
     this.set('isFinalized', this.get('report.isFinalized'));
+
+    // check if the d/l link points at the same domain as the current page is on.
+    let parser = document.createElement('a');
+    parser.href = this.get('report.absoluteFileUri');
+    let backendDomain = parser.hostname;
+    let frontendDomain = window.location.hostname;
+    this.set('downloadFromSameDomain', (backendDomain === frontendDomain));
```
jrjohnson: Putting all of this into the didReceiveAttrs() hook has gotten me into trouble in the past.  It doesn't seem to run every time.  Can you create computed properties for backendDomain, frontendDomain, and downloadFromSameDomain instead?  That will for sure update incase the report.absoluteFileUrl or the report itself gets modified.

stopfstedt: i don't think that's necessary in this particular situation. `didReceiveAttrs` _will_ re-fire each time `report` changes.

"Since the didReceiveAttrs hook is called every time a component's attributes are updated whether on render or re-render, you can use the hook to effectively act as an observer, ensuring code is executed every time an attribute changes."

https://guides.emberjs.com/v2.6.0/components/the-component-lifecycle/#toc_formatting-component-attributes-with-code-didreceiveattrs-code

jrjohnson: What about if just the `absoluteFileUri` changes? It isn't a very likely change given that it is generated on the server.  But if some kind of future enhancement updated that field then the download wouldn't work.  That is where I got into trouble with this on the user profile page. Changes to properties on the object don't trigger an update.

Issue #1935, opened 2016-08-08
Deleted Course / Bad URL Fails Incessantly


dartajax:
While verifying that a course (#800) had been deleted using the following link ...

https://curriculum.ucsf.edu/courses/800

The squiggly line (shown here in snap shot) just goes back and forth on the screen.  It should display an error if the URL passed was bad.

![image](https://cloud.githubusercontent.com/assets/7493331/17496118/b1c64c76-5d70-11e6-9051-a4615820e29e.png)

jrjohnson:
Closed in #2204

Issue #1715, opened 2016-05-22
error message traps users


saschaben:
right now when an error page comes up, the user needs to have a full refresh to get back to the application. Until this pattern is altered, every error page should include the message "Please hit the 'refresh' button on your browser to return to Ilios."

jrjohnson:
Closed in #2204

Issue #994, opened 2015-11-03
When you are viewing an error page, the menu is still present.


saschaben:
No navigation should appear on the error page.

jrjohnson:
Closed in #2204

PR #2204, opened 2016-10-07
Improved error handling

### LINKED ISSUE #1772: Handle 500 Errors better
When the API server returns a 500 error we are currently bubbling that all the way to ember data which reads it as 'Empty Content Type' which is not parseable by users and difficult to debug.  Instead we need to trap 500 errors earlier and notify the user that there is a problem connecting to the server and probably not send that error detail back to the server since the failure was there and not on the client.

### LINKED ISSUE #2087: Display 404 page when model is missing
When a user visits a model that cannot be found in the API we need to show a 404 error and not a generic error page.

jrjohnson:
Several improvements to overall error handling:
1) When a model is not found display a 404 instead of a generic error
message. Fixes #2087 #1935
2) When on the error route clicking a link in navigation clears errors. Fixes #1772 #1715 #994
3) Provided a button to escape from error page
4) Wen all else has failed provide a refresh the browser button as well.


jrjohnson:
Deploy preview ready!

Built with commit d8210ce28ca51d25afbdbbb606d7e9d6b8b962e9

http://deploy-preview-2204.ilios-frontend.netlify.com

saschaben:
much improved. :+1:

thecoolestguy:
@homu r+

PR #2249, opened 2016-10-18
check if competencies array is empty before rendering a list.

### LINKED ISSUE #393: Zero competencies - remove box
It would be great if we could hide the empty box when there are no competencies - it looks kinda strange since we don't show anything when there are 0 Learning Mats etc...

screen shot 2015-03-30 at 12 59 11 pm

stopfstedt:
fixes #393

### DIFF
--- a/app/components/school-competencies-list.js
+++ b/app/components/school-competencies-list.js
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 
 const { Component, computed } = Ember;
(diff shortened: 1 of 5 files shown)


jrjohnson:
Deploy preview ready!

Built with commit c0b9de23b132375f87e63d3bf833fe11d7e96fc7

https://deploy-preview-2249--ilios-frontend.netlify.com

Issue #2243, opened 2016-10-17
ILIOS_FRONTEND_API_HOST doesn't handle slash well


jrjohnson:
If you specify this env variable with a trailing slash you end up with a double // on requests to the server which then die.  We should handle a trailing slash better.

stopfstedt:
i don't really see the issue here. as long as we know that the backend URL is not coming with a trailing slash out of the config, and we know that reliably, then we can code accordingly and always add the slash ourselves.

stopfstedt:
fixed via #2310

PR #2269, opened 2016-10-23
ensure trailing slash on host-portion of rollover endpoint.

### LINKED ISSUE #2268: course rollover url broken when proxying to backend
this is a regression that happened after moving backend config into `.env`. we've lost the trailing slash.

selection_038

stopfstedt:
fixes #2268.

### DIFF
--- a/app/components/course-rollover.js
+++ b/app/components/course-rollover.js
@@ -79,7 +79,7 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
     if (expandAdvancedOptions && skipOfferings) {
       data.skipOfferings = true;
     }
-    const host = this.get('host')?this.get('host'):'/';
+    const host = this.get('host') ? (this.get('host').replace(/\/+$/, '') + '/') : '/';
     const namespace = this.get('namespace');
 
     let url = host + namespace + `/courses/${courseId}/rollover`;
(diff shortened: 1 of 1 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 4899235390a753d1197f14c08e3738a73e2d722b

https://deploy-preview-2269--ilios-frontend.netlify.com

app/components/course-rollover.js
```diff
@@ -79,7 +79,7 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
     if (expandAdvancedOptions && skipOfferings) {
       data.skipOfferings = true;
     }
-    const host = this.get('host')?this.get('host'):'/';
+    const host = this.get('host') ? (this.get('host').replace(/\/+$/, '') + '/') : '/';
```
jrjohnson: This feels like it can be applied generally and belongs in the config service instead of dealing with it every time we ask for the host somewhere.

jrjohnson: Doing this would also fix #2243

stopfstedt:
ok, then how about assuming no trailing slash _always_? 

our home-grown config service does not read environment variables; this is handled via third-party vendor lib [ember-cli-servervariables](https://github.com/blimmer/ember-cli-server-variables). now, we could wrap around this service and provide a `getBackendHost()` method that would append the trailing slash on, but IMO that's not worth the troubles.

PR #2270, opened 2016-10-24
Activate rule to prevent self closing tags

jrjohnson:
HTML5 doesn't not allow self closing tags.  Void elements like `<input>`
and `<br>` should end without closing.  Other elements require a closing
tag even if they are empty.


jrjohnson:
Deploy preview ready!

Built with commit 96c149c03afecaf1b304c8c1d092aec2b16199fd

https://deploy-preview-2270--ilios-frontend.netlify.com

PR #2273, opened 2016-10-25
reworked 'save' task as action.

### LINKED ISSUE #2179: Error Received When Rolling Over a Course
image

stopfstedt:
fixes #2179.

Turns out that using a task here, and checking for the `isRunning` attribute of that task in the template (e.g. to lock buttons and input fields), is not going to work.
Ember will fail to change the component's state (see the aforementioned `isRunning` toggle) after transitioning to the new route. So we can't do it like that.
My proposed solution converts the task into an action, which works.

### DIFF
--- a/app/components/course-rollover.js
+++ b/app/components/course-rollover.js
@@ -52,50 +52,7 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
   startDate: null,
   skipOfferings: false,
   title: null,
-
-  save: task(function * (){
-    yield timeout(10);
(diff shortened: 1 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 7e5d37803ce0f7dd09f77f775ae5c96eaecee9f5

https://deploy-preview-2273--ilios-frontend.netlify.com

stopfstedt:
found a better solution. closed.

PR #2275, opened 2016-10-25
Better File Upload

### LINKED ISSUE #1245: Trap file upload errors and give the user more info
We should check file size before uploading:
http://stackoverflow.com/a/4307882

If we get other errors back then we should display a nice message to the user about what went wrong.  Probably requires:
https://github.com/ilios/ilios/issues/1196

### LINKED ISSUE #2060: A Better Enforcement for File Size of LM's Is Needed
Currently, if a user tries to upload a file that is too large (larger than 110MB), this is what is returned.  

image

There needs to be a message returned to the user advising about what just happened -- or a message indicating that file chosen is too large and can't be uploaded preventing the upload attempt from even taking place.

jrjohnson:
Improved the process for uploading a learning material file in several ways.
1) Upload files to the API server based on the configuration and not to /upload on whatever host the fronted is on.
2) Detect if a file is too large and warn the user that it cannot be uploaded.
3) If an upload fails because of network or server issues try it again up to 3 times transparently before ultimately throwing an exception and notifying the user of the issue.

Fixes #2060
Fixes #1245

### DIFF
--- a/app/components/file-upload.js
+++ b/app/components/file-upload.js
@@ -1,8 +1,9 @@
 import Ember from 'ember';
 import EmberUploader from 'ember-uploader';
+import readableFileSize from 'ilios/utils/readable-file-size';
 
 const { FileField, Uploader } = EmberUploader;
-const { inject, computed } = Ember;
+const { inject, computed, isEmpty } = Ember;
 const { service } = inject;
 
 let IliosUploader = Uploader.extend({
@@ -17,6 +18,8 @@ let IliosUploader = Uploader.extend({
 
 export default FileField.extend({
   session: service(),
+  iliosConfig: service(),
+  i18n: service(),
   url: '',
   headers: computed('session.isAuthenticated', function(){
     let headers = {};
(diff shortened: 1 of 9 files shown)


jrjohnson:
Deploy preview ready!

Built with commit b5ec64cd4c8c3d0f431b2bc9c921c2f059b664a6

https://deploy-preview-2275--ilios-frontend.netlify.com

jrjohnson:
@dartajax this needs to be click tested.  File uploads for learning materials.

saschaben:
issue: when retrying to attach a file, it is not obvious to the user what is happening. If possible, provide a "whoops! something went wrong: retrying" message when re-attempting. If we can't that's OK, but it would be optimal if we can.

issue: oversize files just throw the network error message after multiple retries. Doesn't ID that the trouble is the file is too big.  currently testing with the 215MB uncompressed docker app on my mac.

dartajax:
Is the maximum supposed to be 50MB?

dartajax:
Granted I had to pick a 40MB file and unplug my network connection to get this but when the connection is "dropped", the screen hangs for a while and then returns this error.

image

dartajax:
:+1:

PR #2294, opened 2016-10-27
show spinner and lock 'show more' button while search is running

### LINKED ISSUE #949: mesh and lm search "show more" button needs an in process spinner
since the pagination calls can take a number of seconds, we should add a spinner to the "show more" button when there is a chance to do so.

stopfstedt:
fixes #949 

Learning material search:

selection_051

MeSH search:

selection_048

### DIFF
--- a/app/components/learningmaterial-search.js
+++ b/app/components/learningmaterial-search.js
@@ -18,6 +18,25 @@ export default Component.extend({
   searching: false,
   searchReturned: false,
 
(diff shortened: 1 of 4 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 89bbc74e2e739f33153b71e91ff4e6a333cc8625

https://deploy-preview-2294--ilios-frontend.netlify.com

PR #2297, opened 2016-10-28
Style new-myreport specifically.

### LINKED ISSUE #1139: MyReports "which is" Drop-Down needs to be limited in Width
See the screen shot below ...

image

There should be an indicator that this is a drop-down menu.  The reason is shown below ..

I couldn't show the drop-down but the following SOD topic is way too long and causes the whole drop-down to be far too wide and hence not indentifiable as a drop-down.

image

In other words we need to word wrap after a certain point and not allow the drop-down to grow in width without restriction.

jrjohnson:
Remove generic form styles and specifically style this component so that
we can fix issues with large select lists.

Fixes #1139

### DIFF
--- a/app/components/new-myreport.js
+++ b/app/components/new-myreport.js
@@ -21,7 +21,7 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
   i18n: service(),
   currentUser: service(),
   flashMessages: service(),
-  classNames: ['form-container', 'detail-view', 'new-myreport', 'mesh-manager'],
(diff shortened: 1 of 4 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 38861b31dfc0a012f94c2180b4c7154cf7f13a18

https://deploy-preview-2297--ilios-frontend.netlify.com

dartajax:
Nice :+1:

PR #2299, opened 2016-10-28
Validate API Version

### LINKED ISSUE #2284: Check API version for mis-match
After https://github.com/ilios/ilios/issues/1655 we should look at the API version when the frontend loads and if it doesn't match then we need to display an error message.  Otherwise a version mis-match could cause data to be removed accidentally when old models are PUT to the API.

jrjohnson:
If a frontend attempts to contact an API with a mismatched version
string the user will see a message an be unable to continue. This will
prevent any kind of weird situation where an incorrect version gets
shipped or an advanced setup is set to point to an older API install.

Fixes #2284

### DIFF
--- a/app/components/api-version-check.js
+++ b/app/components/api-version-check.js
@@ -1,0 +1,20 @@
+import Ember from 'ember';
+import ENV from 'ilios/config/environment';
+
+const { Component, RSVP, computed, inject } = Ember;
+const { Promise } = RSVP;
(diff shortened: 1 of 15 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 3de65a8a7fd7034c2577936172876ca5ed2a2ba5

https://deploy-preview-2299--ilios-frontend.netlify.com

PR #2318, opened 2016-11-03
Update EmberJS and Ember Data to the latest versions

jrjohnson:
We avoided doing this before when I thought glimmer2 was included in this update and needed to be tested, however that wasn’t the case - this is essentially 2.8.3 as it has very little new in it and should be safe to update to.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "ilios",
   "dependencies": {
-    "ember": "~2.8.2",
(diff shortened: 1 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 65186edf9432f256f6c1129eb84445d773caad69

https://deploy-preview-2318--ilios-frontend.netlify.com

PR #2364, opened 2016-11-17
Remove code climate test reporter

jrjohnson:
We are no longer using this and it has an incompatibility which makes it
impossible to run ember s.


jrjohnson:
Deploy preview ready!

Built with commit 83dc29d91b3bb54f1df426edc784d131aa288e95

https://deploy-preview-2364--ilios-frontend.netlify.com

PR #2373, opened 2016-11-21
Cache frontend using service workers

### LINKED ISSUE #1823: Add caching to improve initial boot time
Once a user has downloaded out app they shouldn't need to download it again.  If we setup a cache this will be true.

AppCache is sort of insanely hard to use well.  Service workers seems to be a better choice, but support is still limited in some browsers.  However, I think in this case we should implement the forward thinking solution that only works in chrome and firefox.  Safari and IE will catch up and then we will be in good shape.

Blockers:
- [ ] Need to be able to get the script tag from ember-cli-deploy-jason-config into S3
- [ ] Change the way we parse JSON in symfony to grab the scripts

jrjohnson:
After a lot of experimentation I don't think we're ready for offline fallback of API requests. What we *are* ready for is aggressive caching and background loading of assets. This change will cause the app to boot much faster the second time it is used because all of the JS and CSS will be pre-loaded by the browser.

- [x] Needs merge of upstream PR
- [x] Will require backend modifications
- [x] Should probably be an API bump depending on backends ability to handle changes
- [x] Need to figure out how to version the sw.js file or at least separate staging and prod versions.

Fixes #1823


jrjohnson:
Deploy preview ready!

Built with commit 7b25208c723ea02f149bf922f1a99f361096e82d

https://deploy-preview-2373--ilios-frontend.netlify.com

PR #2378, opened 2016-11-22
added citation link to print course view

### LINKED ISSUE #2139: Citation LM's Need to Display The Citation - Print View
The citation itself needs to be displayed in the Summary Print View of a Course.

Currently, this is what shows up.  We will be adding links and reorganizing the screen significantly.  Even so, more information should be visible immediately from the Print View.

image

gboushey:
fixes #2139 

Added citation column to learning materials in print course view.

Note that this field only exists when learning material is a citation.  As a result, most learning materials will have a blank value for the citation column.  I think that citation is the only learning material type that has a "citation" field, whereas all other fields (display name, type, required, notes) are common to all learning material types (link, file, etc).  

As a result, we are taking up some real estate in the learning materials table for what may usually be an empty field.  There might be a better way to display citations (for example, we could add them to the notes field, which exists for all learning materials, rather than creating a new column).


jrjohnson:
Deploy preview ready!

Built with commit a11f8454217ff1ccee66055e830fe4345adaeb6b

https://deploy-preview-2378--ilios-frontend.netlify.com

.tm_properties
```diff
@@ -0,0 +1,3 @@
+softWrap = true
```
jrjohnson: Should be ignored.  Probably in your ~/.gitignore_global

app/templates/components/print-course.hbs
```diff
@@ -113,6 +113,7 @@
                   <th class='text-center'>{{t 'general.type'}}</th>
                   <th class='text-center'>{{t 'general.required'}}</th>
                   <th class='text-left'>{{t 'general.notes'}}</th>
+                  <th class='text-left'>{{t 'general.citation'}}</th>
```
jrjohnson: I believe there are very few learning materials that are citations.  @saschaben is it worth spending the screen real-estate on a column for this or should it be squeezed into the title column?

saschaben:
@jrjohnson  @gboushey  adding this to the notes field doesn't compute: notes act already in dual purpose as either specific action notes to students, OR as instructional information seen only by faculty.  Citation info is requisite.  We could extend the value of the new occupied real estate by making it a "Description / Citation"  field;  that would make it more effective generally for non-citations. The descriptions do have the potential to run long, but a quick review of our local  data shows only a few dozen LMs with descriptions greater than 750 characters, and only a fraction of those which are citations. So this might be an opportunity to extend what we are able to present, as well as get the damned citations to be visible.

gboushey:
We might want to expand the allocation of space for descriptions with a larger column width in the table and reduce space for "Type" and "Required"

gboushey:
There are still a few issues with this pull request.  

- I'm using percentages to set column width for the learning materials table in print-course.js.  Currrently, allotment for the description field (which also includes citations) is 50%.  Move this as a fixed length to ilios.css?

- Descriptions are enclosed in <p></p> tags.  This means that descriptions are automatically offset in the description, which throws off the alignment.  We probably do want to keep and apply the html, since descriptions can be long and need the formatting and line breaks, just not (in this case) at the top.  @jrjohnson any thoughts on how to do this?

jrjohnson:
I don't love using width over colspan, but I'm guessing you went this way for a reason.  Yes this style stuff should get moved into css and out of the HTML.

Once that is done then you can also apply a 
```css
p {
  margin: 0;
}
```

style as well to hopefully remove the offset.

gboushey:
I changed to colspan, not a problem.

I'd rather not style all `<p>` tags in the description,  since some of them may have been added as markup by the admin user as part of the description. The user isn't adding the outer `<p>...</p>` to the description, those appear to be added by the system.  Could we remove the system added tags from the description and put them into the page where paragraph tags around the description are required?

 Not sure how extensive a change this is, happy to look into it though.

gboushey:
@saschaben @jrjohnson This latest pull request strips html formatting from the description field in the print course page.  This avoids collision of html tags with formatting within a cell (which is not typical for how we show descriptions in ilios, which is probably why the paragraph tags haven't caused alignment issues before).  It could lead to a long monolithic block of text if the description is very long though.  We also might want to consider capping the description display to a limited number of characters to avoid throwing off alignment and formatting of the table.

app/templates/components/print-course.hbs
```diff
@@ -113,23 +113,31 @@
...
                     </td>
-                    <td class='text-left'>{{lm.learningMaterial.notes}}</td>
+                    <td class='text-left text-top'> 
+                      {{#if lm.notes}}
+                        <span class='add'>{{t 'general.yes'}}</span>
+                      {{else}}
+                        <span class='remove'>{{t 'general.no'}}</span>
+                      {{/if}}
+                    </td>
+                    <td class='text-left text-top' colspan=4>{{sanitize-html lm.learningMaterial.description}} {{lm.learningMaterial.citation}}</td>
```
jrjohnson: Instead of sanitizy-html you can use an `{{big-text text=lm.learningMaterial.description}}` it will strip the HTML and also an ellipse to trim the text for expansion if it is too long.

tests/acceptance/course/printcourse-test.js
```diff
@@ -56,6 +56,8 @@ module('Acceptance: Course - Print Course', {
       userRole: 1,
       copyrightPermission: true,
       courseLearningMaterials: [1],
+      citation: 'Lathrop, Emmett, Flux Capacitor, Journal of Time Travel, 5 Nov 1955',
```
jrjohnson: 💯

gboushey:
Cool, big text works nicely.  I rolled back to avoid including any of the various js files that ember-sanitize might drop into the app now that we aren't using it.  

@dartajax @saschaben Dave, would you mind taking a look at this?  big-text should mitigate some of the problems that can arise from putting lengthy text with html tags into a table cell.  Still, it would be a good idea to see how this holds up under various input (long citations, strange formatting, lengthy descriptions, etc).

jrjohnson:
After discussion this morning the path forward is to not use `big-text` because it trims the description too much to be useful.  Instead LM description should be stripped of HTML but displayed as is.

app/templates/components/print-course.hbs
```diff
@@ -210,7 +218,7 @@
 
       <section class='detail-block'>
         <div class='detail-title'>
-          {{t 'general.learningMaterials'}} ({{session.sessionLearningMaterials.length}})
+          {{t 'general.learningMaterials'}}<br/>({{session.sessionLearningMaterials.length}})
```
jrjohnson: 'Br' here on purpose?

tests/unit/helpers/remove-html-tags-test.js
```diff
@@ -0,0 +1,12 @@
+
+import { removeHtmlTags } from 'ilios/helpers/remove-html-tags';
+import { module, test } from 'qunit';
+
+module('Unit | Helper | remove html tags');
+
+// Replace this with your real tests.
```
jrjohnson: This comment should go.

.gitignore
```diff
@@ -18,3 +18,4 @@ testem.log
 lcov.dat
 sauce-example.log
 .env
+.tm_properties
```
jrjohnson: We don't have any other IDEs in here. @stopstedt thought on including specific ignore stuff?

stopfstedt:
@jrjohnson you are correct. IDE specific excludes should be defined outside this repo, see section on global gitignore here https://help.github.com/articles/ignoring-files/

gboushey:
Notes are in the commit - one thing, I did alter .eslintrc to allow jquery.  I'm using a jquery method to remove html rather than writing my own regex.

app/helpers/remove-html-tags.js
```diff
@@ -0,0 +1,10 @@
+import Ember from 'ember';
+
+export function removeHtmlTags(params) {
+  if (!(params[0] === undefined)) {
+    return jQuery("<p>" + params[0] + "</p>").text();
```
jrjohnson: The easiest (and most reliable) way to get access to jQuery is with `Ember.$` instead of using the global.

PR #2409, opened 2016-12-06
Temporarily disable userId in Google Analytics

### LINKED ISSUE #2408: User permissions not re-setting after login
Immediately after logging in a user doesn't get all of the permissions they deserve.

jrjohnson:
Looking up the user on the login screen when they were not authenticated
causes their permissions to be locked at that point.  Those sections of
the app which rely on the currentUser properties for access to services
do not work correctly until the page is refreshed.

Fixes #2408


jrjohnson:
Deploy preview ready!

Built with commit 66be26c5fc67293be90ed4555108c82664d4c806

https://deploy-preview-2409--ilios-frontend.netlify.com

PR #2441, opened 2016-12-14
Replace deprecated contains with includes

jrjohnson:
The `contains` method on ennumerables was deprecated because correct ES
syntax is to use `includes`.


jrjohnson:
Deploy preview ready!

Built with commit 02000c25274c31945460d2bb497c4ed0f730f4dd

https://deploy-preview-2441--ilios-frontend.netlify.com

Issue #2506, opened 2017-01-11
all X for Instructor Y report throws error


saschaben:
attempting to run any report which requires selection of an instructor throws an error on the attempt to search for a name:

```
Cannot read property 'includes' of null

TypeError: Cannot read property 'includes' of null
    at a.<anonymous> (https://d26vzvixg52o0d.cloudfront.net/assets/ilios-b1a704a492f2f86452a77cf468ffd7cd.js:10:18569)
    at c.p.get (https://d26vzvixg52o0d.cloudfront.net/assets/vendor-47a3abc3eb8af5f1d8abe0c7961b893f.js:11:10306)
    at Object.r [as get] (https://d26vzvixg52o0d.cloudfront.net/assets/vendor-47a3abc3eb8af5f1d8abe0c7961b893f.js:12:17420)
    at t.compute (https://d26vzvixg52o0d.cloudfront.net/assets/vendor-47a3abc3eb8af5f1d8abe0c7961b893f.js:10:26754)
    at t.value (https://d26vzvixg52o0d.cloudfront.net/assets/vendor-47a3abc3eb8af5f1d8abe0c7961b893f.js:10:25436)
    at t.e.value (https://d26vzvixg52o0d.cloudfront.net/assets/vendor-47a3abc3eb8af5f1d8abe0c7961b893f.js:20:19540)
    at e.initialize (https://d26vzvixg52o0d.cloudfront.net/assets/vendor-47a3abc3eb8af5f1d8abe0c7961b893f.js:17:22745)
    at e.peek (https://d26vzvixg52o0d.cloudfront.net/assets/vendor-47a3abc3eb8af5f1d8abe0c7961b893f.js:17:22404)
    at t.evaluate (https://d26vzvixg52o0d.cloudfront.net/assets/vendor-47a3abc3eb8af5f1d8abe0c7961b893f.js:19:26485)
    at e.execute (https://d26vzvixg52o0d.cloudfront.net/assets/vendor-47a3abc3eb8af5f1d8abe0c7961b893f.js:21:23895)
```

dartajax:
I noticed that previously saved reports are fine - this is with the creation of a new report.

PR #2515, opened 2017-01-12
Update reports to include terms

gboushey:
Added learner group and objective to subject list  added term as prepositional object (filter) for course, session, program, program year, session type, learner group, and objective.  Note that there are still issues with learner group and objective, it is not currently possible to filter these by term, an error will result

### DIFF
--- a/app/services/reporting.js
+++ b/app/services/reporting.js
@@ -209,5 +209,23 @@ export default Service.extend({
       };
     });
     return RSVP.resolve(map);
-  }
+  },
+  termsResults(results){
+    return RSVP.map (results.toArray(), result => {
+      return new RSVP.Promise(resolve => {
+        result.get('vocabulary').then(vocabulary => {
+          result.get('titleWithParentTitles').then(titleWithParentTitles => {
+            const title = vocabulary.get('title') + ' > ' + titleWithParentTitles;
+            resolve({value: title});
+          });
+        });
+      });
+    });
+  },
+  objectivesResults(results){
+    return this.titleResults(results);
+  },
+  learnerGroupsResults(results){
+    return this.titleResults(results);
+  },
 });
--- a/app/components/new-myreport.js
+++ b/app/components/new-myreport.js
@@ -6,6 +6,7 @@ import ValidationErrorDisplay from 'ilios/mixins/validation-error-display';
 const { Component, computed, inject, RSVP, isEmpty, isPresent } = Ember;
 const { service } = inject;
 const { PromiseArray } = DS;
+const { oneWay } = computed;
 
 const Validations = buildValidations({
   title: [
@@ -40,6 +41,7 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
       {value: 'learning material', label: this.get('i18n').t('general.learningMaterials')},
       {value: 'competency', label: this.get('i18n').t('general.competencies')},
       {value: 'mesh term', label: this.get('i18n').t('general.meshTerms')},
+      {value: 'term', label: this.get('i18n').t('general.terms')},
       {value: 'session type', label: this.get('i18n').t('general.sessionTypes')},
     ];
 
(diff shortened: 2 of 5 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 95392851e0564160b254c6ce1b62e254498a3579

https://deploy-preview-2515--ilios-frontend.netlify.com

jrjohnson:
Can you also add `term` to the fields where it is the subject?  Then we can test both axes of the report table and see what shakes out.

gboushey:
currently unable to get session types, learner groups, or objective by associated terms (no association error).  

currently unable to get terms for 'course', 'session', 'program', 'program year', 'session type', 'learner group', 'objective'

Some additional factors:

web services to find terms for course, session, program, program year are working properly.
Sample API endpoints for these all return data: 

/api/v1/terms?filters[courses]=331
/api/v1/terms?filters[sessions]=32469
/api/v1/terms?filters[programs]=1
/api/v1/programyears?filters[terms]=131

However the frontend app responds with a TypeError: _this[mapper] is not a function.  The services do exist, so this is probably an issue with the frontend.  

No web service exists(?) for session types, learner groups, and objectives.  In this case, we get a 505 error from frontend and a "has no field or association" error from ilios backend.  So for these, I'm assuming we need to create a new api endpoint and service.

gboushey:
I added objectives to the report as a subject and as a prepositionalObject.  My guess is that we won't want this in all spots, since objectives can be a very long list (especially for "associated with", where they all appear in a drop down list).  I put it in here so that we can run the reports with existing associations for objectives to get a sense of how to handle them.

dartajax:
Hey @gboushey can you do rebase on this one?  I have been wanting to test it but am getting the following error ...

image

saschaben:
1/31 looking really good at this point. *Action Item* - we still need to pull OBJECTIVE and LEARNER GROUP out of the select lists where they appear. We will add them back in after the remodel of the whole reports implementation.

jrjohnson:
Deploy preview ready!

Built with commit 6f64579915a29320e7cf13ae3594067f9c57ea0c

https://deploy-preview-2515--ilios-frontend.netlify.com

jrjohnson:
@gboushey couple of notes:
1) I added a tests for new-myreports when fixing #2506 which checked the dropdowns - that is why your PR is failing after you merged.  You will have to modify https://github.com/ilios/frontend/blob/master/tests/integration/components/new-myreport-test.js now that terms are in the list.
2) Can you rebase this?  I want to eliminate the merge commit and the last minor change one.

saschaben:
OK. Structurally we are good here. in QA I am getting some data anomalies though.

1: run report for Medicine > All terms for course > Airway, blood, circulation (2016-17) (it's the 2nd one listed).
2: check the return values, note the inclusion of "Resources > Wet Laboratory".
3: run report for Medicine > All courses for term > "Reources > Wet Laboratory".
4: Note that the report comes up with no return values (it should return Airway, Blood, Circulation).

Let's validate that the queries for "All term X for value Y" and "All value X for Term Y" match in result sets. Right now it looks like they do not (at least for course).

jrjohnson:
Issue with term results is a missing API filter https://github.com/ilios/ilios/issues/1737

app/services/reporting.js
```diff
@@ -209,5 +209,23 @@ export default Service.extend({
...
+        result.get('vocabulary').then(vocabulary => {
+          result.get('titleWithParentTitles').then(titleWithParentTitles => {
+            const title = vocabulary.get('title') + ' > ' + titleWithParentTitles;
+            resolve({value: title});
+          });
+        });
+      });
+    });
+  },
+  objectivesResults(results){
```
jrjohnson: Leftover - I think it does no harm though.

app/services/reporting.js
```diff
@@ -209,5 +209,23 @@ export default Service.extend({
...
+    return this.titleResults(results);
+  },
+  learnerGroupsResults(results){
```
jrjohnson: Left over - but probably benign.

PR #2531, opened 2017-01-18
Update ajax headers when we get a new token

### LINKED ISSUE #2527: Sending wrong token on second refresh
Our JWTs automatically refresh as they approach their expiration time.  The first refresh works just fine it grabs a new token which is then used for all subsequent interactions with the API.  The problem is that the second time we refresh it does not send the new JWT instead it sends the original login JWT.  This causes the second refresh to fail and then user to be logged out.

jrjohnson:
The CP which loads the Token into AJAX requests was not depending on the
right properties of the session.  This caused refreshes of the token to
not be reflected in traffic we were sending to the server.

Fixes #2527

### DIFF
--- a/app/services/ajax.js
+++ b/app/services/ajax.js
@@ -11,9 +11,10 @@ export default AjaxService.extend({
 
   host: reads('iliosConfig.apiHost'),
 
-  headers: computed('session.isAuthenticated', function(){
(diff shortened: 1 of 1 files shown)


jrjohnson:
Deploy preview ready!

Built with commit cc302be98654f2f76485cba51b06138380be1461

https://deploy-preview-2531--ilios-frontend.netlify.com

PR #2606, opened 2017-02-05
Add school session attribute configuration

### LINKED ISSUE #2503: Add per-school ability to enable/display specific course/session attributes
Several schools have expressed the need for being able to remove default course/session attributes that they do not use, and to be able to add/enable others that are not provided by default.  We need to be able to add new attribute fields to the database by request, and in-turn allow school's to enable the functionality and display of these attributes as needed on their respective install.

In Ilios 2, school's could re-label an attribute they don't use (eg, "Special Attire Required") in order to reappropriate it for another not already included in the system (eg, "Attendance Required"). 

Due to the nature of the download-on-demand Frontend of Ilios 3, schools do not have access to which attributes should or shouldn't be displayed, nor do they have the ability to change their respective label upon display.

We need to investigate and implement the best route to allow for this to happen.  Currently, the processflow being considered is the following:

1. Schools request the desired custom attribute they need and we add it to the database via a migration.
2. We provide an interface for them to enable/display the extra attributes.
3. Attributes should have 3 states: Never Set (null), enabled/displayed, disabled/not displayed.

jrjohnson:
This allows schools to remove choices from the session attributes so
they can enable only those which are needed.

Fixes #2503

WIP:
- [x] Translation
- [x] API support for configuration
- [x] API version bump
- [x] Other tickets to cover using this setting for display of session settings and user calendars.
- [x] ~Maybe turn schoolConfig into a service so it can be used easily elsewhere to get values~
- [x] Change `key` to `name` to match API


jrjohnson:
Deploy preview ready!

Built with commit b76097c232ecb3bdc32fcc3ff72e3cc3c466dc94

https://deploy-preview-2606--ilios-frontend.netlify.com

app/components/school-session-attributes-expanded.js
```diff
@@ -0,0 +1,109 @@
+import Ember from 'ember';
+import { task } from 'ember-concurrency';
+
+const { Component, RSVP, inject, computed, isEmpty } = Ember;
+const { service } = inject;
+const { all } = RSVP;
+
+export default Component.extend({
```
stopfstedt: i see a lot of similarities between this and the corresponding collapsed component. 
if possible, refactor this out, or perhaps extend the collapsed component (from what i can see, only `tagName` would have to be overridden).

stopfstedt:
otherwise, this LGTM.

jrjohnson:
@stopfstedt ready for review again.  I tried to test these new methods on the school model but I was unable to work around Ember's async testing issues.

Issue #2034, opened 2016-08-25
replace any usage of DS.PromiseObject and DS.PromiseArray


stopfstedt:
Find and replace all occurences within these building blocks

- [x] mixins
- [ ] models
- [x] components
- [x] services
- [ ] controllers

find all occurrences by running `egrep -r 'Promise(Array|Object)' app` from the command line and within your application's root directory.

stopfstedt:
remaining items: 

```bash
 egrep -r 'PromiseArray|PromiseObject' app
app/models/session.js:const { PromiseArray, PromiseObject } = DS;
app/models/session.js:      return PromiseArray.create({
app/models/session.js:    return PromiseObject.create({
app/models/session.js:    return PromiseObject.create({
app/models/session.js:    return PromiseObject.create({
app/models/session.js:    return DS.PromiseArray.create({
app/models/session.js:    return DS.PromiseArray.create({
app/models/session.js:    return DS.PromiseArray.create({
app/models/school.js:    return DS.PromiseArray.create({
app/models/school.js:    return DS.PromiseArray.create({
app/models/learner-group.js:    return DS.PromiseArray.create({
app/models/learner-group.js:    return DS.PromiseObject.create({
app/models/learner-group.js:    return DS.PromiseArray.create({
app/models/learner-group.js:    return DS.PromiseArray.create({
app/models/objective.js:const { Model, attr, belongsTo, hasMany, PromiseArray } = DS;
app/models/objective.js:    return PromiseArray.create({
app/models/objective.js:    return PromiseArray.create({
app/models/program-year.js:    return DS.PromiseArray.create({
app/models/ilm-session.js:const { PromiseArray } = DS;
app/models/ilm-session.js:    return PromiseArray.create({
app/models/cohort.js:const { Model, PromiseArray } = DS;
app/models/cohort.js:    return PromiseArray.create({
app/models/user.js:const { PromiseArray } = DS;
app/models/user.js:    return PromiseArray.create({
app/models/user.js:      return PromiseArray.create({
app/controllers/pending-user-updates.js:const { PromiseArray } = DS;
app/controllers/pending-user-updates.js:    return PromiseArray.create({
app/controllers/courses.js:const { PromiseArray } = DS;
app/controllers/courses.js:    return PromiseArray.create({
app/controllers/courses.js:    return PromiseArray.create({
app/controllers/courses.js:    return PromiseArray.create({
app/controllers/courses.js:      return PromiseArray.create({
[...]

PR #2622, opened 2017-02-08
replaced DS.PromiseArray with RSVP.Promise in dashboard-calendar component

stopfstedt:
did some bolshoi re-structuring/cleanup while at it.

refs #2034.

### DIFF
--- a/app/components/dashboard-calendar.js
+++ b/app/components/dashboard-calendar.js
@@ -60,6 +60,15 @@ export default Component.extend({
   calendarDate: momentFormat('selectedDate', 'YYYY-MM-DD'),
 
   /**
+   * @property courses
+   * @type {Ember.computed}
+   * @public
+   */
+  courses: computed('allCourses.[]', 'selectedCourses.[]', function(){
+    return this.get('allCourses');
+  }),
+
+  /**
    * @property filteredEvents
    * @type {Ember.computed}
    * @public
@@ -311,9 +320,7 @@ export default Component.extend({
       });
     });
   }),
-  courses: computed('allCourses.[]', 'selectedCourses.[]', function(){
-    return this.get('allCourses');
-  }),
+
   selectedSchool: computed('schoolPickedByUser', 'currentUser.model.school', function(){
     const schoolPickedByUser = this.get('schoolPickedByUser');
     return new Promise(resolve => {
--- a/app/components/dashboard-calendar.js
+++ b/app/components/dashboard-calendar.js
@@ -288,23 +288,27 @@ export default Component.extend({
     return this.get('allCohorts');
   }),
   selectedCourses: [],
+
+  /**
+   * @property allCourses
+   * @type {Ember.computed}
+   * @protected
+   */
   allCourses: computed('selectedSchool', 'selectedAcademicYear', function(){
-    let defer = defer();
-    this.get('selectedSchool').then((school) => {
-      this.get('selectedAcademicYear').then((year) => {
-        this.get('store').query('course', {
-          filters: {
-            school: school.get('id'),
-            year: year.get('title')
-          },
-          limit: 1000
-        }).then((courses) => {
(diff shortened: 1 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 4330fb05d5c79bd535325540cb660065c2f119ec

https://deploy-preview-2622--ilios-frontend.netlify.com

app/components/dashboard-calendar.js
```diff
@@ -22,8 +21,14 @@ export default Component.extend({
   currentUser: service(),
   store: service(),
   i18n: service(),
+  activeFilters: null,
+  schoolPickedByUser: null,
   selectedDate: null,
   selectedView: null,
+  selectedCohorts: [],
```
jrjohnson: These should be `null` initially otherwise the Array object can get passed by reference and not cleaned up when the component is removed from the page.

app/components/dashboard-calendar.js
```diff
@@ -57,232 +62,139 @@ export default Component.extend({
...
+    });
+  }),
+
+  /**
+   * @property cohorts
+   * @type {Ember.computed}
+   * @public
+   * @todo Check if this can be replaced with allCohorts(). [ST 2017/02/07]
+   */
+  cohorts: computed('allCohorts.[].displayTitle', 'selectedCohorts.[]', function(){
```
jrjohnson: Why does this CP depend on `displayTitle` and `selectedCohorts.[]` but not use them in the body fo the method?

app/components/dashboard-calendar.js
```diff
@@ -57,232 +62,139 @@ export default Component.extend({
...
+    return this.get('allCohorts');
+  }),
+
+  /**
+   * @property courseLevels
+   * @type {Ember.computed}
+   * @public
+   * @todo Review whether this needs to be a computed property, get rid of it if applicable. [ST 2017/02/07]
+   */
+  courseLevels: computed('selectedCourseLevels.[]', function(){
```
jrjohnson: I agree with you todo.  Seems like this could be done as either a static array or could be built in `init()` if the loop is variable.  CP which depends on something else is weird since this will always return [1,2,3,4,5] no matter what.

app/components/dashboard-calendar.js
```diff
@@ -57,232 +62,139 @@ export default Component.extend({
...
+    return levels;
   }),
+
+  /**
+   * @property courses
+   * @type {Ember.computed}
+   * @public
+   * @todo Check if this can be replaced with allCourses(). [ST 2017/02/07]
+   */
+  courses: computed('allCourses.[]', 'selectedCourses.[]', function(){
```
jrjohnson: I agree with your TODO.  This should be replaced with `allCourses`.

app/components/dashboard-calendar.js
stopfstedt: good question. no idea!

stopfstedt: see comment in docblock above. i'll see if i can refactor this away.

app/components/dashboard-calendar.js
stopfstedt: noted. will refactor.

app/components/dashboard-calendar.js
stopfstedt: on it, then.

app/components/dashboard-calendar.js
stopfstedt: agreed. working on it.

jrjohnson:
Code LGTM.  @dartajax please check the calendar on this PR and make sure filters do what they are supposed to and then this can be merged.

stopfstedt:
@dartajax please give this a quick round of manual testing.

all of this pertains to filtering of events in the dashboard calendar. please check that filtering by the various criteria still works as before.

if it holds up then please merge.

thanks!

dartajax:
LGTM :+1:

PR #2637, opened 2017-02-09
Replaced DS.PromiseArray with RSVP.Promise in detail-terms-list component

stopfstedt:
refs #2034.

### DIFF
--- a/app/components/detail-terms-list.js
+++ b/app/components/detail-terms-list.js
@@ -1,9 +1,7 @@
 import Ember from 'ember';
-import DS from 'ember-data';
 
 const { Component, computed, RSVP } = Ember;
 const { all, defer, Promise } = RSVP;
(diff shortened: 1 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit a75401b0534177a13c5e64a12d78a50a165045d6

https://deploy-preview-2637--ilios-frontend.netlify.com

app/components/detail-terms-list.js
```diff
@@ -82,7 +72,7 @@ export default Component.extend({
    *
    * @property filteredTerms
    * @type {Ember.computed}
-   * @public
+   * @protected
```
jrjohnson: This might be internal, but there is no reason it couldn't be used outside of this component.  All that is to say I don't think it should be marked private.

PR #2703, opened 2017-03-01
Handle page title and header better

### LINKED ISSUE #2680: Caption Does Not Update with Back Button Action
I was reviewing the My Materials from the upper right drop-down and then decided to click the back button on my browser (Chrome).  The caption did not update to reflect the fact that I am not back on My Profile.

image

jrjohnson:
Instead of setting this directly on the application controller we will
use ember-cli-document-title to consolidate titles from each of our
routes.  We then use those to build both the title for the page in
head.hbs and to create the header in the orange bar.

Fixes #2680


jrjohnson:
Deploy preview ready!

Built with commit 8971968afe714731827af31beb6b19fcc5bbcb0b

https://deploy-preview-2703--ilios-frontend.netlify.com

PR #2705, opened 2017-03-02
work with current user model directly to catch changes to its reports list

### LINKED ISSUE #2702: addition/removal of MyReports reports fails
when either deleting an existing report, or creating a new report, a page refresh is required to complete the action. until that is done, deleted reports will still run and be visible in the list, and new reports will not show up. (also it appears to randomly generate 500 server errors, but I can't nail down the process to replicate.)

stopfstedt:
fixes #2702

### DIFF
--- a/app/components/dashboard-myreports.js
+++ b/app/components/dashboard-myreports.js
@@ -1,24 +1,38 @@
 import Ember from 'ember';
+import { task } from 'ember-concurrency';
+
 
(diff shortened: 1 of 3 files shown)


jrjohnson:
Deploy preview ready!

Built with commit ef30159dcdcc8e83c7d5f10f6850022815d3ddd8

https://deploy-preview-2705--ilios-frontend.netlify.com

PR #2711, opened 2017-03-02
Check our addons for conflicting dependencies

jrjohnson:
When addons all depend on the same addon it can cause mysterious issues
as their output has the possibility of overriding each other.  This
ensures that when we add or update and addon that we test for this
problem.  The only conflict we currently have is with
ember-getowner-polyfill - I excluded it from our currently check and we
can work to bring versions into line through PRs.

### DIFF
--- a/config/dependency-lint.js
+++ b/config/dependency-lint.js
@@ -1,0 +1,8 @@
+/* eslint-env node */
+'use strict';
+
+module.exports = {
+  allowedVersions: {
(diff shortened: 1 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 2314bb0cb95547fa707fa18bc5eef4c470258680

https://deploy-preview-2711--ilios-frontend.netlify.com

PR #2717, opened 2017-03-06
Fill some sample charts with data

jrjohnson:
Lets use some actual data from courses and some actual charts to see
what this might look like.

These are not *good* looking charts, but they are filled with real data.


jrjohnson:
Deploy preview ready!

Built with commit d655a92d2eaffa631dcfd0093193b3b5bd259841

https://deploy-preview-2717--ilios-frontend.netlify.com

PR #2740, opened 2017-03-15
Provide attendanceRequired to the calendar

### LINKED ISSUE #2626: Apply schoolConfig session attribute options
Session overview can now be customized as of #2606.  We need to apply this setting in two areas:
1) Session creation / editing
2) Users calendar view

@saschaben I believe you said we should always hide the options on the administration of sessions once they are hidden by the school config, but if there has already been a value set for a session created in the past that we should show that value on the calendar even if the option is now turned off for the school.  Is that right?

jrjohnson:
This will allow it to show up on the single event page once the calendar is updated, but this can be merged before that is done without harm.

Fixes #2626

### DIFF
--- a/app/locales/en/translations.js
+++ b/app/locales/en/translations.js
@@ -29,6 +29,7 @@ export default {
     'associatedCourses': 'Associated Courses',
     'associatedGroups': 'Associated Groups',
     'associatedWith': 'Associated with',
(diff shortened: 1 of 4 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 3c097076a32373758e689caa95459e3cb1f856b0

https://deploy-preview-2740--ilios-frontend.netlify.com

PR #2750, opened 2017-03-16
Watch for configuration value changes

### LINKED ISSUE #2749: School Session Attributes don't refresh on save
Steps:
Visit school session attributes
Click Manage
make a change
the panel will close and the changes will not be reflected even though a refresh shows them as persisted.

jrjohnson:
We were watching the school configuration array which will catch
addition and deletion of configs, but doesn't notice if a value itself
changes.  We need to watch for that specifically.

Fixes #2749

### DIFF
--- a/app/components/school-session-attributes.js
+++ b/app/components/school-session-attributes.js
@@ -11,21 +11,22 @@ export default Component.extend({
   store: service(),
   schoolSessionAttributes,
   classNames: ['school-session-attributes'],
+  school: null,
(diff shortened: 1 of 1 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 9cfbb36eebc8ff69af5320a8f8c9e2ebdc10aadb

https://deploy-preview-2750--ilios-frontend.netlify.com

PR #2754, opened 2017-03-16
listen for changes to deletedUpdates in displayedUpdates comp prop.

### LINKED ISSUE #2726: resolving pending updates from Campus directory leaves phantom records in list
picking any resolution ("update ilios", "sync exception" etc) is applied successfully, however the resolved user record is not removed from the list of pending updates.

selection_152

stopfstedt:
fixes #2726

### DIFF
--- a/app/controllers/pending-user-updates.js
+++ b/app/controllers/pending-user-updates.js
@@ -49,7 +49,7 @@ export default Controller.extend({
     });
   }),
 
-  displayedUpdates: computed('allUpdates.@each.user.fullName', 'filter', 'offset', 'limit', 'deletedUpdates.[]',  function(){
+  displayedUpdates: computed('allUpdates.@each.user', 'filter', 'offset', 'limit', 'deletedUpdates.[]', function(){
     const limit = this.get('limit');
     const offset = this.get('offset');
     const end = limit + offset;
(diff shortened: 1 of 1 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 7a9a8f6f11bf99ca08701dcb991196683431222f

https://deploy-preview-2754--ilios-frontend.netlify.com

app/controllers/pending-user-updates.js
```diff
@@ -49,7 +49,7 @@ export default Controller.extend({
     });
   }),
 
-  displayedUpdates: computed('allUpdates.@each.user.fullName', 'filter', 'offset', 'limit', function(){
+  displayedUpdates: computed('allUpdates.@each.user.fullName', 'filter', 'offset', 'limit', 'deletedUpdates.[]',  function(){
```
jrjohnson: `allUpdates.@each.user.fullName` is also invalid and may be partially to blame for this issue.  `@each` doesn't work on nested properties.  Should be changed to `allUpdates.@each.user` and then see if we need the deletedUpdates watcher.

stopfstedt: corrected. that didn't do anything. listening for changes to `deletedUpdates` array is still they key to fixing this issue.

PR #2760, opened 2017-03-17
Hide school session attributes until they are fulfilled

### LINKED ISSUE #2759: Flicker of no in school session attributes
The school session attributes initialize to null before the promise resolves this means they flicker no when the page loads even if they should be yes.

jrjohnson:
If we don't do this they flicker as false until the promises render.

Fixes #2759

### DIFF
--- a/app/components/school-session-attributes.js
+++ b/app/components/school-session-attributes.js
@@ -46,8 +46,10 @@ export default Component.extend({
         toSave.pushObject(config);
       }
     }
-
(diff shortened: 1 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 41d6919a05a322abe149e31d5eab19563fd2744d

https://deploy-preview-2760--ilios-frontend.netlify.com

PR #2772, opened 2017-03-21
User model cleanup

stopfstedt:
just some cleanup, to prepare this model for the inevitable switching of Promises implementation.

### DIFF
--- a/app/models/user.js
+++ b/app/models/user.js
@@ -80,17 +80,15 @@ export default Model.extend({
     inverse: 'directors'
   }),
 
-  isStudent: computed('roles', {
-    get() {
-      const isStudent = this.get('roles').then((roles) => {
-        return !!roles.find((role) => role.get('title') === 'Student');
-      });
+  isStudent: computed('roles', function() {
+    const isStudent = this.get('roles').then((roles) => {
+      return !!roles.find((role) => role.get('title') === 'Student');
+    });
 
-      return ProxyContent.create({
-        promise: isStudent
-      });
-    }
-  }).readOnly(),
+    return ProxyContent.create({
+      promise: isStudent
+    });
+  }),
 
   cohorts: hasMany('cohort', {
     async: true,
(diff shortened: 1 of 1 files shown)


jrjohnson:
Deploy preview ready!

Built with commit d3b2c7768d9fd9efbbd8b33ea6368a62379fa941

https://deploy-preview-2772--ilios-frontend.netlify.com

app/models/user.js
```diff
@@ -1,106 +1,105 @@
...
-  examined:  DS.attr('boolean'),
-  userSyncIgnore:  DS.attr('boolean'),
-  icsFeedKey:  DS.attr('string'),
-  root: DS.attr('boolean'),
-  reminders: DS.hasMany('user-made-reminder', {async: true}),
-  reports: DS.hasMany('report', {async: true}),
-  school: DS.belongsTo('school', {async: true}),
-  authentication: DS.belongsTo('authentication', {async: true}),
-  directedCourses: DS.hasMany('course', {
+const ProxyContent = Object.extend(PromiseProxyMixin);
```
jrjohnson: Can we get rid of this and just use a returned Promise where it is used?

app/models/user.js
```diff
@@ -109,34 +108,32 @@ var User = DS.Model.extend({
...
         });
-        RSVP.all(promises).then(schools => {
+        all(promises).then(schools => {
           schools.pushObject(primarySchool);
-          defer.resolve(schools.uniq());
+          deferred.resolve(schools.uniq());
         });
       });
     });
     return PromiseArray.create({
```
jrjohnson: PromiseArray is a private API of ember-data - while cleaning this up I think switching to returning a plain Promise is probably a good idea.

stopfstedt:
@jrjohnson thanks for the review. all of the above will be addressed in a second PR. i want to keep the change set small on this one, and just focus on cosmetics now before moving onto actual re-implementations.

PR #2775, opened 2017-03-22
Improved Donut Chart

jrjohnson:
With working labels, a tooltip which is out of band, but present and a really dumb transition this component is becoming more like what we want it to be.


jrjohnson:
Deploy preview ready!

Built with commit cbd250929859dcfc2d70218c6dccf13992b24ab8

https://deploy-preview-2775--ilios-frontend.netlify.com

PR #2815, opened 2017-03-29
Improve the header on calendar filters

### LINKED ISSUE #855: Make Filter Type Header More Obvious
Please review the screen shot I am pasting in here for ideas.  This mock that I put in here only shows the "Session Types" filter.  The others would need a similar or standardized treatment.  Maybe even a background shading on the first row only to differentiate it as a header rather than data.  

sesstypefilter

jrjohnson:
Increased font weight and spacing to set the headers off from the
content.

Fixes #855

### DIFF
--- a/app/styles/components/_dashboard.scss
+++ b/app/styles/components/_dashboard.scss
@@ -57,6 +57,11 @@ $updated-color: #945c58;
       overflow-y: scroll;
       padding-bottom: 1em;
 
+      h5 {
+        font-weight: bold;
+        margin: 0 0 .5em;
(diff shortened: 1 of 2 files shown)


jrjohnson:
@dartajax how does this look?  I didn't change the color - but I think this still sets the header on lists off nicely.

jrjohnson:
Deploy preview ready!

Built with commit 60bd336f9b525f95a95f242a31a2d54bbaa7c762

https://deploy-preview-2815--ilios-frontend.netlify.com

Issue #2713, opened 2017-03-02
Delete Learner Group Fails When Sub Groups with Members Exist


dartajax:
See below ...

![image](https://cloud.githubusercontent.com/assets/7493331/23529363/2a12cdf2-ff52-11e6-94d0-49c1f51d7dd8.png)

After clicking as shown above ...

![image](https://cloud.githubusercontent.com/assets/7493331/23529483/aab7cc50-ff52-11e6-8c6b-dc033e96f6e2.png)

And then ...

Failure.

![image](https://cloud.githubusercontent.com/assets/7493331/23529538/dd052914-ff52-11e6-846e-346f77851680.png)

stopfstedt:
this behaviour is correct. we "just" need to check for groups with populated subgroups, and then not show the trashbin/delete icon.

stopfstedt:
alternatively, we could lift this restriction and cascade deletions down to learnergroup/learner associations.
@saschaben please advise. thanks!

stopfstedt:
discussed this with @saschaben - i'll try to resolve this first by checking for populated (sub)-groups in the frontend, putting spinners into place on each list item etc.
if this doesn't hold up due to performance reasons we'll have to consider alternatives.

jrjohnson:
If we can delete groups with learners why would we not allow a cascading delete of groups with subgroups with learners?  Or should we be preventing both?

PR #2817, opened 2017-03-29
dropped obsolete user count from learner group deletion message.

### LINKED ISSUE #2714: Member Count Is Incorrect (Doubled) on Learner Group Deletion
image

stopfstedt:
selection_191

since we locked down deletion of populated learner groups, it is not necessary to list the number of users anymore. 

refs #2713.
fixes #2714


jrjohnson:
Deploy preview ready!

Built with commit 8178d6ceb984359f36aa07d567ce81abdb312cb7

https://deploy-preview-2817--ilios-frontend.netlify.com

dartajax:
To be honest with you (which I generally am), the part about sub groups should be removed as well since no groups with sub groups can be deleted either.  Or ... we can merge this one and I will put in a new ticket with that as the content.  Whaddya think?

PR #2834, opened 2017-04-02
Update our deployments to be tag based

jrjohnson:
This switches our strategy to only deploy git tags in travis and to use
the tag name as the deployment key.  We also activate each tag
automatically instead of needed to do that step manually.


jrjohnson:
Deploy preview ready!

Built with commit 70c178beb80b2bc3adf4fba4a6ca2791f9b45657

https://deploy-preview-2834--ilios-frontend.netlify.com

PR #2870, opened 2017-04-15
Use Yarn for package management

jrjohnson:
Yarn is deterministic so we will always get the same set of packages
when running an install.  It is also cross compatible with NPM so we
can make this transition gradually.

At this point I have setup travis to NOT use the lockfile.  This will
ensure that our automatic greenkeeper updates are being tested
correctly.


jrjohnson:
Deploy preview ready!

Built with commit be10d566c5ac6153af7ede1b9165b55383c37daf

https://deploy-preview-2870--ilios-frontend.netlify.com

PR #2876, opened 2017-04-18
Fix overflow issue for tooltips

gabycampagna:
Add "auto scroll" to tooltips when content exceeds max-height of 150px. Update parent selector for .bar:hover to .ilios-chart. Change font-weight value to "500" for tooltips titles and add bottom-border to separate titles from tooltip content.

### DIFF
--- a/app/styles/newcomponents/chart-bar.scss
+++ b/app/styles/newcomponents/chart-bar.scss
@@ -1,4 +1,4 @@
-.ilios-chart {
+.chart-bar {
 
   .bar:hover {
     fill: $scheduled-brown;
--- a/app/styles/newcomponents/chart-bar.scss
+++ b/app/styles/newcomponents/chart-bar.scss
@@ -1,4 +1,4 @@
-.chart-container {
+.ilios-chart {
 
   .bar:hover {
     fill: $scheduled-brown;
--- a/app/styles/newcomponents/ilios-chart-tooltip.scss
+++ b/app/styles/newcomponents/ilios-chart-tooltip.scss
@@ -9,16 +9,18 @@
   margin: 0;
   max-height: 150px;
   opacity: .9;
+  overflow: auto;
   padding: 10px;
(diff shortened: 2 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 5cc8fe5fa85b605cccb9a9160db95b3d1e4dd4d6

https://deploy-preview-2876--ilios-frontend.netlify.com

app/styles/newcomponents/chart-bar.scss
```diff
@@ -1,4 +1,4 @@
-.chart-container {
+.ilios-chart {
```
jrjohnson: Since this is in the chart-bar file I think it should probably by `.chart-bar {`

app/styles/newcomponents/ilios-chart-tooltip.scss
```diff
@@ -9,16 +9,18 @@
   margin: 0;
   max-height: 150px;
```
jrjohnson: What if we just got rid of max height?  I'm guessing scrolling inside the tooltip is going to be a pain.

gabycampagna: We can remove it max-height. The height of the tooltip will scale to fit all the content.

jrjohnson: That's what I'm thinking.  Does it break everything?

app/styles/newcomponents/chart-bar.scss
gabycampagna: That makes sense. I see how it works in the visualizer-course-objectives.hbs

app/styles/newcomponents/ilios-chart-tooltip.scss
gabycampagna: Nothing breaks.  Just some tooltips take up more real estate.

PR #2879, opened 2017-04-19
Truncate long group names before saving

### LINKED ISSUE #2869: Fail Better When Generated Sub Group Names Are Too Long
Here is what happens ...

Attempting to add (using the auto generate of sub groups which adds the final two or more characters to the newly generated group names) a title longer than 60 characters results in the following error in the logs ...

```[2017-04-14 22:16:22] request.ERROR: Uncaught PHP Exception Symfony\Component\HttpKernel\Exception\HttpException: "Object(Ilios\CoreBundle\Entity\LearnerGroup).title:     This value is too long. It should have 60 characters or less. (code d94b19cc-114f-4f44-9cc4-4138e80a87b9) " at /var/www/ilios/src/Ilios/ApiBundle/Controller/ApiController.php line 473 {"exception":"[object] (Symfony\\Component\\HttpKernel\\Exception\\HttpException(code: 0): Object(Ilios\\CoreBundle\\Entity\\LearnerGroup).title:\n    This value is too long. It should have 60 characters or less. (code d94b19cc-114f-4f44-9cc4-4138e80a87b9)\n at /var/www/ilios/src/Ilios/ApiBundle/Controller/ApiController.php:473)"} []```

* Steps to reproduce 
Pick a group with a 59 char name like this one ...

`S17_CP113 Pharm Prac_PharmD Class of 2020 Conference Groups`

image

After the attempt at `Save` ...

image

image

After clicking as shown above, do NOT refresh the page and try to add a Learner to this non-existent Sub Group.

image

Same error ...

image

jrjohnson:
If a learner group is approaching the database limit we need to truncate
the title before attempting to append a counter to it when generating
subgroups.  This prevents issues on the API server when we attempt to
save groups with bad titles.

Fixes #2869

### DIFF
--- a/app/components/learnergroup-subgroup-list.js
+++ b/app/components/learnergroup-subgroup-list.js
@@ -8,6 +8,7 @@ const { Promise } = RSVP;
 
 export default Component.extend({
   store: service(),
+  flashMessages: service(),
   parentGroup: null,
   classNames: ['learnergroup-subgroup-list'],
   tagName: 'section',
@@ -52,9 +53,11 @@ export default Component.extend({
           parentGroup.get('cohort').then((cohort) => {
             let groups = [];
             const padBy = countDigits(offset + parseInt(num, 10));
+            const parentTitle = parentGroup.get('title');
+            const shortenedParentTitle = parentTitle.substring(0, 60 - 1 - padBy);
             for (let i = 0; i < num; i++) {
               let newGroup = store.createRecord('learner-group', {
-                title: parentGroup.get('title') + ' ' + pad(offset + i, padBy),
+                title: shortenedParentTitle + ' ' + pad(offset + i, padBy),
                 parent: parentGroup,
                 cohort
               });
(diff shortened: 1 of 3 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 16f4e754b799504ebdbbf55625eca9e14c417b52

https://deploy-preview-2879--ilios-frontend.netlify.com

jrjohnson:
@saschaben please checkout this solution which truncates the end of a long group name with no warning to the user. Is this OK, or do we need to provide some kind of message that the group names will be different?

saschaben:
Since this addresses the edge case of auto-generating names of subgroups for a learner group with a name already at the maximum, I am ok with this solution -- but it must be noted in the user guide. Please have @dartajax add a note to this effect to document.

dartajax:
Oh man, now I duped it. I just put #19 in user-guide. I will close one of the dupes.

dartajax:
This occurs when you edit a Learner Group name that is already at its maximum as well. The field allows more information to be entered. Need to make the field only accept 60 chars, no?

jrjohnson:
@dartajax please create a separate issue for editing.  It is related to this - but the solution is different enough that I want to address it separately.

PR #2880, opened 2017-04-19
Focus when editing

### LINKED ISSUE #446: Click to Edit Field Activates Field and sets Focus on Selected Field

jrjohnson:
When we open an editable field focus is places in the element we are now
editing. This makes it much easier to start working right away without
an extra click.

Fixes #446

### DIFF
--- a/app/components/editable-field.js
+++ b/app/components/editable-field.js
@@ -31,4 +31,12 @@ export default Component.extend({
     yield this.get('close')();
     this.set('isEditing', false);
   }).drop(),
+
+  edit: task(function * () {
+    this.set('isEditing', true);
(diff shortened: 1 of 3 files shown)


jrjohnson:
Deploy preview ready!

Built with commit e91b56c2ebc40d20a1cb17191e6910f8cabbdc3d

https://deploy-preview-2880--ilios-frontend.netlify.com

PR #2882, opened 2017-04-19
Highlight back to session link

### LINKED ISSUE #593: Return to Session List Link Should be Available from Review Missing Items
Related to #409 

It is difficult to figure out how to navigate back to the Session List from the Review Missing Items area for a Session.  I will put some screen shots in here tomorrow to illustrate.

What I noticed ...
- Select a Course
- Select a Session 
- Select "Review x Missing Items"

If the user decides that he or she wishes to return all the way back to the full session list for the course, there is no way to do so from here.  The user must ...
- Avoid clicking what looks like a link but actually will end up in the editing mode of the Session Title.  I think it looks like a link and has the same font as the one below it.
- Click the lower link (which needs formatting -- it is not indented the same as the edit link -- too far outdented) to go back to the Session itself.
- Then click "Back to Sessions List".

It doesn't seem very intuitive.

jrjohnson:
This link is a bit difficult to pick out on the page so I've highlighted
it with some spacing, an icon, and some font-weight.

Fixes #593

### DIFF
--- a/app/styles/components/_detailview.scss
+++ b/app/styles/components/_detailview.scss
@@ -258,3 +258,12 @@ tbody .confirm-removal {
     padding-right: .2em;
   }
 }
+
+.session-publicationcheck-detail-view {
+  .back-to-session {
+    border-top: 1px solid $ilios-orange;
+    font-weight: bold;
+    margin: 1em 1.5em;
+    padding-top: 1.5em;
+  }
+}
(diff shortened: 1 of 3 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 7c63f6ebe9ae9ca93170c01499bc41edc16f7ec9

https://deploy-preview-2882--ilios-frontend.netlify.com

dartajax:
How about adding a dividing line like shown below ...
image

dartajax:
It's the orange line below `Associated Groups:  None` - it can be dashed like the lower one.

dartajax:
That border looks exactly like the one I suggested. LGMT 👍

PR #2883, opened 2017-04-19
Validate learner group title changes

### LINKED ISSUE #2881: Do Not Allow A Learner Group Greater Than 60 Chars to Be Entered in Edit Mode
Whilst we do capture the fact that upon initial data entry, a Learner Group cannot be entered with greater than 60 characters, this is not true when the Learner Group name is edited.

See below.

* Try and enter a Learner Group with more than 60 characters -- Proper Validation

image

* Edit a Learner Group and make the title exactly 60 characters in length. -- Not a Problem but at the maximum length

image

* Edit the Learner Group created in the previous step and add even more characters and try to save.

image

image

related to #2879

jrjohnson:
We were allowing edits to break all the rules for the learner group
title - which is no good, obviously.

Fixes #2881

### DIFF
--- a/app/components/learnergroup-header.js
+++ b/app/components/learnergroup-header.js
@@ -1,8 +1,21 @@
 import Ember from 'ember';
+import { task } from 'ember-concurrency';
+import { validator, buildValidations } from 'ember-cp-validations';
+import ValidationErrorDisplay from 'ilios/mixins/validation-error-display';
 
 const { Component } = Ember;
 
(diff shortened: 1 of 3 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 32b795dd0676669b41088f6ea7f58191e47e18f6

https://deploy-preview-2883--ilios-frontend.netlify.com

PR #2885, opened 2017-04-20
Handle initial active users for user-search better

### LINKED ISSUE #2770: Adding Program Directors -- Once Selected Should Be Disabled
image

jrjohnson:
When and empty array was sent to the user-search currentlyActiveUsers
property we were replacing it with a new empty array. This prevented
updates to the array from being tracked. Instead check for null
explicitly and user empty arrays as is.

Fixes #2770

### DIFF
--- a/app/components/user-search.js
+++ b/app/components/user-search.js
@@ -57,7 +57,7 @@ export default Component.extend({
       };
     }
     let users = yield this.get('store').query('user', query);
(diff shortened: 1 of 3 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 3eb8376fb7014cfc5274f8f48f9bc5aeb1b7e9b2

https://deploy-preview-2885--ilios-frontend.netlify.com

PR #2886, opened 2017-04-20
Reset learner group list after saving changes

### LINKED ISSUE #2627: Remove Learner Count Does Not Refresh / Get Re-Calculated Properly - Learner Groups
Try this ...

Select a Learner Group

image

Select two Learners to remove from the selected Learner Group 

Click as shown below (correct button captions at this point)

image

After the move event is processed, the captions remain (now incorrect)

image

If you want to then move one person, the caption reads 3 as shown below.

Only that one Student was selected, not three - needs refreshment (as do I)

image

jrjohnson:
We were not zeroing out the list of checked users so the count was
causing the buttons to be in a confusing and wrong state.

Fixes #2627

### DIFF
--- a/app/components/learnergroup-cohort-user-manager.js
+++ b/app/components/learnergroup-cohort-user-manager.js
@@ -56,6 +56,7 @@ export default Component.extend({
     yield timeout(10);
     yield this.get('addUsersToGroup')(users);
     this.get('usersBeingMoved').removeObjects(users);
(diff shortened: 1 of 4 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 0adc3ac4c433e96f23f2519cad7f430f6c459590

https://deploy-preview-2886--ilios-frontend.netlify.com

PR #2888, opened 2017-04-20
Improve reports without titles

### LINKED ISSUE #1223: Suggested Improvement To Automated Report Names Generated by My Reports
Currently when a report is saved in MyReports on the Dashboard, the automated title appears as follows.

One. Create a report specifying "All Courses for Instructor [David Leonard].  Save the report.

image

Two. The report is saved as follows.  This is the same as it was in 2.4.x so this is not a bug at all; just a suggested readability improvement.

image

The report title would be better as "All Courses ..." rather than "All course ..." so I am suggesting to capitalize the noun as well as pluralizing it since that is the report that is being generated.

jrjohnson:
Instead of using the report subject directly in the generated title we
can translate it into a more pleasing capitalized and pluralized form.

Fixes #1223

### DIFF
--- a/app/models/report.js
+++ b/app/models/report.js
@@ -28,7 +28,23 @@ export default DS.Model.extend({
         resolve(title);
         return;
       }
+      const subjectTranslations = {
+        'course': 'general.courses',
+        'session': 'general.sessions',
+        'program': 'general.programs',
+        'program year': 'general.programYears',
+        'instructor': 'general.instructors',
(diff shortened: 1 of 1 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 056102865860d96e409274c751f73b45d28a7dbc

https://deploy-preview-2888--ilios-frontend.netlify.com

PR #2920, opened 2017-04-27
correctly pass login error message keys. provide fallback by passing …

### LINKED ISSUE #2207: Validate login errors before displaying them
We are currently trusting the server to return a good translatable error messages, but it is also possible to get the attached:
screen shot 2016-10-09 at 9 10 20 pm

stopfstedt:
…raw response message.

fixes #2207 

the fallback is necessary, b/c Ember i18n does not provide a means to check if a given translation key is backed by a translation.

the recommended way for dealing with missing translations is documented [here](https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Missing-Translations) - overriding the callback won't work for us.

### DIFF
--- a/app/authenticators/ilios-jwt.js
+++ b/app/authenticators/ilios-jwt.js
@@ -43,8 +43,11 @@ export default JwtTokenAuthenticator.extend({
           response = Ember.merge(response, tokenExpireData);
           resolve(this.getResponseData(response));
         }, e => {
-          const errors = e.errors || [];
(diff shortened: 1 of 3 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 25aea9498b131d713ceb9713a1ffafb8ebefc031

https://deploy-preview-2920--ilios-frontend.netlify.com

PR #2929, opened 2017-04-30
Remove animation from donut chart for icon entry points

gabycampagna:
Remove animation from donut chart for icon entry points

### DIFF
--- a/app/components/chart-donut.js
+++ b/app/components/chart-donut.js
@@ -33,7 +33,7 @@ export default Component.extend({
     const donutWidth = width * .2;
     const color = scaleOrdinal(schemeCategory10);
     const isIcon = get(this, 'icon');
-    
+
     let t = transition().duration(500).ease(easeLinear);
 
     let createArc = arc().innerRadius(radius - donutWidth).outerRadius(radius);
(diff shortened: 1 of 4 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 13189ae93c89715ffc376b3d9dbbe306b8ebadfa

https://deploy-preview-2929--ilios-frontend.netlify.com

jrjohnson:
I'm not sure if you already tried this - is it possible to just have both animation happen at the same time?  Does that remove the flicker in the icon?  

If not then I think the chart should detect if it is an icon based on the height / width passed in. My concern is that passing the `icon` value all up and down the chart components is probably going to be annoying in the long run.  It would make more sense for the donut chart to decide that at anything less than a size 100x100 for example) some animations don't run.  And maybe the pie chart makes a different decision at a different break point.

gabycampagna:
I just tried the animation running at the same time and the flicker is still there. The labels are also displaying and is cluttering up the icon.

**With labels rendering:**
screen shot 2017-04-30 at 12 50 42 pm

No labels: 
screen shot 2017-04-30 at 12 49 43 pm

I understand what you are saying about passing the ```icon``` value on all components in the long run. 

I'll see if I can figure out how the solution you have requested.

app/components/chart-donut.js
```diff
@@ -20,6 +20,7 @@ export default Component.extend({
   data: null,
   width: null,
   height: null,
+  icon: false,
```
jrjohnson: Old code. Should be removed

gabycampagna: Awe! Ok!

PR #2935, opened 2017-05-02
Don't allow shadow variables

jrjohnson:
This is mainly needed for an update to babel6 which doesn't play well with us shadowing the global `Object`.  But I cleaned up all of our other shadows as well.


jrjohnson:
Deploy preview ready!

Built with commit dd4f595cf8c1b9690f17fd03f4b7a05810da8f8e

https://deploy-preview-2935--ilios-frontend.netlify.com

PR #2941, opened 2017-05-03
Include all cohorts that overlap with date, not just by start date 

gboushey:
@2909

@dartajax - would you let me know if this is showing the cohorts that should be in the drop down menu per issue 2909?

Can't merge yet because the field doesn't populate properly by default - you have to make a selection from the date list or the cohort list won't refresh.


jrjohnson:
Deploy preview ready!

Built with commit 30433216eced21c467ced180e01efaffff89c57d

https://deploy-preview-2941--ilios-frontend.netlify.com

dartajax:
I will take a look.

dartajax:
There is one slight bug that I believe I have found ...

Review the screen shots below. Following the exact steps below, the available cohorts do not get displayed initially. You end up having to select a different year and then return to the correct year to populate the cohort list.

image

image

image

image

image

image

gboushey:
@dartajax thanks for the review.  I'll have to figure out how to get the right data loaded by default, rather than by switching the date.  Is the data accurate once you make that change?  Ie., once you get the data to display through the steps you describe, are you seeing the cohorts that should show up?  That'll tell me if the basic filter is accurate, and I can move on to figuring out why the initial rendering isn't happening properly.

dartajax:
The answer to your question is "yes". The cohorts appear correct once everything is loaded up. If you choose a different path ...

* Show Calendar
* Show All Events
* Show Filters 
* Switch Filter 
* Select Correct Year

... it works out fine. It's only when those last two steps are reversed that the Cohorts don't show up.

jrjohnson:
Superseded by ##2950

PR #2942, opened 2017-05-03
Add Session Type Management for schools

### LINKED ISSUE #2910: Make the selection/customization of calendar colors a school configurable option
Schools should be allowed to select and manage their own palette of session type colors. This will both provide the ability to address legacy issues without changing code, and will also allow for the (slow) migration away from the use of specific colors to deliver content information, and toward a more accessible and ADA compliant approach to delivery.
colorpicker
colorpicker2
colorpicker3

jrjohnson:
Adds a 'Session Type' section to each school manager which lists the session types and their properties including the new session type calendar color property. Each on can be edited here as well. To limit the scope of this work I left off creating a new type and deleting types that are not in use.

Fixes #2910
Requires https://github.com/ilios/ilios/pull/1823

When a sessionType has sessions attached to it only the color is editable.


jrjohnson:
Deploy preview ready!

Built with commit 7b3a0d28a7a426b6801e1494f2ab64efaedadcd9

https://deploy-preview-2942--ilios-frontend.netlify.com

PR #2948, opened 2017-05-05
Hackey fix for flaky tests

jrjohnson:
When pending-updates-summary-test is run before user-profile-schools it
breaks the  clicking write selects read test. I wasn't able to come up
with a good solution, but using our 'is-in' helper and a set of computed
properties on the component resolves it for some reason.

The easiest way to see this failure is to visit http://localhost:4200/tests?testId=c2c9b934&testId=f41a6e34 which will run these two tests together.

### DIFF
--- a/app/components/user-profile-schools.js
+++ b/app/components/user-profile-schools.js
@@ -100,71 +100,72 @@ export default Component.extend({
     });
   }),
 
-  secondarySchools: computed('user.school', 'currentUser.model.school', function(){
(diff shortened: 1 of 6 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 5245c1a2194ff752aa65bf23a3defbd08f5b6b59

https://deploy-preview-2948--ilios-frontend.netlify.com

PR #2974, opened 2017-05-12
Remove abilities as a permissions checking system

jrjohnson:
We were only using this in a single remaining place which was converted
to using permissions in the currentUser service.

Also removes abilities observers from currentUser service
These observers and values were used in the abilities system which has
now been removed.  The only thing still using them was reporting which
has been updated to calculate its own abilities.


jrjohnson:
Deploy preview ready!

Built with commit ec4bfba187d287077bb4d6f504464dd70f9f2553

https://deploy-preview-2974--ilios-frontend.netlify.com

PR #2988, opened 2017-05-17
Use frozen lockfile with yarn on travis

jrjohnson:
This will fail if the lockfile is out of date, which is better than
`yarn check` which doesn't seem to do anything when package.json is
updated and the yarn.lock file isn't.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_install:
   - npm install -g bower
 
 install:
-  - yarn install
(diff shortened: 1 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 54655120422ce4cc96b60ebd4aa40b61675884de

https://deploy-preview-2988--ilios-frontend.netlify.com

PR #2992, opened 2017-05-18
Remove pluralize dependency

jrjohnson:
We were using this in the early days when we needed to use express to
serve the mock API, but it is no longer needed.


jrjohnson:
Deploy preview ready!

Built with commit 8521aaa9327f288f806cc8156aa5557d7e008a67

https://deploy-preview-2992--ilios-frontend.netlify.com

PR #3015, opened 2017-06-05
Resize SVG element for donut chart model using JS and CSS

gabycampagna:
screen-shot-2017-06-05-at-7 59 43-am
screen-shot-2017-06-05-at-8 02 11-am

### DIFF
--- a/app/components/chart-donut.js
+++ b/app/components/chart-donut.js
@@ -43,6 +43,14 @@ export default Component.extend({
       return;
     }
 
+    if (!isIcon) {
+      svg.classed("svg-container", true)
+      .attr("preserveAspectRatio", "xMinYMin meet")
+      .attr("viewBox","0 0 " + width + " " + height)
+      //class to make it responsive
+      .classed("svg-content-responsive", true);
+    }
+
     let chart = svg.append('g').attr('class', 'pie').attr('transform', 'translate(' + (width / 2) +  ',' + (height / 2) + ')');
     let path = chart.selectAll('path').data(createPie(dataOrArray)).enter()
       .append('g').attr('class', 'slice')
--- a/app/components/chart-donut.js
+++ b/app/components/chart-donut.js
@@ -44,11 +44,9 @@ export default Component.extend({
     }
 
     if (!isIcon) {
-      svg.classed("svg-container", true)
+      svg.classed("svg-container", true) // responsive svg container
       .attr("preserveAspectRatio", "xMinYMin meet")
-      .attr("viewBox","0 0 " + width + " " + height)
-      //class to make it responsive
-      .classed("svg-content-responsive", true);
+      .attr("viewBox","0 0 " + width + " " + height);
     }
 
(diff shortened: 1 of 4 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 3389e02c6438a7ee188500994d112a70325b3cdc

https://deploy-preview-3015--ilios-frontend.netlify.com

app/components/chart-donut.js
```diff
@@ -43,6 +43,14 @@ export default Component.extend({
       return;
     }
 
+    if (!isIcon) {
+      svg.classed("svg-container", true)
```
jrjohnson: As I understand this code the class `.svg-container` gets added here, is that right?

app/components/chart-donut.js
```diff
@@ -43,6 +43,14 @@ export default Component.extend({
...
+      .attr("preserveAspectRatio", "xMinYMin meet")
+      .attr("viewBox","0 0 " + width + " " + height)
+      //class to make it responsive
+      .classed("svg-content-responsive", true);
```
jrjohnson: Do we need both of these classes? Shouldn't we just make all SVGs responsive? Or put another way - is there an `.svg-container` that isn't also going to be responsive?

app/styles/newcomponents/ilios-chart.scss
```diff
@@ -3,5 +3,37 @@
...
+    }
+  }
+
+  .svg-container {
+    overflow: visible;
+    padding-bottom: 100%; // aspect ratio
+    vertical-align: top;
+  }
+
+  @include media($small-screen) {
```
jrjohnson: These media queries feel a bit inverted to me can this be written as 
```
.svg-container {
  overflow: visible;
  padding-bottom: 100%; // aspect ratio
  vertical-align: top;

  @include media($small-screen) {
    padding-bottom: 26%;
    width: 50%;
  }
}
```

app/components/chart-donut.js
gabycampagna: Yes, but.. It should actually be added to a parent div or other container.  Wasn't sure if that is the route we want to go, so I added it to the svg element.  Please let me know you thoughts about that.

app/components/chart-donut.js
gabycampagna: We could or "should" make all SVG responsive, with the exception of the Icons of course.  We could add a parent container to the SVG as I mentioned in the comment above this one.

app/components/chart-donut.js
jrjohnson: Ah ok. Yes. Maybe we can apply all of this to the ilios-chart component. I'm thinking:
1) Get rid of width / height properties on all chart components
2) Handle size through CSS completely
3) Keep isIcon and just apply that as a class on the ilios-chart component.

Will that work with the D3 svg? I'm basically just wishing and making stuff up so let me know if that wont work at all.

gabycampagna: I will give it a try and let you know!  But I think the currently solution we have is the simplest one. I will think this through and get back to you.

jrjohnson:
@dartajax please review animations throughout the application to ensure they aren't wonky. Then this can be merged.

PR #3031, opened 2017-06-15
2972 session type aamc methods

gboushey:


jrjohnson:
@gboushey I think this could benefit from squashing down into a single commit with an interactive rebase. That might deal with some of the errors as well as clean up the commit history which has a lot of merges in it.

jrjohnson:
Deploy preview ready!

Built with commit a346f5dd60d66dd30ad455fe02f0071faf1077b2

https://deploy-preview-3031--ilios-frontend.netlify.com

saschaben:
uncertain if this is a typo in the db or an issue in the text display rendering, but the AAMC Method of `Discussion, Small Group (<12)` is being rendered as `Discussion, Small Group (?12)`. If this is a rendering issue, let's be sure to fix. If it is a data issue, let's also be sure to fix.

saschaben:
Edge case: if an existing session type has its mapping modified from an Instructional Method to an Assessment Method, or vice versa, the assigned value for `isAssessment` does not change. We should consider disallowing this type of alteration. One option: once a session type is saved, filter the selectable aamcMethods to ONLY allow selection of values which match the current value of `isAssessment`.

If we do not lock this down, while an edge case, it leaves open to user error the real corruption of data.

jrjohnson:
Yes, when assessment is checked only AAMC methods with an id like `AM001` (the IM being assessment method). Otherwise only those with `IMxxx` should be choosable. 
When the assessment slider is clicked we should also reset the AAMC method to null so it isn't possible to save bad data.

saschaben:
Just to clarify: `AMxxx` is "Assessment Method"; `IMxxx` is "Instructional Method. and `IRxxx` is Instructional Resource, for what it's worth....

PR #3033, opened 2017-06-16
Aamc methods

gboushey:
#3031

PR #3047, opened 2017-06-22
Aamc methods

stopfstedt:
supersedes #3033

### DIFF
--- a/app/components/school-session-type-form.js
+++ b/app/components/school-session-type-form.js
@@ -3,7 +3,7 @@ import ValidationErrorDisplay from 'ilios/mixins/validation-error-display';
 import { validator, buildValidations } from 'ember-cp-validations';
 import { task } from 'ember-concurrency';
 
-const { Component, inject } = Ember;
+const { computed, Component, inject, isPresent } = Ember;
 const { service } = inject;
 
 const Validations = buildValidations({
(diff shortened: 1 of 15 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 95c0efe8c8a7770eea3cca2c4c94b93a30c13315

https://deploy-preview-3047--ilios-frontend.netlify.com

saschaben:
still needs to filter AAMC method options on edit of existing types, so that an assessment method cannot map to an instructional, and vice versa.

jrjohnson:
Still awaiting tests - question for @saschaben I made the AAMCMethod not-required (can be blank). Is that right? This can be click tested again now.

I also prevented aamcMethod from being changed if there are attached sessions.

app/components/school-session-type-form.js
```diff
@@ -19,22 +20,57 @@ const Validations = buildValidations({
...
 
 export default Component.extend(ValidationErrorDisplay, Validations, {
   store: service(),
   classNames: ['school-session-type-form'],
   title: null,
+  selectedAamcMethodId: null,
   calendarColor: null,
   assessment: null,
   assessmentOptionId: null,
+  aamcMethods: computed('assessment', async function(){
```
jrjohnson: Performance can be improved here by splitting this into an allAaamcMethods property and a filtered aamcMethids property.

saschaben:
`I made the AAMCMethod not-required (can be blank). Is that right?` THIS IS CORRECT.

`I also prevented aamcMethod from being changed if there are attached sessions.` THIS IS INCORRECT: A MAPPING MAY BE CHANGED, EVEN AFTER SESSIONS ARE ATTACHED.
>> The Rule: Once created&saved, a session type may only change type within its class -- either instructional or assessment. Once created&saved, an assessment type may not be changed from summative /formative selection. Once _mapped_, a session type _MAY_ its mapping changed, but only within its class (instructional/assessment).

Currently hitting a few errors on the form:

For an unassociated session type (no sessions in use), the following fields are shown as modifiable, but ALL throw error:
preexist2nomap

err1

we should lock down all form fields according to the rules described above. The assessment toggle should not be active after save. The AAMC Method select list should be constrained to the class selected, and active. If assessment, the formative/summative selector should be inactive after save.

preexist1

saschaben:
Looks good. Approved from this end. :+1:

PR #3051, opened 2017-06-23
Return to the warm embrace of NPM

jrjohnson:
Npm v5 is fast and cachy and deterministic. All the things that yarn is,
but better supported by the larger community. Like the Prodigal Son we
will be welcomed with open arms as we return to NPM.


jrjohnson:
Deploy preview ready!

Built with commit 168a5e9201493cddaa84a80512d67b8874820b5e

https://deploy-preview-3051--ilios-frontend.netlify.com

PR #3074, opened 2017-06-29
Bump our API version

jrjohnson:
No model changes are needed here, the new API has more data in the
userevents and schoolevents APIs which we don't need to change any code
to handle.

Handles API bump from https://github.com/ilios/ilios/pull/1836

### DIFF
--- a/config/api-version.js
+++ b/config/api-version.js
@@ -1,3 +1,3 @@
 /* eslint-env node */
 
-module.exports = 'v1.19';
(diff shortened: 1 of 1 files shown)


jrjohnson:
Deploy preview ready!

Built with commit c51d6364fb3138bc06f116b593c3b33ba1ba4fca

https://deploy-preview-3074--ilios-frontend.netlify.com

PR #3089, opened 2017-07-10
Provide bulk user sample file as a data-uri link

### LINKED ISSUE #3023: Sample User Upload link points to non-existent file
The SampleUserUpload.txt file, linked to from /users?addUsers=true&filter=example, does not contain a file.  This asset is included on the frontend, but it does not exist at the link, because it does not reside on the backend.  We probably need to add the SampleUserUpload.txt file to the backend, something like what we did in #1811, or we should discuss.
sampleuserupload

jrjohnson:
We've been having trouble getting this file from our CDN with a correct
link, instead we can just put the data into the link directly.

Fixes #3023

### DIFF
--- a/app/components/bulk-new-users.js
+++ b/app/components/bulk-new-users.js
@@ -113,7 +113,7 @@ export default Component.extend(NewUser, {
   getFileContents(file){
     this.set('fileUploadError', false);
     return new Promise(resolve => {
-      let allowedFileTypes = ['text/plain', 'text/csv'];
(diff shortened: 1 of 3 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 6c7416473bb38d4deb99f494f69651d514d93327

https://deploy-preview-3089--ilios-frontend.netlify.com

saschaben:
:+1:

PR #3105, opened 2017-07-24
Drop support for older browsers

jrjohnson:
By only targeting modern evergreen browsers in our build we can
significantly reduce the amount of extra code we generate to be
compatible with older browsers. This will make our application smaller
to download and faster to parse.

This list is dynamic so overtime the actual result of our build will change as newer browsers are released. This should mean we continue to take advantage of new features as they are released in our oldest supported browser.

### DIFF
--- a/config/targets.js
+++ b/config/targets.js
@@ -1,9 +1,13 @@
 /* eslint-env node */
+
+// Which browsers are returned can be found at http://browserl.ist/
 module.exports = {
   browsers: [
-    'ie 9',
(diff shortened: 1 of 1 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 790b051f536065f97da5699f9d9332391d4ec97a

https://deploy-preview-3105--ilios-frontend.netlify.com

jrjohnson:
@saschaben I think this is ready to merge. Do you want to do the honors?

PR #3110, opened 2017-07-26
Add status back into session search string

### LINKED ISSUE #3109: Filter Session By Status Does Not Work - Regression Bug
This used to work I believe.

image

The only reason the one below gets returned is because the word "Published" is in the Title of the Session.

image

jrjohnson:
Search string provided by the session model doesn't have status anymore
since we remove i18n from the PublishableModel mixin. This means we need
to construct this string ourselves so we can filter correctly.

Fixes #3109


jrjohnson:
Deploy preview ready!

Built with commit d41495b3db27a446847d905c8ee556ae80f6384a

https://deploy-preview-3110--ilios-frontend.netlify.com

PR #3111, opened 2017-07-26
Don't reload records when navigating back to school

jrjohnson:
We need to look inside the store first to see if we have already loaded
all of these records. Otherwise we will fire off network requests to
repeat the same work again.

This can be tested by navigating back and forth between courses in a session. When going back to a course no new network requests should be sent.


jrjohnson:
Deploy preview ready!

Built with commit a179daf61453a6016dd008c29ca26d18a1fc01dd

https://deploy-preview-3111--ilios-frontend.netlify.com

PR #3120, opened 2017-07-31
Remove reliance on the Ember object

jrjohnson:
Instead of destructuring everything off of Ember we can import the JS modules. This has big performance gain possibility as ember-cli gets better at shaking out unused parts of the framework so our users don't need to download them anymore.

WIP:
- [ ] ~Waiting on isTesting to get added (already did the mod, just need upstream update)~ we should merge without waiting for this per the ember core team
- [ ] ~Waiting on onerror to get added (a decision about where to put this hasn't been made yet)~ we should merge without waiting for this per the ember core team
- [x] Disallow importing Ember in an eslint rule


jrjohnson:
Deploy preview ready!

Built with commit d488cb4c5444ea3c514c95fb9bdf13c4e8d8226c

https://deploy-preview-3120--ilios-frontend.netlify.com

jrjohnson:
@stopfstedt I split this into three commits that all touch a bunch of files, but in the exact same way so hopefully it isn't too onerous to review.

jrjohnson:
@stopfstedt fixed, thanks for the catastrophe avoiding catch.

PR #3121, opened 2017-07-31
Convert all acceptance tests to use async/await syntax

jrjohnson:
This is much clearer than `andThen()` for waiting for asynchronous
events in a test.

WIP:
- [x] Need to figure out how to disallow `andThen` global in our tests. This global is set by the `env` `embertest: true` option and comes from https://github.com/sindresorhus/globals/blob/master/globals.json#L1239 but I'm not sure how to unset this. Maybe @Trott knows how to customize an `eslint` environment? Otherwise we can just declare all of those globals (except `andThen`) in our own config.

### DIFF
--- a/tests/.eslintrc.js
+++ b/tests/.eslintrc.js
@@ -1,7 +1,4 @@
 module.exports = {
-  env: {
-    embertest: true
-  },
   "globals": {
     "$": true,
     "getElementText": true,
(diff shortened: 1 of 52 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 59655b1b7683097baa1d4dc1a0bc78bd93aef14b

https://deploy-preview-3121--ilios-frontend.netlify.com

PR #3153, opened 2017-08-15
Refactor styles for publicationcheck

jrjohnson:
Both session and course check were using the inherited detail-view
styles which are on their way out. Now they are styles on their own.

To Test:
- [x] That publication check for course and session still look correct.


jrjohnson:
Deploy preview ready!

Built with commit 801e2e9c76298e2788dbd202d85f3583b31ad59b

https://deploy-preview-3153--ilios-frontend.netlify.com

Issue #3038, opened 2017-06-19
Week at Glance -- Provide A Method of Choosing Whether to Open or To Save a PDF File 


dartajax:
It would be helpful for students to be able to choose whether they wish to ...

* A. download a document (PDF's especially) or 
* B. open the document in a browser window immediately

We can start with PDF's since these are the most likely to be easily opened in a new browser window; although we would need to test with all browsers possible.

jrjohnson:
This should also apply to the materials list and the single event view.

dartajax:
This applies to ...

* Week at A Glance
* My Materials
* Event Detail

stopfstedt:
done and done.

PR #3162, opened 2017-08-16
Inline and Download links for PDF learning materials

stopfstedt:
refs #3038 

My Materials:
selection_370

Learning Materials manager:
selection_371

### DIFF
--- a/app/locales/en/translations.js
+++ b/app/locales/en/translations.js
@@ -93,6 +93,7 @@ export default {
     'confirmVocabularyRemoval': 'Are you sure you want to delete this vocabulary?',
     'contentAuthor': 'Content Author',
     'copiedSuccessfully': 'Copied Successfully',
+    'copyLink': 'Copy link',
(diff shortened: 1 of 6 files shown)


jrjohnson:
Deploy preview ready!

Built with commit c038ee84df47524b2545d8c57e4cccae4633e06e

https://deploy-preview-3162--ilios-frontend.netlify.com

saschaben:
this looks great. Let's get this in place everywhere!

dartajax:
any chance we could put a mouse-over caption to explain this ... 
image

dartajax:
or what is that supposed to do? copy the file to the clipboard?

dartajax:
image

dartajax:
we should do a mouse-over caption over the file name too to indicate to the user that it will open inline - rather than think they can interpret the URL at the bottom of the screen -- are these suggestions possible or out of scope?

saschaben:
@dartajax is right -- the copy icon should have a mouseover of "copy link"

PR #3181, opened 2017-08-23
Deploy assets to S3 in a single tar archive

jrjohnson:
Instead of sending individual files we upload a versioned tar archive which can later be extracted and used by our API server to create local servable assets.

This will let us do some better optimization on the API server as well as ensuring that service workers (sw.js) files which are *not* fingerprinted will not be allowed to overwrite one another.

Requires ilios/ilios#1912


jrjohnson:
Deploy preview ready!

Built with commit 798917dc94ab9a47b541ec999114acdfa06d3c45

https://deploy-preview-3181--ilios-frontend.netlify.com

PR #3185, opened 2017-08-25
Fix sessions filter on list to wait for input

### LINKED ISSUE #3184: Session list filter is janky
We need to debounce user input here to avoid re-rendering the list so often.

jrjohnson:
Debouncing this filter and caching the value locally ensures that every
typed character doesn't cause the table to re-render which is slow and
very intensive. Instead we record each char as it is typed so the input
field stays up to date, but we only send the value up to be processed if
there is a 250ms delay.

Fixes #3184

### DIFF
--- a/app/components/session-table.js
+++ b/app/components/session-table.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import Table from 'ember-light-table';
 import escapeRegExp from '../utils/escape-reg-exp';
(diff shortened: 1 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit b4a8a5a5b05a1653609a56f00bab842a6637d87c

https://deploy-preview-3185--ilios-frontend.netlify.com

PR #3193, opened 2017-08-27
Remove default limits from API queries

### LINKED ISSUE #3192: Remove huge limits
Now that ilios/ilios#1910 has landed and the API no longer has default limits we should strip all of our limit code our of the frontend where it was input simply to ensure we got all data `{limit: 1000}`, {limit: 100000}` etc...

jrjohnson:
All of these limits were in place to ensure we got all data from the API
when it was limiting requests to 20 items by default. As this default
was removed in v1.23 we can remove this code as well.

Fixes #3192

### DIFF
--- a/app/components/assign-students.js
+++ b/app/components/assign-students.js
@@ -68,7 +68,6 @@ export default Component.extend({
       filters: {
         schools: [school.get('id')],
       },
-      limit: 1000,
     });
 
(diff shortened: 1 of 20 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 1faed1129c0e1b2dd656bea2138795c4c7231585

https://deploy-preview-3193--ilios-frontend.netlify.com

PR #3195, opened 2017-08-28
Remove actions that don't work anymore

jrjohnson:
We now display the entire session list at the same time so there is no
paging necessary.


jrjohnson:
Deploy preview ready!

Built with commit 1d969dec7a2fb521818deb21d6ac34833c607d6c

https://deploy-preview-3195--ilios-frontend.netlify.com

PR #3196, opened 2017-08-28
Fix incorrectly named test

jrjohnson:
This test name was wrong, now it is right. Entropy is still going to
kill us all, but at least this test name will now be correct.


jrjohnson:
Deploy preview ready!

Built with commit 36a511310c7efba6af9ceb2a6724941997ef39bf

https://deploy-preview-3196--ilios-frontend.netlify.com

PR #3202, opened 2017-08-30
Calendar learning unit - work in progress, do not merge

gboushey:
Added a calendar to the new offering form on the courses and sessions page.  Displays events attached to offerings derived from the learner groups selected in the "Available Learner Groups" menu (in the form).  

@saschaben @jrjohnson would you take a look and 1) let me know if the data is what you're looking for, and 2) what layout and style elements you'd like for this (position of calendar, toggle on/off for calendar display, color of scheduled events and proposed new offering.  

Thanks!


jrjohnson:
Deploy preview ready!

Built with commit 19afa73b9abddc59fbbcc609b396e299acae103e

https://deploy-preview-3202--ilios-frontend.netlify.com

saschaben:
1) Is it possible to put the calendar one level up -- so that instead of being a part of the individual offering form, it is above the offering list, and responsive to ALL offerings for a single session? If would be inline with the "Offering" label and the "add New" button.

2) Colors: the presentation of new offerings as they are scheduled should reflect in the calendar with their session_type color; ALL other events on the calendar that are NOT a part of the current session can be greyed out, I think...block color is  grey, and text/icon/detail info dark grey.

3) Info: I think this is the info needed. If we could also add in the mouseover or on the calendar blocks the Learner Groups associated with an activity, and the Instructor (initial + lastname), that would be ideal.

4) I think a toggle on/off is in order, especially if we can tackle item #1. But regardless, users should be able to open/collapse the calendar view.

jrjohnson:
@saschaben:
I don't understand why it would go a level up since the context is that offering in comparison with every other offering in the session. It will show up when you open any one of the offerings, but it will always be centered on and highlight the offering you are editing / creating. If we move it up a box what day would it even show?

2 - @gboushey you can get the correct color for the current offering from session.sessionType.calendarColor

3 - changing the info in the box is probably going to be a lot more work since we will need to rework the calendar components themselves and then refactor the entire calendar to figure out how to pass in the right display component. I would estimate this as adding 2 - 3 weeks to this project.
Right now it contains the same details as the dashboard calendar so it was easy to create. 

I think we probably *do eventually* want to go this route (maybe even having the dashboard calendar for faculty have some of this same stuff, but it will be a lot more work right now unless we just shoehorn it into the title field.

saschaben:
@jrjohnson wrote:
>> I don't understand why it would go a level up since the context is that offering in comparison with every other offering in the session. It will show up when you open any one of the offerings, but it will always be centered on and highlight the offering you are editing / creating. If we move it up a box what day would it even show?

**A:** ---OK. I was not seeing anything but the offering currently being worked on showing in the calendar on the PR preview. As long as it presents ALL offerings for a session, and ALL other offerings for associated peeps (in this course and elsewhere), we're good. Location of the calendar is going to be tricky though as it pushes so much of the offering form down offscreen. Let's consider layout options.

>> 3 - changing the info in the box is probably going to be a lot more work since we will need to rework the calendar components themselves and then refactor the entire calendar to figure out how to pass in the right display component. I would estimate this as adding 2 - 3 weeks to this project.
Right now it contains the same details as the dashboard calendar so it was easy to create. 

>> I think we probably *do eventually* want to go this route (maybe even having the dashboard calendar for faculty have some of this same stuff, but it will be a lot more work right now unless we just shoehorn it into the title field.

**A:** ---OK. Not a problem for this pass. Let's review how to expose this information properly moving forward as you suggest. Although shoehorning it in to the title field wouldn't be that awful. Just put 'em in brackets appending the title.

jrjohnson:
Commits moved to ##3285 as I'm taking over from @gboushey.

PR #3216, opened 2017-09-01
Prevent login for from displaying before redirect

### LINKED ISSUE #3200: Flash of login form when using shibboleth redirect
If you visit the frontend of a shib enabled site you get a flash of the login for before being redirected to the correct IDP. This isn't good and should be fixed.

jrjohnson:
Without this change there is a moment before a redirect takes place
where the route thinks it should display the login form. This moment can
last a very long time on slow connections and looks ugly, so lets prevent
it by not resolving a promise... forever.

Fixes #3200


jrjohnson:
Deploy preview ready!

Built with commit 57eb12ddc03486a497dbf746a8e686207d5e8c64

https://deploy-preview-3216--ilios-frontend.netlify.com

PR #3222, opened 2017-09-06
Separate travis build into stages

jrjohnson:
This ensures that deploy will only run if everything else passes.
Also added an install stage to pre-cache yarn and bower deps to speed up the build a scooch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,51 +1,47 @@
 ---
 dist: trusty
 language: node_js
-node_js:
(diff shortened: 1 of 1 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 47b859ce61275d88e9b6a20c75a50621133cc0c7

https://deploy-preview-3222--ilios-frontend.netlify.com

PR #3224, opened 2017-09-06
Display notice when user is offline

### LINKED ISSUE #3212: Provide better error message for offline browser
right now our errors are pretty ugly. something more user-friendly would be nice, maybe a modified "rats" insert page....

jrjohnson:
The more our app works while disconnected the more important it will be
to display a warning to users when they are not online.  This listens
for the browser offline/online events and puts up a warning when a user
is offline. It also starts a timer to check the connection status again,
but this is mainly an empty gesture for a better experience.

This needs tests, but I'm hoping for some feedback. @saschaben / @dartajax can you give this a try? After you load the preview you can either disconnect your wifi or else open the chrome developer tools network tab and choose 'Offline'. You will then get a message:

offline

How can the style be improved?
Do the messages make sense?
Do the buttons make sense and do what you expect?

fixes #3212


jrjohnson:
Deploy preview ready!

Built with commit 58ec924250c0941fb73b6c33b6d15adc587be75a

https://deploy-preview-3224--ilios-frontend.netlify.com

dartajax:
You spelled "unable" incorrectly. It is not "unalbe".

dartajax:
I'd say make it fifteen seconds. My $.02 - other than that - cool.

jrjohnson:
It increases each time. So 5, 10, 20, 40 and then sticks at 50 seconds. Does that not make sense?

saschaben:
@jrjohnson wrote:
> It increases each time. So 5, 10, 20, 40 and then sticks at 50 seconds. Does that not make sense?

That makes sense.

PR #3228, opened 2017-09-07
Invalidate the cloudfront cache when we activate

jrjohnson:
When a new frontend.tar.gz is activated we need to clear the cloudfront
cached previous version. Without this change users would need to wait
for cloudfront to notice the new version at all.


jrjohnson:
Deploy preview ready!

Built with commit b7a47155687b6ddef64af47009f0383d35ae8be0

https://deploy-preview-3228--ilios-frontend.netlify.com

PR #3245, opened 2017-09-16
Cache index.html with a service worker

jrjohnson:
This brings us several steps closer to offline support, but more importantly it avoids a network request when visiting the app which will speed up initial rendering.

Includes a modified error page that notices if a user is offline and replaces all error data with an offline message since errors are almost certainly coming from that anyway.

### DIFF
--- a/app/components/connection-status.js
+++ b/app/components/connection-status.js
@@ -13,7 +13,7 @@ export default Component.extend(DomMixin, {
   didInsertElement() {
     this._super(...arguments);
     if (!navigator.onLine) {
-      this.wentOffline();
+      this.get('changeConnectionState').perform(false);
     }
     this.addEventListener(window, 'online', () => {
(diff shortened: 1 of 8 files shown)


jrjohnson:
Deploy preview ready!

Built with commit bfe3722cc9aef35f2576f1c3daf38f5b77f714fd

https://deploy-preview-3245--ilios-frontend.netlify.com

dartajax:
Here's what happens with a hard refresh when not on the network. The Ilios flyout menu appears to be trying to load. 

image

jrjohnson:
@dartajax what do you think we should do about that? It *is* loading, but it will never finish with no connection. I can sort of see some options, but I'm wondering what you see as the best way to handle that?

dartajax:
I don't have a problem with the way that happened. I just wanted to point it out. I'm going to merge this - it looks good.

PR #3246, opened 2017-09-18
Rebuild a single test in a new pattern

jrjohnson:
Using page objects and data-test-* selectors this is a more maintainable way to write acceptance tests. While this just modifies a single test, the long term plan is to refactor every test this way.


jrjohnson:
Deploy preview ready!

Built with commit f6a8ca7a7c604178b522e77c01bae8e9c15cc5a2

https://deploy-preview-3246--ilios-frontend.netlify.com

PR #3256, opened 2017-09-20
hide "create school" button from non-developers.

### LINKED ISSUE #3114: Permission error when creating a new school
We have a permission mis-match here. The frontend allows too many people to try and create a school, whereas we really only want some roles to have this ability.

We should limit access to the create school button to those role with this permissions.

stopfstedt:
fixes #3114 

according to the rules enforced in the [Ilios backend](https://github.com/ilios/ilios/blob/v3.37.2/src/Ilios/AuthenticationBundle/Voter/Entity/SchoolEntityVoter.php#L45), only users with Developer role can create new schools.

### DIFF
--- a/app/components/school-list.js
+++ b/app/components/school-list.js
@@ -26,6 +26,7 @@ const Validations = buildValidations({
 });
 
 export default Component.extend(ValidationErrorDisplay, Validations, {
+  currentUser: service(),
   store: service(),
   classNames: ['school-list'],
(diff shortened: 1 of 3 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 44a4662b732ad10177790cd16a55a8ebd4980584

https://deploy-preview-3256--ilios-frontend.netlify.com

PR #3279, opened 2017-09-29
Copy learner groups

jrjohnson:
Groups are copied in place on either the cohort list or a subgroup list.
This makes it easier for users to duplicate structure and even membership
quickly.

Needs:
- [x] Translations
- [x] Review of all the ways groups can be copied to ensure this works

### DIFF
--- a/app/utils/clone-learner-group.js
+++ b/app/utils/clone-learner-group.js
@@ -1,0 +1,44 @@
+import RSVP from 'rsvp';
+
+const { map } = RSVP;
+
+
+/**
+ * Clones a group and all children returning all the groups in the order they were created
+ * so the first group is the top most group in the tree. If they are saved in order then each parent
+ * will be correctly saved before its children
+ * @method cloneLearnerGroup
+ * @param {Object} store
+ * @param {Object} group
+ * @param {Object} cohort
+ * @param {Boolean} withLearners
+ * @param {Object} | null parent
+ * @return {Promise.<Array>}
+ */
+export default async function cloneLearnerGroup(store, group, cohort, withLearners, parent = null) {
+  let newGroup = store.createRecord('learner-group', group.getProperties(
+    'title',
+    'location'
+  ));
+  newGroup.set('cohort', cohort);
+  if (parent) {
+    newGroup.set('parent', parent);
+  }
+  if (withLearners) {
+    const users = await group.get('users');
+    await map(users.toArray(), async user => {
+      await newGroup.addUserToGroupAndAllParents(user);
+    });
+  }
+  const instructors = await group.get('instructors');
+  newGroup.set('instructors', instructors);
+  const children = await group.get('children');
+  let newChildren = await map(children.toArray(), async child => {
+    return await cloneLearnerGroup(store, child, cohort, withLearners, newGroup);
+  });
+  let flat = newChildren.reduce((flattened, obj) => {
+    return flattened.pushObjects(obj.toArray());
+  }, []);
+
+  return [].concat([newGroup], flat);
+}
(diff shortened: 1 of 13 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 7e5a552358af516409e4b587c122b35379a730ba

https://deploy-preview-3279--ilios-frontend.netlify.com

jrjohnson:
@dartajax can you click test copying both top level learner groups and sub groups?

dartajax:
Can we get a call out on the mouse over? `copy`?

dartajax:
I love it.

jrjohnson:
@dartajax good call. Tooltip added.

app/utils/clone-learner-group.js
```diff
@@ -0,0 +1,32 @@
...
+  const instructors = await group.get('instructors');
+  newGroup.set('instructors', instructors);
+  const children = await group.get('children');
+  let newChildren = await map(children.toArray(), async child => {
+    return await cloneLearnerGroup(store, child, cohort, withLearners, newGroup);
+  });
+  let flat = newChildren.reduce((flattened, obj) => {
+    return flattened.pushObjects(obj.toArray());
+  }, []);
+  const newGroups = [].concat([newGroup], flat);
```
stopfstedt: just return here, no need for variable `newGroups`.
change to: `return [].concat([newGroup], flat);`

app/utils/clone-learner-group.js
```diff
@@ -0,0 +1,32 @@
+import RSVP from 'rsvp';
```
stopfstedt: just curious why this imports from `rsvp` directly, and not via `Ember.RSVP`. 
should we do this elsewhere too?

app/utils/clone-learner-group.js
```diff
@@ -0,0 +1,32 @@
...
+
+const { map } = RSVP;
+
+export default async function cloneLearnerGroup(store, group, cohort, withLearners, parent = null) {
```
stopfstedt: please add a docblock here, outlining structure of returned array (first item is copy of of given group etc).

app/utils/clone-learner-group.js
jrjohnson: This change will actually happen for us automatically with a code mode I have in #3120 so we don't need to worry about it too much. It is the new right way to do this, but in most cases we need the Ember object anyway so until that goes away either way is fine.

jrjohnson:
@stopfstedt good notes. Can you check my docblock I'm not sure what format to use for booleans is it just `bool` or does it get a `{Bool}`?

app/utils/clone-learner-group.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Clones a group and all children returning all the groups in the order they were created
+ * so the first group is the top most group in the tree. If they are saved in order then each parent
+ * will be correctly saved before it's children
+ * @method cloneLearnerGroup
+ * @param {Object} store
+ * @param {Object} group
+ * @param {Object} cohort
+ * @param bool withLearners
```
stopfstedt: should be `@param {Boolean} withLearners`

app/utils/clone-learner-group.js
```diff
@@ -0,0 +1,44 @@
+import RSVP from 'rsvp';
+
+const { map } = RSVP;
+
+
+/**
+ * Clones a group and all children returning all the groups in the order they were created
+ * so the first group is the top most group in the tree. If they are saved in order then each parent
+ * will be correctly saved before it's children
```
stopfstedt: typo: its, not it's

app/utils/clone-learner-group.js
```diff
@@ -0,0 +1,44 @@
...
+ * @method cloneLearnerGroup
+ * @param {Object} store
+ * @param {Object} group
+ * @param {Object} cohort
+ * @param bool withLearners
+ * @param {Object} | null partne
```
stopfstedt: typo: should be parent, not partne

jrjohnson:
OK ready again. This time with 2 less typos and one correct param.

PR #3280, opened 2017-09-29
escape regex further downstream and dont print out escaped search filter value.

### LINKED ISSUE #3260: Course search: backslash escaping avalance
using a backslash as input in the course search will trigger a cascade of escapes:

selection_396

fix this.

stopfstedt:
also, use filter value from request param if given.

fixes #3260


jrjohnson:
Deploy preview ready!

Built with commit 6ba57fffc78fd1ede45bcdcd09fd054fe456d0c4

https://deploy-preview-3280--ilios-frontend.netlify.com

PR #3286, opened 2017-10-02
Slight build improvements

jrjohnson:
Removed the prepare stage as it wasn't really helping and maybe was causing us grief. Replaced it with a greenkeeper stage which will only run on a greenkeeper branch.

I'm no longer caching `node_modules` as a big part of every build is taken up with Travis reading and writing that very large cache which is basically made irrelevant by the yarn cache.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,11 @@ script: ember exam --split=$SPLIT --partition=$PARTITION --filter=$FILTER --para
 
 jobs:
   include:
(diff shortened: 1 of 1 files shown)


jrjohnson:
Deploy preview ready!

Built with commit e71e55903a873c4108e10f9a7786a332493b2cf3

https://deploy-preview-3286--ilios-frontend.netlify.com

PR #3292, opened 2017-10-05
Display the users current timezone in the offering editor

### LINKED ISSUE #3291: Add Time Zone To Offerings Editor
In case a user happens to have their computer set to a different (non-West Coast LA) time zone, or just to confirm all is well, we should display the time zone to which the offering changes will be set. This could go somewhere like what is shown below.

image

jrjohnson:
This makes it easier for users to notice if they are changing a time
incorrectly because their browsers timezone isn't what they expect it to
be.

Fixes #3291

### DIFF
--- a/app/helpers/browser-timezone.js
+++ b/app/helpers/browser-timezone.js
@@ -1,0 +1,8 @@
+import Ember from 'ember';
+import moment from 'moment';
+
+export function browserTimezone() {
+  return moment.tz.guess();
+}
+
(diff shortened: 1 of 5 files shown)


jrjohnson:
Deploy preview ready!

Built with commit a5bee23697f6467be72195a3a9ccc4abdefbc523

https://deploy-preview-3292--ilios-frontend.netlify.com

Issue #3298, opened 2017-10-06
Apply regex fixes from frontend#3280 to all search/filter points globally


dartajax:
I noticed that when performing a search for a learner group that contained a period "." in its title that strange stuff was happening - I understand that is not a very technical description. 

Refer to #3280 

- [x] courses
- [x] programs
- [x] instructor groups
- [x] learner groups
- [x] sessions
- [x] my materials

saschaben:
places to address this:

``` $ grep -ri 'escape-reg-exp' app
[app/controllers/courses.js:import escapeRegExp from '../utils/escape-reg-exp';]
app/controllers/programs.js:import escapeRegExp from '../utils/escape-reg-exp';
app/controllers/instructor-groups.js:import escapeRegExp from '../utils/escape-reg-exp';
app/controllers/learner-groups.js:import escapeRegExp from '../utils/escape-reg-exp';
app/components/session-table.js:import escapeRegExp from '../utils/escape-reg-exp';
app/components/my-materials.js:import escapeRegExp from '../utils/escape-reg-exp'; ```

stopfstedt:
fixed.

PR #3299, opened 2017-10-06
fixes title filtering for programs and instructor-groups.

stopfstedt:
refs #3298.

### DIFF
--- a/app/controllers/instructor-groups.js
+++ b/app/controllers/instructor-groups.js
@@ -43,11 +43,8 @@ export default Controller.extend({
     'changeTitleFilter.lastSuccessful.value',
     'instructorGroups.[]',
     async function(){
-      let title = this.get('changeTitleFilter.lastSuccessful.value');
-      if (!isPresent(title)) {
-        const titleFilter = this.get('titleFilter');
-        title = isBlank(titleFilter) ? '' : titleFilter ;
-      }
+      const titleFilter = this.get('titleFilter');
+      const title = isBlank(titleFilter) ? '' : titleFilter ;
       const cleanTitle = escapeRegExp(title);
       let exp = new RegExp(cleanTitle, 'gi');
       const instructorGroups = await this.get('instructorGroups');
--- a/app/controllers/instructor-groups.js
+++ b/app/controllers/instructor-groups.js
@@ -1,12 +1,11 @@
 import Ember from 'ember';
-import DS from 'ember-data';
 import { task, timeout } from 'ember-concurrency';
 import escapeRegExp from '../utils/escape-reg-exp';
 
-const { computed, Controller, RSVP, isEmpty, isPresent, inject } = Ember;
+const { computed, Controller, RSVP, isBlank, isEmpty, isPresent, inject } = Ember;
 const { gt } = computed;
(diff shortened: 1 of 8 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 3712fb0f39e5e518e3b8df063ac591bd88afd1b1

https://deploy-preview-3299--ilios-frontend.netlify.com

app/controllers/instructor-groups.js
```diff
@@ -18,59 +17,49 @@ export default Controller.extend({
   schoolId: null,
   titleFilter: null,
   showNewInstructorGroupForm: false,
-  newInstructorGroups: [],
-  instructorGroups: computed('selectedSchool', function(){
-    let defer = RSVP.defer();
+  newInstructorGroup: null,
+  deletedInstructorGroups: [],
```
jrjohnson: Why is one `null` and on and Array?

app/controllers/instructor-groups.js
```diff
@@ -18,59 +17,49 @@ export default Controller.extend({
...
-          });
-        }
-        defer.resolve(filteredInstructorGroups.sortBy('title'));
-      });
-
-
-      return PromiseArray.create({
-        promise: defer.promise
-      });
+      if (!isPresent(title)) {
```
jrjohnson: I don't understand why this double check? Can't we just depend on `this.get('titleFilter')` and not worry about the task value?

app/controllers/instructor-groups.js
stopfstedt: agreed, will change this. also, this doesn't need to be an array at all, since the UI doesn't support bulk deletions at this point. i'm going to change this to be a property that tracks just the most recently deleted instructor group, should be sufficient.

app/controllers/instructor-groups.js
stopfstedt: you are absolutely correct. i'm simplifying this.

app/controllers/instructor-groups.js
```diff
@@ -18,59 +17,46 @@ export default Controller.extend({
...
+    this.set('titleFilter', value);
     yield timeout(250);
-
-    return clean;
+    return value;
   }).restartable(),
 
   hasMoreThanOneSchool: gt('model.schools.length', 1),
   filteredInstructorGroups: computed(
     'changeTitleFilter.lastSuccessful.value',
```
jrjohnson: Since this now depends online on `titleFilter` `'changeTitleFilter.lastSuccessful.value'` should be `'titleFilter'`.

app/controllers/programs.js
```diff
@@ -21,40 +20,31 @@ export default Controller.extend({
...
-            return isPresent(program.get('title')) && program.get('title').match(exp);
-          });
-        }
-
-        defer.resolve(filteredPrograms.sortBy('title'));
-      });
-
-      return PromiseArray.create({
-        promise: defer.promise
+  filteredPrograms: computed('changeTitleFilter.lastSuccessful.value', 'programs.[]', async function() {
```
jrjohnson: Same thing here. `changeTitleFilter.lastSuccessful.value` should be `titleFilter`.

PR #3332, opened 2017-10-18
Don't override font-face

jrjohnson:
Ilios/common is providing a nice set of default system fonts that we
don't need to override here.


jrjohnson:
Deploy preview ready!

Built with commit d9d9f86c28d1ec957bd02013df25dd63ab7d779e

https://deploy-preview-3332--ilios-frontend.netlify.com

PR #3335, opened 2017-10-18
Preload the user model and the users roles

jrjohnson:
This makes the initial page rendering (especially the navigation) much
smoother.

The best way to see this behavior is to load the dashboard in a large screen (or zoomed way out) so the navigation block shows up on the left. Before this fix the block would show up with a spinner initially, after this change it does not.


jrjohnson:
Deploy preview ready!

Built with commit 418fdff24796b59d07616a666300890da40b9f3a

https://deploy-preview-3335--ilios-frontend.netlify.com

PR #3343, opened 2017-10-24
Don't attempt to deploy cron run builds

jrjohnson:
Travis runs our tests daily, but if we try and deploy these builds then
there is an error since that version has already been deployed.


jrjohnson:
Deploy preview ready!

Built with commit 25a5e0f7731de7915e431ebe7c9eb0be128828ed

https://deploy-preview-3343--ilios-frontend.netlify.com

PR #3375, opened 2017-11-08
Use computed properties instead of tasks in learner group management

### LINKED ISSUE #3374: Learner Not Available For Addition To Learner Group Until After Refresh Previously In Other School
There is really no reason for her not to be available immediately. She was in SOD previously but has been added to the SOM Cohort Class of 2021 BRIDGES as her primary cohort. This user is in many of the other Learner Groups in her cohort already.

This ticket may need to be adjusted since there are at least two issues.

1. Hard refresh should not be required to reveal users for selection.
2. Her Primary School should now be SOM not SOD. That should have happened when the Primary Cohort was changed.

User ID: 8702
Learner Group: 9822

https://ilios3-demo.ucsf.edu/users/8702

image

jrjohnson:
Tasks are great for managing some concurency, but they don't manage
keeping themselves up to date when changes are made as well as CPs do.
For displaying this learner group data CPs are the way to go.

Fixes #3374

### DIFF
--- a/app/components/learnergroup-cohort-user-manager.js
+++ b/app/components/learnergroup-cohort-user-manager.js
@@ -16,8 +16,8 @@ export default Component.extend({
   classNames: ['learnergroup-cohort-user-manager'],
   sortBy: 'firstName',
   users: [],
-  usersBeingMoved: [],
-  selectedUsers: [],
+  usersBeingMoved: null,
+  selectedUsers: null,
   sortedAscending: computed('sortBy', function(){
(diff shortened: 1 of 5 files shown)


jrjohnson:
Deploy preview ready!

Built with commit c5c6f54ab909f7d14c4da3a0402a0c55b429c334

https://deploy-preview-3375--ilios-frontend.netlify.com

dartajax:
I have tested and will merge soon. 👍

PR #3376, opened 2017-11-08
Use correct store method to fetch all

jrjohnson:
store.query requires some parameters to be sent. When fetching all
records we need store.findAll


jrjohnson:
Deploy preview ready!

Built with commit 5def32eb9d642a3061721b9a0996e8d186de71c3

https://deploy-preview-3376--ilios-frontend.netlify.com

PR #3378, opened 2017-11-09
update to latest ilios-common

stopfstedt:
this pertains to corrections made in ilios/common#113.

### DIFF
--- a/app/components/learnergroup-summary.js
+++ b/app/components/learnergroup-summary.js
@@ -63,14 +63,13 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
       });
     });
   }),
-  saveSomeGroups(arr){
+  async saveSomeGroups(arr){
     let chunk = arr.splice(0, 5);
-    return all(chunk.invoke('save')).then(() => {
-      if (arr.length){
-        this.set('currentGroupsSaved', this.get('currentGroupsSaved') + chunk.length);
-        return this.saveSomeGroups(arr);
-      }
-    });
+    await chunk.invoke('save');
+    if (arr.length){
+      this.set('currentGroupsSaved', this.get('currentGroupsSaved') + chunk.length);
+      await this.saveSomeGroups(arr);
+    }
   },
   addUserToGroup: task(function * (user) {
     const learnerGroup = this.get('learnerGroup');
@@ -102,7 +101,6 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
     this.set('isSaving', false);
     this.set('totalGroupsToSave', 0);
     this.set('currentGroupsSaved', 0);
-
     this.set('isSaving', false);
   }).enqueue(),
   removeUsersToCohort: task(function * (users) {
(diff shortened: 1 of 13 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 000b252579c686b0d0bd03dd37fc3800d04c7bba

https://deploy-preview-3378--ilios-frontend.netlify.com

app/components/learnergroup-summary.js
```diff
@@ -63,14 +63,13 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
...
-  saveSomeGroups(arr){
+  async saveSomeGroups(arr){
     let chunk = arr.splice(0, 5);
-    return all(chunk.invoke('save')).then(() => {
-      if (arr.length){
-        this.set('currentGroupsSaved', this.get('currentGroupsSaved') + chunk.length);
-        return this.saveSomeGroups(arr);
-      }
-    });
+    await chunk.invoke('save');
```
jrjohnson: I think this should be `await all(chunk.invoke('save'));` because `invoke` doesn't return only one promise it returns an array of promises.

app/components/offering-editor-learnergroups.js
```diff
@@ -1,21 +1,22 @@
...
 import { isPresent, isEmpty } from '@ember/utils';
-const { sort } = computed;
+import RSVP from 'rsvp';
+
+const { map } = RSVP;
 
 export default Component.extend({
   classNames: ['offering-editor-learnergroups'],
 
   revisedLearnerGroups: computed('cohort.filteredAvailableLearnerGroups.[]', {
```
jrjohnson: While you are in here anyway can you get rid of the `get()` and `readonly` stuff and just make this looks like all the other CPs?

stopfstedt: on it.

app/components/learnergroup-summary.js
stopfstedt: adding an `all` to that results in the bulk-save to never complete. the modal just stays up forever, which may be indicative of an un-awaited/un-yielded promise somewhere. 

tried to approach this from multiple angles, but to no avail. i'm kinda stumped at this right now.

jrjohnson: Is it possible that it is throwing an error? Maybe slap a try / catch around it and see what shakes out?
`invoke` returns and array and `save` returns a promise. So this should be an array of promises.
https://emberjs.com/api/ember/2.16/classes/EmberArray/methods/invoke?anchor=invoke

Well if this for sure works with large groups without `all` I can live with merging as is. I'm just super curious now.

stopfstedt: me too. i'll further investigate, we have this "invoke/save" in a couple of places.

stopfstedt: no errors are being thrown, it just seems to leave the component(s) in a funky state after saving.
adding `all()` works, but then yielding from that function seems to complete the task at the point. resulting in the `isSaving` attribute to never reset to `false`, thus leaving the modal up forever.

i'll give up on this for the time being - removing the modal and adding `all()` seems to be the sensible thing to do.

PR #3387, opened 2017-11-16
Limit the length of the URL when we pre-load data

### LINKED ISSUE #3385: Unable to load complex course
If a course becomes too large or complex it starts to bump up against limits in the length of the URL when it attempts to pre-fetch filtered data in the course route.

jrjohnson:
When there are hundreds of sessions in a course the `filter` parameters
we send to pre-load objectives and session types for sessions can get
quite long and eventually break the Apache 8K upper limit for URL size.
This breaks these up into several smaller requests which do not reach
this limit.

Fixes #3385

### DIFF
--- a/app/routes/course/index.js
+++ b/app/routes/course/index.js
@@ -22,14 +22,25 @@ export default  Route.extend({
       return;
     }
 
-    return all([
-      store.query('session', {filters: {course}}),
-      store.query('offering', {filters: {courses}}),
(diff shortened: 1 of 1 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 6a456483a242914e29dcfba13dafa282190ff6ad

https://deploy-preview-3387--ilios-frontend.netlify.com

PR #3397, opened 2017-11-20
Improve print course experience

### LINKED ISSUE #410: Print Summary Causes Ilios Menu and Header to Disappear
I was testing the Print Summary button on a Course and then used the browser (Chrome) back button.  All the header with Ilios logo and the left side with the toolbars had disappeared.  There is no way to navigate away from the Print Summary screen without using the back button of the browser.  Close the window and you are out of Ilios.  Also, the Logout button is missing since it is on the header.

### LINKED ISSUE #878: returning from print view fails to load sidebar menu
if you navigate to the event detail print view, and then return (using the "back" button) to the dashboard, the dash loads but the slide out nav menu fails to load, requiring a full browser refresh.

jrjohnson:
We were using a custom outlet and going a bunch of work to hide the
navigation and the header. Instead we can just use CSS to hide this
stuff when the course is actually printed.

Fixes #410 
Fixes #878

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -10,8 +10,6 @@ export default Controller.extend({
   i18n: service(),
   headData: service(),
 
-  showHeader: true,
-  showNavigation: true,
   titleTokenKeys: null,
 
   currentlyLoading: false,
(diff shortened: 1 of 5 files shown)


jrjohnson:
Deploy preview ready!

Built with commit aa77e994ad044a9689a3f385f65b4aeb673c89df

https://deploy-preview-3397--ilios-frontend.netlify.com

jrjohnson:
Deploy preview ready!

Built with commit 3d3af796b3ad19b59e83c670d9738473f6432e88

https://deploy-preview-3397--ilios-frontend.netlify.com

dartajax:
Greatly improved indeed!

PR #3406, opened 2017-11-28
Cleanup some Cruft

jrjohnson:
The session afterModel hook was doing some weird promise stuff to grab
the session types which it turns out we are not actually even using anymore so I removed the
entire block.

We are no longer using showBackToCourseListLink anywhere to hide this
link so there is no reason to set it.


jrjohnson:
Deploy preview ready!

Built with commit a4f4548420c9b5597bfbf415c2b5ab88ba0a495e

https://deploy-preview-3406--ilios-frontend.netlify.com

PR #3421, opened 2017-12-03
Enable Order Rules

jrjohnson:
Enabled three rules that enforce consistent property ordering in ember component, controller, and route objects. Fixing these all at once seemed like a big job so instead I added an exclude comment to each file so we can fix these as we edit the file or when we have some time, but still enforce the rules on all new files.


jrjohnson:
Deploy preview ready!

Built with commit eef4794654f1a4b4670e729392f0964829f62811

https://deploy-preview-3421--ilios-frontend.netlify.com

PR #3426, opened 2017-12-05
Add CSV download for reports

### LINKED ISSUE #3410: Add the ability to download/export report results as csv

jrjohnson:
Lets users export the report results they're currently looking at.

Fixes #3410

### DIFF
--- a/app/components/dashboard-myreports.js
+++ b/app/components/dashboard-myreports.js
@@ -2,12 +2,14 @@ import { inject as service } from '@ember/service';
 import { computed } from '@ember/object';
 import Component from '@ember/component';
 import RSVP from 'rsvp';
-import { task } from 'ember-concurrency';
-
+import { task, timeout } from 'ember-concurrency';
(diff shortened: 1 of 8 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 976a85d1b2a256c48b3e80be31df03abeb8d1b10

https://deploy-preview-3426--ilios-frontend.netlify.com

saschaben:
seems we have a bug in the academic year selector in MyReports: if you select the first item in a select list, instead of picking the selection, it seems to grab the first deefault ID in the list. EX:

build the report 
for MEDICINE
all TERMS
associated with the COURSE
for academic year 2016-17
Airway Blood & Circulation

the result will build the report for Prologue --- which is the default load selected prior to picking a year.

This does not occur if you pick a course which is NOT the first in the year-filtered list.

jrjohnson:
Nice catch @dartajax. Should be fixed now.

dartajax:
As far as I can tell, this looks good.

PR #3429, opened 2017-12-06
Select correct default when a year is selected

### LINKED ISSUE #3427: Report Selector with Academic Year not selecting correct predicate
seems we have a bug in the academic year selector in MyReports: if you select the first item in a select list, instead of picking the selection, it seems to grab the first deefault ID in the list. EX:

build the report
for MEDICINE
all TERMS
associated with the COURSE
for academic year 2016-17
Airway Blood & Circulation

the result will build the report for Prologue --- which is the default load selected prior to picking a year.

This does not occur if you pick a course which is NOT the first in the year-filtered list.

jrjohnson:
We needed to reset the selected object id anytime the year was changed,
otherwise if a selection wasn't made from the dropdown the wrong default
would be used instead.

Fixes #3427

### DIFF
--- a/app/components/new-myreport.js
+++ b/app/components/new-myreport.js
@@ -323,6 +323,14 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
     this.set('schoolChanged', true);
   }),
 
+  resetCurrentPrepositionalObjectId: task(function* () {
+    const list = yield this.get('filteredPrepositionalObjectIdList');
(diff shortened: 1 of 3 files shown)


jrjohnson:
Deploy preview ready!

Built with commit f097b1360aec51dc889015ccd9ce09706901724a

https://deploy-preview-3429--ilios-frontend.netlify.com

PR #3437, opened 2017-12-08
Simplify rendering of learner groups in offering form

### LINKED ISSUE #2850: Attempting to editing offerings twice in same web-session throws blank page on update
NOTE: This error was reproducible in some sessions/offerings, but not in others.  User experienced the problem in production, but we reproduced on https://ilios3-stage.ucsf.edu.  Here's the problem:

When editing an offering's details in the UI, you can do so once to any offering and save the item without problem or delay.  

HOWEVER, in most cases, when you attempt a second edit (of the same offering OR another!) during the same user session, one of three things happens:

1. clicking on the edit icon for (any) offering throws a blank page.
2. clicking on the edit icon allows you to edit the offering, but throws a blank page upon save.  Refreshing the browser from the blank page shows that the change has actually been saved and the offering is updated, but it takes a refresh to see that.
3. The task works as expected. 

Here are the links we used to create both of the problem scenarios listed above (#1 and 2), but we get mixed results when trying each multiple times: sometimes it works, sometimes it does not:
1. https://ilios3-stage.ucsf.edu/courses/1143/sessions/33533
2. https://ilios3-stage.ucsf.edu/courses/978/sessions/30677

jrjohnson:
We're bumping up against an infinite rendering limit in ember that is
alleviated by simplifying this data and the template.

Fixes #2850 and unblocks updating ember-concurrency.

There is already a test in place for this issue, but offering management, specifically learner group selection, should be click tested.

### DIFF
--- a/app/components/learnergroup-selection-manager.js
+++ b/app/components/learnergroup-selection-manager.js
@@ -1,37 +1,11 @@
 /* eslint ember/order-in-components: 0 */
 import { inject as service } from '@ember/service';
 import Component from '@ember/component';
-import { computed } from '@ember/object';
-import RSVP from 'rsvp';
-const { Promise, all } = RSVP;
 
 export default Component.extend({
   i18n: service(),
-  init(){
-    this._super(...arguments);
-    this.set('sortBy', ['title']);
-  },
   classNames: ['learnergroup-selection-manager'],
   filter: '',
-  sortBy: null,
   cohorts: null,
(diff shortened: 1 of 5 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 3f97f3a91ac13538660fc124cb6d55f7646330fa

https://deploy-preview-3437--ilios-frontend.netlify.com

PR #3446, opened 2017-12-14
Immediately load new worker if there is an API update

jrjohnson:
The user should not need to take any action in this case. We should
immediately load any new workers that are waiting.


jrjohnson:
Deploy preview ready!

Built with commit 2b028894fbbf07560715f91b6866854dac9ada43

https://deploy-preview-3446--ilios-frontend.netlify.com

PR #3474, opened 2018-01-17
Fix cancel editing of empty session description

### LINKED ISSUE #3210: After Inline Save Cancel, Hard Refresh Is Required To Clear Entered Data
What I mean by that is ...

Edit a Session's description and then click the Cancel button. This is demonstrated below.

image

image

After clicking Cancel, the updated Session Description still remains.

image

Do a hard refresh on the screen and it's gone.

Back to the original state. This should have occurred earlier after pressing the Cancel button.

image

jrjohnson:
When there is no description originally and we cancel editing we should
revert the description to null otherwise the user has to refresh the
page.

Fixes #3210


jrjohnson:

Built with commit ba725404e6a2c0c51963957a70e977136c89e2fd

https://deploy-preview-3474--ilios-frontend.netlify.com

PR #3477, opened 2018-01-18
Display all users in the tree when managing subgroup

### LINKED ISSUE #3476: Learner group manager not loading the full tree
When a user clicks on Manage for a learner group they should see all the learners in the current group and all the learners in all subgroups as well.

jrjohnson:
If learner groups are built in-correctly it is possible for a learner to
be in a sub-group and not in the parent. We should account for this be
loading the entire group tree when we need it for editing.

Fixes #3476


jrjohnson:

Built with commit a20a76e76dfa66c48d6ad835a2554d96e5b4387e

https://deploy-preview-3477--ilios-frontend.netlify.com

PR #3479, opened 2018-01-19
Avoid destroying the offering form when attributes change

### LINKED ISSUE #2838: Instructor Picker for Offering fails when desired name is already listed as Course Director
It works the way it should be working when assigning instructors to ILM's. The ILM Instructor selector is working fine. The standard Offering Instructor picker is NOT.

See below ... 

"Stan Fields" is a Course Director as shown below.

image

# ILM ...

"Stan Fields" can be added as Instructor for the ILM as shown below.

image

image

When an User is already assigned as Director for a Course, that user can't be added as Instructor for an Offering in that Course.  Conversely, when a User is already added as Instructor for an Offering, they CAN still be added as Director for the Course.

# Standard Offering

Running a search for "Stan Fields" blows up. It does not say no records returned. It does not provide other results. "Stan Fields" cannot be added because he is a Course Director in this Course.

image

And then ...

No "Stan" - no results at all - "Stan" can't be added.

image

jrjohnson:
By removing this check we avoid a re-render anytime something is loaded
or changed. Specifically this was an issue when some users attached to a
course as a director were queried from the API. It would cause a
re-render which destroyed the search box.

Fixes #2838

### DIFF
--- a/app/components/offering-form.js
+++ b/app/components/offering-form.js
@@ -107,6 +107,7 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
   offeringsToSave: 0,
   savedOfferings: 0,
   recurringDayOptions: null,
+  loaded: false,
   associatedSchools: computed('cohorts.[]', function(){
     return new Promise(resolve => {
       const cohorts = this.get('cohorts');
@@ -267,7 +268,11 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
   }),
 
 
-  loadDefaultAttrs(){
+  loadDefaultAttrs() {
+    let loaded = this.get('loaded');
+    if (loaded) {
+      return;
+    }
     let startDate = moment(this.get('defaultStartDate')).hour(8).minute(0).second(0).toDate();
     let endDate = moment(this.get('defaultStartDate')).hour(9).minute(0).second(0).toDate();
     const room = 'TBD';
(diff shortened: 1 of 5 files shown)


jrjohnson:

Built with commit 78ff7b2aea4b9f62c8a1a82c520ce43b13d510ec

https://deploy-preview-3479--ilios-frontend.netlify.com

dartajax:
image

So Dabney and Pharm1 are directors of this course. 

Add Dabney (or Nic Gutierrez if Dabney is not available).

image

As soon as you starting searching for Pharm, Dabney is already removed from the list of instructors.

dartajax:
Dabney is missing - Pharm remains.

image

dartajax:
If you save them individually, it appears to work. This is probably not the most logical workflow for users since they may add a number of instructors at the same time.

app/components/user-search.js
```diff
@@ -42,7 +42,7 @@ export default Component.extend({
   roles: '',
   availableInstructorGroups: null,
   currentlyActiveInstructorGroups: null,
-  search: task(function * (searchTerms = '') {
+  search: task(function* (searchTerms = '') {
```
stopfstedt: why?

jrjohnson: VSCode must have done that for me. 😠

jrjohnson: Fixed.

dartajax:
This now appears to work as it should. Shall I merge it?

PR #3497, opened 2018-01-26
fixed indeterminate checkbox

stopfstedt:
you can't set the `indeterminate` attribute in HTML, so i replaced `<input>` with `{{input}}` component.

### DIFF
--- a/app/templates/components/assign-students.hbs
+++ b/app/templates/components/assign-students.hbs
@@ -45,10 +45,11 @@
       <thead>
         <tr>
           <th class='text-left clickable' colspan="1" onclick={{action 'toggleCheck'}}>
-            <input
+            {{input
              type='checkbox'
-             checked={{eq selectedUserIds.length filteredStudents.length}}
-             indeterminate={{and (gt selectedUserIds.length 0) (lt selectedUserIds.length filteredStudents.length)}}>
+             checked=(eq selectedUserIds.length filteredStudents.length)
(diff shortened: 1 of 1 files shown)


jrjohnson:

Built with commit 91dfca8fb6fe912176752384c687c2588d9512ae

https://deploy-preview-3497--ilios-frontend.netlify.com

stopfstedt:
i'm referring to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Indeterminate_state_checkboxes

> In addition to the checked and unchecked states, there is a third state a checkbox can be in: indeterminate. This is a state in which it's impossible to say whether the item is toggled on or off. This is set using the HTMLInputElement object's indeterminate property via JavaScript (it cannot be set using an HTML attribute):

jrjohnson:
I'm pretty sure the input here isn't real HTML then it must get fixed when ember compiles the template because there is an indeterminate state on master right now that works. Unless I'm misunderstanding where this bug actually shows up.

stopfstedt:
the bug label was a bit strong, i must admit, since it actually _works_.
ember "magic", i don't like it.
i'll close it down tho.

jrjohnson:
I think it isn't magic so much as the fact that none of our HTML is really html it is all handlebars templates that get compiled into javascript anyway. So when these checkboxes are rendered they are placed in the DOM by JS and can take advantage of `indeterminate` state.

Issue #3466, opened 2018-01-12
Replace all oneway controlls


jrjohnson:
Ember Once Way Controls Deprecated

https://github.com/DockYard/ember-one-way-controls#deprecated-ember-one-way-controls

... bad news for us as we use this syntax heavily.

stopfstedt:
```
$ grep -r 'oneWay' app
app/mixins/newuser.js:const { oneWay } = computed;
app/mixins/newuser.js:  cohorts: oneWay('loadCohorts.lastSuccessful.value'),
app/controllers/curriculum-inventory-reports.js:const { gt, oneWay, sort } = computed;
app/controllers/curriculum-inventory-reports.js:  schools: oneWay('model'),
app/components/detail-mesh.js:import { oneWay, sort } from '@ember/object/computed';
app/components/detail-mesh.js:  terms: oneWay('subject.meshDescriptors'),
app/components/course-objective-manager.js:const { filterBy, gt, none, oneWay, sort, uniq } = computed;
app/components/course-objective-manager.js:  id: oneWay('cohort.id'),
app/components/session-offerings.js:const { alias, oneWay } = computed;
app/components/session-offerings.js:  offerings: oneWay('session.offerings'),
app/components/session-offerings-list.js:const { oneWay, sort } = computed;
app/components/session-offerings-list.js:  offerings: oneWay('session.offerings'),
app/components/session-overview.js:const { oneWay, sort } = computed;
app/components/session-overview.js:  publishTarget: oneWay('session'),
app/components/learnergroup-instructor-manager.js:const { oneWay } = computed;
app/components/learnergroup-instructor-manager.js:  availableInstructorGroups: oneWay('learnerGroup.cohort.programYear.program.school.instructorGroups'),
app/components/new-myreport.js:const { oneWay } = computed;
app/components/new-myreport.js:  value: oneWay('model.id'),
app/components/user-search.js:const { oneWay } = computed;
app/components/user-search.js:  sortTerm: oneWay('content.title'),
app/components/user-search.js:        sortTerm: oneWay('content.fullName'),
```

PR #3501, opened 2018-01-29
converted {{one-way-search}} components to markup equivalents.

stopfstedt:
refs #3466

### DIFF
--- a/app/templates/components/ilios-users.hbs
+++ b/app/templates/components/ilios-users.hbs
@@ -1,10 +1,11 @@
 <div class="filters">
   <div class="filter user-search">
-    {{one-way-search
-      value=query
-      update=setQuery
-      placeholder=(t 'general.searchUsers')
-    }}
+    <input
+      type="search"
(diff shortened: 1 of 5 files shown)


jrjohnson:

Built with commit 261f3808e555027b8b480bced0117ae771d63c35

https://deploy-preview-3501--ilios-frontend.netlify.com

app/templates/components/ilios-users.hbs
```diff
@@ -1,10 +1,11 @@
 <div class="filters">
   <div class="filter user-search">
-    {{one-way-search
-      value=query
-      update=setQuery
```
jrjohnson: Why is this action going away? If I'm reading `users.hbs` right setting the query also changes the URL so it needs to happen at a higher level.

stopfstedt: ok, i missed that URL changing part. taking it back to the workbench. thanks!

PR #3505, opened 2018-01-29
one-way-controls replacements in new user form.

stopfstedt:
refs #3466 

this form here:

selection_455

### DIFF
--- a/app/components/new-user.js
+++ b/app/components/new-user.js
@@ -58,4 +58,23 @@ const Validations = buildValidations({
 
 export default Component.extend(NewUser, Validations, {
   classNames: ['new-user'],
+
+
+  keyUp(event) {
+    const keyCode = event.keyCode;
+    const target = event.target;
+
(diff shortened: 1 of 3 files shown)


jrjohnson:

Built with commit b33b45895688182e0e43a2e8f1c9579153bbe1ef

https://deploy-preview-3505--ilios-frontend.netlify.com

stopfstedt:
^ validation works for me, mine looks exactly like this on blank form submission. @dartajax how'd you bypass validation?

dartajax:
Well when you tab through, after entering something and then deleting it, it should give the message. #3505 gives the message but only after you click `Done` - in my example from demo3, I never clicked `Done`. I just tabbed through and as you navigate through the form, you are informed what's missing.

dartajax:
^ so you know as a user what's missing before clicking `Done` - that other way is fine too though. It's just not completely identical.

PR #3524, opened 2018-02-05
Import small group tool from microapp into Ilios

### LINKED ISSUE #3496: Bulk group membership tool
Users need a way to assign many users to many groups from a spreadsheet. We would add an uploader to each learner group where users could be attached.

This would integrate https://github.com/ilios/small-group-tool

Call the button "Bulk Assign"

jrjohnson:
Fixes #3496

wip:
- [x] Translations
- [x] Loading indicator after the file is selected, but before it is parsed
- [x] When creating a new group send the cohort data


jrjohnson:

Built with commit 955b695a4b64ae1763f33c305b7146b568d44bb7

https://deploy-preview-3524--ilios-frontend.netlify.com

dartajax:
Why is there no save button at the bottom?

image

Everything was matched up so there should have been a way to process the save.

dartajax:
If the spreadsheet contains the matching sub-group, that should be the default.

dartajax:
When a new spreadsheet is uploaded, the previous listing and matchups are not cleared.

dartajax:
Better layout idea - table with rows that need work (ie. sub group name in spreadsheet doesn't match with sub group name in the database) have yellow (Ilios gold) highlighting. 

Rows that match should default to the matching value (still have the option of changing) and they should be highlighted in green - good to go.

jrjohnson:
@dartajax I believe I have addressed these issues, please test again.

PR #3603, opened 2018-03-12
Convert Acceptance Tests to a new API

jrjohnson:
As of Ember 3.0 there is a new way to setup tests which offers some significant improvements in organization for us such as the removal of globals, the automatic running of initializers, and the ability to group tests into modules.

Because acceptance tests all rely on some of the same helpers and setup functionality I updated them all at once. Smaller PRs should be possible for our component tests, but splitting this one up would have been difficult and risked leaving us in a perpetual state of unfinished migration.

Most of these are just small structure updates for the tests with some exceptions:
1. Courses and Sessions tests have be completely re-written using the page object pattern. This was necessary to get around all kinds of weird async issues I was seeing originally.
2. The time picker was re-written to be more testable and reliable
3. Course and session objectives templates were modified to be more testable
4. The offering list template was modified to consolidate some duplication that was causing testing issues.

Otherwise the only changes to the application were to add `data-test-*` markers to identify specific elements for testing.


jrjohnson:
@dartajax can you click test 1,2,3 sections listed above?

jrjohnson:

Built with commit ece1e4d04d7c103065e77b5715b6862c7b8eda36

https://deploy-preview-3603--ilios-frontend.netlify.com

dartajax:
I doubt this is part of this but it might be. When you want to add a Parent Objective to a Course Objective, the screen routes you down to Session Objectives or just kind of to nowhere. It's hard to tell where and how to select Parent Objectives.

PR #3605, opened 2018-03-12
Lint before committing

jrjohnson:
Avoid issues with committing poorly formatted JS or SCSS by linting
before each commit.  Takes about 15 seconds, but it is worth it I think.

This can always be skipped with `--no-verify`.


jrjohnson:

Built with commit 52c8aa65348ad182b464f0eff92f8ee3e42742ea

https://deploy-preview-3605--ilios-frontend.netlify.com

PR #3626, opened 2018-03-19
Drop babel polyfill

jrjohnson:
We needed this to support async / await and other promise syntax, but
this is handled for us by ember-maybe-regenerator now so we can drop the
polyfill and save some file size.

Needs to be tested in all of our supported browsers:
- [x] Chrome 64
- [x] Firefox 58
- [x] Safari11
- [x] Edge16
- [x] iOS Safari11.0-11.2
- [ ] Firefox52 
- [x] Opera 51


jrjohnson:

Built with commit ee160af4d47bdc172101740e1177649294ece07a

https://deploy-preview-3626--ilios-frontend.netlify.com

PR #3627, opened 2018-03-20
Add stylelint explicitly

jrjohnson:
This ends up getting installed from ember-cli-stylelint, but we're
getting a lot of errors because yarn isn't sure we have it. Let's just
add it as a dependency and take away all doubt.


jrjohnson:

Built with commit 67eb68ecd7c91c2d6febf51a629e271b63c0daa4

https://deploy-preview-3627--ilios-frontend.netlify.com

PR #3639, opened 2018-03-26
Improve our unsupported browser error experience

### LINKED ISSUE #3311: Need To Display A Message To Users with Outdated Browsers
This could be a message up at the top advising the logged-in user to update the browser they are using. This is not going to be a link to upgrade the browser -- just an informative message based on our definition of supported browsers at that time.

jrjohnson:
Instead of just waiting to display the error message we can display it
as soon as we trap a runtime exception. Since Ember will trap these for
us once the app is booted this will only show up for pre-load errors.

Builds the list of support browsers from our targets.js file and
displays them when there is an error.  This isn't the same as our
official support matrix, but it is dynamic and will always be correct.

Fixes #3311

Example of the error page:
ie11-error


jrjohnson:

Built with commit 8f5bf6508fde069e9669f614776997234f2c75d8

https://deploy-preview-3639--ilios-frontend.netlify.com

PR #3660, opened 2018-04-06
layout fix for user search results

### LINKED ISSUE #3656: Search Results for Learner Group Default Instructor Not Being Displayed
The search does appear to run properly. When the botched class is removed, the results can be seen.

Right now, Default Instructors cannot be selected for Learner Groups as a result of this.

image

stopfstedt:
fixes #3656 

selection_011

side-effect of this change is that search results are no longer shrink-wrapped and will claim the entire available width of their containing element. Example from the program year director search:

selection_010

IMO, this is an acceptable trade-off for the time being.

### DIFF
--- a/app/styles/components/user-search.scss
+++ b/app/styles/components/user-search.scss
@@ -11,7 +11,6 @@
     left: 0;
     max-height: 15rem;
     overflow: scroll;
-    position: absolute;
     top: 2rem;
     transition: all .2s ease-in-out;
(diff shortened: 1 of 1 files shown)


jrjohnson:

Built with commit 7b6df2d409508bf338f735213eae3020bdec0ed2

https://deploy-preview-3660--ilios-frontend.netlify.com

PR #3734, opened 2018-04-27
Remove all associations from mirage factories

jrjohnson:
These auto create the related entities and can cause mysterious test
failures, we shouldn't be relying on them so lets just take them out.


jrjohnson:

Built with commit 209fc3eeb22d9ca280a0a3b03203dc8fbb112833

https://deploy-preview-3734--ilios-frontend.netlify.com

PR #3738, opened 2018-04-28
Remove remaining schools references

jrjohnson:
Found a few more places where we were using school references that required a permissions lookup that no longer exists.
Wip:
- [x] Validate permission to use cohort
- [x] No need to traverse schools to cohorts if we're going to display them all.

### DIFF
--- a/app/components/detail-cohort-manager.js
+++ b/app/components/detail-cohort-manager.js
@@ -1,17 +1,16 @@
 /* eslint ember/order-in-components: 0 */
 import { inject as service } from '@ember/service';
 import Component from '@ember/component';
-import RSVP from 'rsvp';
+import { all, map } from 'rsvp';
 import EmberObject, { computed } from '@ember/object';
 
 import { translationMacro as t } from "ember-i18n";
 
 const { sort } = computed;
-const { all, Promise } = RSVP;
 
 export default Component.extend({
   i18n: service(),
-  currentUser: service(),
+  store: service(),
   init() {
     this._super(...arguments);
     this.set('sortBy', ['title']);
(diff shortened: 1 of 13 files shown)


jrjohnson:

Built with commit bc13fbe8654fd61bd55cb21629c40764250c9c4f

https://deploy-preview-3738--ilios-frontend.netlify.com

jrjohnson:
@saschaben there is one questionable part of this change. Who should be allowed to attach a cohort to a school? Do you need to have permission to update a course in that school? To update the cohort you are adding? Or just to update the course you are adding the cohort to?

saschaben:
@jrjohnson I'm going to assume^ that when you wrote 
> `Who should be allowed to attach a cohort to a `school`?` 

you *actually* meant 
> `Who should be allowed to attach a cohort to a `course`?`

If that is the case, then I think the key should be that someone must have the permission to do at least one of the following:
- manage a course in that school;
- manage learner groups in that school;
- manage cohorts in that school;
- manage the course in question

^ and if my assumption is wrong to begin with -- let me know.

jrjohnson:
Yes, I did mean course not ~school~. 
But now I'm confused about which school you meant. Let's call them `course` and cohort-in-another-school. I think cohorts in the same school as the course should be allowed for anyone with the permission to update the course.  

So if I'm reading your answer correctly then when we present a list of cohorts which can be attached to that school we should included:
1) Cohorts in the same school as the course
2) cohorts-in-another-school where the user has permission to manage all courses.
3) cohorts-in-another-school where the user has permission to manage that cohort

They will (by default) need to be able to manage the course otherwise they can't attach anything.

saschaben:
i think we are aligned on this. make it so.

PR #3755, opened 2018-04-30
Rewrite new program year select

### LINKED ISSUE #3753: Program Year Selector Not Keeping Correct Value on the Screen But DOES Save Correctly
I am using stage, which has the new permission code on it - just as an FYI. This does NOT occur on ilios3-stage, which does not have the new code on it ... yet.

image

When the user tries to select 2021-2022, the drop-down reverts to having the value at the top populated - in this case, 2016-2017.

However, I just noticed if you click 2021-2022 a second time, the drop-down keeps the correct value. Either way, it will process the save for the new 2021-2022 Program Year.

jrjohnson:
Wasn't always behaving as intended so I made everything more explicit.

Fixes #3753 

refs ilios/ilios#1511

### DIFF
--- a/app/components/new-programyear.js
+++ b/app/components/new-programyear.js
@@ -1,23 +1,27 @@
 /* eslint ember/order-in-components: 0 */
 import Component from '@ember/component';
-import { isEmpty } from '@ember/utils';
+import { computed } from '@ember/object';
 import { task } from 'ember-concurrency';
 
 export default Component.extend({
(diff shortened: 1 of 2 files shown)


jrjohnson:

Built with commit dc200b2aa61b6f84d1d3ef1dae66a5e60d2a5d76

https://deploy-preview-3755--ilios-frontend.netlify.com

dartajax:
I tested side by side - this is much better.

PR #3774, opened 2018-05-02
Prevent terms with associations from being deleted

### LINKED ISSUE #3773: Terms Should Not Be Deletable If They Are Attached To Courses and Sessions
A term (topic) such as "Anesthesiology" has been used plenty. 

image

The trash can - allowing for its deletion should not be showing up here.

image

I tested and in prod currently, it looks like this (as it should appear) ...

_NO_ trash can..

image

Related to ilios/ilios#1511

jrjohnson:
Logic issue here was *only* allowing terms with associations to be
deleted.

Fixes #3773

### DIFF
--- a/app/templates/components/school-vocabulary-term-manager.hbs
+++ b/app/templates/components/school-vocabulary-term-manager.hbs
@@ -31,7 +31,7 @@
       {{else}}
         {{title}}
       {{/if}}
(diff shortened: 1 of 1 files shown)


jrjohnson:

Built with commit 05253b5e676be6b27da692c2320560439fcf2d6d

https://deploy-preview-3774--ilios-frontend.netlify.com

PR #3785, opened 2018-05-03
Improve controls when a session type can not be edited

### LINKED ISSUE #3781: Edit Buttons and Links Should Not Be Available When Session Type Editing is Not Allowed
image

image

I was logged in as a user who should not have been able to get to the Edit Session type detail on Schools. Clicking the `Done` button causes an ugly error. This one ...

image

To avoid that, let's not even allow them to get there.

Refer to ilios/ilios#1511

jrjohnson:
In this case instead of Save and Cancel we just have a Close button.

Fixes #3781


jrjohnson:

Built with commit 3d90e2252e1eeca0d3f91049b15984954b9aa485

https://deploy-preview-3785--ilios-frontend.netlify.com

PR #3792, opened 2018-05-07
Update worker when user logs in

jrjohnson:
When a user first logs in if there is an update to the service worker
(and thus the application) we should take that chance to update it.


jrjohnson:

Built with commit 66925fb4ed31d89d22d034ee519b2b60474bacc8

https://deploy-preview-3792--ilios-frontend.netlify.com

PR #3826, opened 2018-05-16
list ILMs in sequence block sessions list and manager, but keep them them not selectable.

### LINKED ISSUE #3825: ILMs not included in CI sequence block session list and manager
fix this omission.

ILMs should be listed, but be their "count as one" checkbox should be disabled. bulk select-all/unselect-all should ignore ILMs accordingly.

ILMs should also be indicated as such by suffixing their titles with **(ILM)**.

Edge case 1: session that were linked and later converted to non-ILM should retain their state. they can neither be selected, nor unselected.

Edge case 2: ILMs that have offerings. At this time, they will not be selectable. Will deal with them in a follow-up ticket as they need to be special treatment in the backend as well.

stopfstedt:
fixes #3825

### DIFF
--- a/app/components/curriculum-inventory-sequence-block-session-list.js
+++ b/app/components/curriculum-inventory-sequence-block-session-list.js
@@ -4,7 +4,7 @@ import { computed } from '@ember/object';
 import { task } from 'ember-concurrency';
 
 export default Component.extend({
-  linkedSessions: null,
+  sessions: null,
   linkableSessionsBuffer: null,
   classNames: ['curriculum-inventory-sequence-block-session-list'],
   tagName: 'section',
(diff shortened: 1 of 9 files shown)


jrjohnson:

Built with commit fef90f18c1c895f7add31a365b873c149e37ee97

https://deploy-preview-3826--ilios-frontend.netlify.com

app/templates/components/curriculum-inventory-sequence-block-session-list.hbs
```diff
@@ -38,7 +38,12 @@
             <td class='text-center' colspan=2>
               {{if (is-in linkedSessions session) (t 'general.yes') (t 'general.no')}}
             </td>
-            <td class='text-left' colspan=3>{{session.title}}</td>
+            <td class='text-left' colspan=3>
+              {{session.title}}
+              {{#if session.isIndependentLearning}}
+                <strong>({{t 'general.ilm'}})</strong>
```
saschaben: After a quick visual review, I think we want to prepend, rather than append the `(ILM)` label to the session title. This will make it visually a little easier to parse.

stopfstedt: on it.

saschaben:
Looks good -- just the right amount of cowbell.

app/components/curriculum-inventory-sequence-block-session-list.js
```diff
@@ -4,28 +4,28 @@ import { computed } from '@ember/object';
...
+  sessions: null,
   linkableSessionsBuffer: null,
   classNames: ['curriculum-inventory-sequence-block-session-list'],
   tagName: 'section',
   sortBy: 'title',
 
   init() {
     this._super(...arguments);
-    this.set('linkedSessions', []);
+    this.set('sessions', []);
```
jrjohnson: Is this needed?  It's going to just end up being the value passed in right?

stopfstedt: yes, it's expected to be passed in. i'll see what happens when i yank it.

stopfstedt: this checks out w/o initializing. i removed that line.

PR #3854, opened 2018-05-31
Update worker immediately when there is an API update

jrjohnson:
The utility we were loading here doesn't seem to exist, but isn't
necessary we can just ask the worker if it has an update.


jrjohnson:

Built with commit fbb21c8be554eca10f7130f47e4cb041e04c65e3

https://deploy-preview-3854--ilios-frontend.netlify.com

PR #3876, opened 2018-06-19
Display full sized session table

### LINKED ISSUE #3861: Return Session List to Its Previous Larger Size
We now have a situation like the one below ...

image

The course has seven sessions. For any course with fewer than 10 sessions, we changed to use a smaller box. 

Only three of the sessions are immediately viewable. The user has to scroll or filter to see the lower four sessions.

The normal sized (previous) box looks like the following with enough room for approximately 13 sessions.

image

jrjohnson:
Even when there are only a few sessions we should display a full sized
session block - otherwise it is confusing where a user should scroll.

Fixes #3861

### DIFF
--- a/app/components/session-table.js
+++ b/app/components/session-table.js
@@ -99,13 +99,6 @@ export default Component.extend({
 
     return columns;
   }),
-
-  height: computed('sessions.length', function(){
-    const sessions = this.get('sessions');
-    const count = sessions?sessions.length:0;
-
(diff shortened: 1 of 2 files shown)


jrjohnson:

Built with commit d520b8ad8a50e2ce1bc71c8f799a361afff471fb

https://deploy-preview-3876--ilios-frontend.netlify.com

PR #3894, opened 2018-06-28
Test the accessibility of our locale chooser

jrjohnson:


jrjohnson:

Built with commit 9cd6ebba8bdd78e5977d2ea570d2f57dc290d8cc

https://deploy-preview-3894--ilios-frontend.netlify.com

jrjohnson:
@saschaben turns out my first pass on this component wasn't completely accessible - in order to pass tests I needed to change the color of the text when the language chooser is open from orange to grey so it would contrast enough with the background. Can you take another look and let me know if that is OK? I'm guessing we will need to revisit this choice with the many places we're using orange against a white background so I wanted to get some consensus on this low impact area.

saschaben:
I don't have an issue with this, but yes -- we'll need to review the application for consistency.

PR #3902, opened 2018-07-06
Re-write user menu to meet accessiblity guidelines

jrjohnson:
Accounts for keyboard navigation and screen reader users.

### DIFF
--- a/app/components/ilios-header.js
+++ b/app/components/ilios-header.js
@@ -1,10 +1,9 @@
 /* eslint ember/order-in-components: 0 */
-import { inject as service } from '@ember/service';
 import Component from '@ember/component';
+import { inject as service } from '@ember/service';
 
 export default Component.extend({
   session: service(),
-  currentUser: service(),
   classNames: ['ilios-header'],
   tagName: 'header',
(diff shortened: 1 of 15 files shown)


jrjohnson:

Built with commit 417da8150db06b4decaa8694d5b2ce81776ef752

https://deploy-preview-3902--ilios-frontend.netlify.com

jrjohnson:
@saschaben in order to meet color contrast guidelines I needed to put the menus into a box, it turns out nothing works against our orange. Please give this a look and let me know what you think.

saschaben:
if we need to use black for the font in these two menus, we should probably modify the icons to match:



Also -- is it possible to reverse the colors for onMouseOver, so that white is the hover and grey is the default? That might make it a bit less glaring in the layout when unselected.

are there any other areas affected by this PR?

saschaben:
current rendition appears to have lost the `border-radius` setting on the menus....

jrjohnson:
Good catch @saschaben - borders are hard! Fixed now.

saschaben:
:+1:

PR #3910, opened 2018-07-11
Notify users when a new version is available

jrjohnson:
If a user leaves their session open for a long time they might want to
know that a new version of ilios is available to use. This notifies them
by polling the server to see what our latest build is.

I used the same UI we use when a background service worker update is detected. These are different cases, but users will see them as the same and the UI should be the same.


jrjohnson:

Built with commit 9a993f2ecb7d77d8a0e148ce984ca33d42006785

https://deploy-preview-3910--ilios-frontend.netlify.com

PR #3926, opened 2018-07-19
copy session into the selected course.

### LINKED ISSUE #3925: Copy Session Copies Session Only Into Course at Top of The List
This is a bug.

How to reproduce ...

Find a session you wish to copy ...

image

This copied session will NOT land in the correct course. It will land in whatever course happens to be at the top of the drop-down list.

stopfstedt:
fixes #3925.

### DIFF
--- a/app/templates/components/session-copy.hbs
+++ b/app/templates/components/session-copy.hbs
@@ -28,7 +28,7 @@
     <label>{{t 'general.targetCourse'}}:</label>
     {{#if (is-fulfilled courses)}}
       {{#if (get (await courses) 'length')}}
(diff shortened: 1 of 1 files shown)


jrjohnson:

Built with commit c9ae0013620251300f3915846b05e18522dc52e9

https://deploy-preview-3926--ilios-frontend.netlify.com

dartajax:
Interesting. It seems to work fine. I didn't recall that as a user, you are routed directly to the new session location after copying. Is that the correct behavior? If so, 👍

dartajax:
This is probably another ticket but the default year should be the same year as the session you are trying to copy - the default is to copy back into 17-18 from 18-19 - less than ideal.

PR #3928, opened 2018-07-19
correct display of "root" attribute in user profile admin console

### LINKED ISSUE #3899: Root user status not properly reflected on users' profile page
When a user is designated as `root` in the `user` table, their status is not correctly reflected on their respective profile page.  

The users' `Root` status reads `Root: No`, regardless of their status in the database.  However, their abilities DO reflect their elevated status, so this seems to just be a frontend/display issue.

stopfstedt:
fixes #3899

### DIFF
--- a/app/templates/components/user-profile-roles.hbs
+++ b/app/templates/components/user-profile-roles.hbs
@@ -91,6 +91,6 @@
   <hr>
   <div class='item'>
     <label>{{t 'general.root'}}:</label>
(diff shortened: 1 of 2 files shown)


jrjohnson:

Built with commit 7fcc963e8518fe90f18ed3e0bcec7c30f3d8ac2e

https://deploy-preview-3928--ilios-frontend.netlify.com

PR #3939, opened 2018-07-23
root-level users can work with inactive users in learner group manager.

### LINKED ISSUE #3912: Allow root user to manage disabled user accounts within the UI
Sometimes it is necessary to remove learners from learner groups even if the user's account is disabled.

It might be easier to allow root users to perform this function rather than having to edit the records directly in the database or temporarily enable the account. These are the current work arounds.

stopfstedt:
fixes #3912 

selection_066

selection_067


jrjohnson:

Built with commit 8c8caa09cd626aadff4d796373fd6a75cf680574

https://deploy-preview-3939--ilios-frontend.netlify.com

PR #3954, opened 2018-08-02
Deploy archived assets

jrjohnson:
Requires ilios/ilios#2198

Use `ember-cli-deploy-compress` to compress our assets during the build process. We were doing this on the API which is extra work for no reason. Doing it here makes us more flexible and puts the responsibility in the right hands.


jrjohnson:

Built with commit 218400650b4fb1acd4c448d5cccdf70e03ec4ec5

https://deploy-preview-3954--ilios-frontend.netlify.com

PR #3983, opened 2018-08-22
Indicate that all sessions are expanding

### LINKED ISSUE #3978: Spinner Needed On New Sessions List When Expanding or Contracting All Sessions
image

When a user clicks as shown above, the screen freezes but there is no indication that the sessions are all being loaded in the expanded view. We should have some indicator like a spinner or something else that there is something happening to advise the user change is afoot.

jrjohnson:
In big courses it can take a while for all of the sessions to open. What
I'm doing here is starting a loading indicator and then opening the
sessions. The browser takes a while to render the newly exposed DOM so
we use Ember.run.next() to wait for control to come back from that
process and then kill the spinner.

Fixes #3978

### DIFF
--- a/app/components/sessions-grid-header.js
+++ b/app/components/sessions-grid-header.js
@@ -1,8 +1,11 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
+import { task, timeout } from 'ember-concurrency';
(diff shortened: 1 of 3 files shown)


dartajax:
Love it.

PR #3998, opened 2018-08-27
Add Instructional Notes to Sessions

### LINKED ISSUE #2609: Instructional Notes for Session
Currently at the session level, we do not provide for any descriptive information which is displayed only to faculty and not to students, along the lines of the instructional notes for learning materials when set to "hide". We should consider adding such a field to allow for session-general instructional information to be added (that is not LM-specific.)

- Would require an additional field attribute to be added; 
- would NOT require a toggle (student-facing information will always go in the description field);
- permission-based access to view for instructors;
- would display in the session manager page ;
- would display on calendar event view for instructors only

jrjohnson:
Administrators can add special instructional notes to each session.

Fixes #2609 

- [x] ~Fix issue with the editor not taking up enough space when opened.~ #4001

### DIFF
--- a/app/components/course-sessions.js
+++ b/app/components/course-sessions.js
@@ -50,6 +50,7 @@ export default Component.extend({
         canUpdate,
         id: session.get('id'),
         title: session.get('title'),
+        instructionalNotes: session.get('instructionalNotes'),
         isPublished: session.get('isPublished'),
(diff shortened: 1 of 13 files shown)


dartajax:
Would it be possible to have the stars fixed and the notes icon to the left when applicable?

image

PR #4022, opened 2018-09-04
display error message on login form if user account does not exist

stopfstedt:
fixes ilios/ilios#2219.


stopfstedt:
needs click testing in shibb'ed environment.

PR #4024, opened 2018-09-05
Improve experience when the API version is wrong

### LINKED ISSUE #4023: API Mismatch Message Should Never Be Shown To Users
When a user hasn't visited Ilios in a while, a message appears on the screen advising that there is an API mismatch. A while later (10-15 seconds maybe), a Huzzah we have made Ilios better message appears. The user then is able to get the API mismatch issue resolved. The process is confusing and less than optimal.

jrjohnson:
Users sometimes see this screen if the service worker doesn't update
before the component renders. I've added a few seconds of delay before
the warning is shows, but in case a user still sees this overlay I've
also improved the experience to start an auto countdown and re-load when
the new worker is ready.

Fixes #4023


app/components/api-version-check.js
```diff
@@ -1,27 +1,58 @@
...
+import { task, timeout } from 'ember-concurrency';
 
 const { apiVersion } = ENV.APP;
 
 export default Component.extend({
   iliosConfig: service(),
   versionMismatch: false,
+  updateAvailable: false,
+  countdownToUpdate: null,
+  updatePending: false,
```
stopfstedt: initialize `noUpdateAvailable` attr.

PR #4045, opened 2018-09-11
use beta Chrome instead of stable, upgrading us out of build errors.

stopfstedt:
next attempt.

refs #4046


stopfstedt:
this fails even faster. next!

PR #4061, opened 2018-09-15
Greenkeeper/ember data 3.4.2

jrjohnson:
Supersedes #4060

### DIFF
--- a/app/components/course-objective-manager.js
+++ b/app/components/course-objective-manager.js
@@ -4,10 +4,9 @@ import Component from '@ember/component';
 import { isEmpty } from '@ember/utils';
 import EmberObject, { computed } from '@ember/object';
 import ObjectProxy from '@ember/object/proxy';
-import RSVP from 'rsvp';
+import { all, map, Promise } from 'rsvp';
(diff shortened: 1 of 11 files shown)


app/components/course-objective-manager.js
```diff
@@ -106,28 +105,44 @@ export default Component.extend({
     }
   }),
 
+  course: computed('courseObjective.courses.[]', async function () {
+    const courseObjective = this.get('courseObjective');
+    if (! courseObjective) {
+      return [];
```
stopfstedt: kind of a weird return value for "does not exists". may i suggest `null` instead?

jrjohnson: I didn't want to mess with the existing functionality too much, took this from https://github.com/ilios/frontend/blob/master/app/components/course-objective-manager.js#L117.  I had a note to myself to refactor after they are moved to common, but I just created an issue for more general use https://github.com/ilios/frontend/issues/4076

stopfstedt: fair enough. empty array evaluates falsy, so this should be alright. let's fix it later then.

PR #4083, opened 2018-09-27
Check for an existing worker before re-loading

jrjohnson:
Hopefully this solves the issues where visiting ilios for the first time would sometimes cause the page to refresh as soon as it was loaded.


dartajax:
It didn't work exactly the way @jrjohnson was thinking it would work. It splashed, logon screen, and then splashed again.

PR #4095, opened 2018-10-01
Update Dependencies

jrjohnson:
The update to terser-js is causing problems with ember builds. The
recommended workaround is to disable collapse_vars and I've done that
here.

This doesn't seem to have made our build any larger or anything so I don't see any harm at this point.

There *is* an exciting improvement in here that should make our builds faster with the elimination of node-sass, but that means this should get some extra click testing in case our styles have been messed up somehow.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -1,4 +1,5 @@
 'use strict';
+/* eslint camelcase: 0 */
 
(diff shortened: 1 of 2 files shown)


dartajax:
Click test passed.

PR #4193, opened 2018-12-03
Update ember-cli and blueprints to v3.5.1

jrjohnson:
Also updated all transitive dependencies


jrjohnson:
We need to wait for a new version of ember-simple-auth which support ember-fetch 6.

PR #4230, opened 2019-01-04
improved new school form state handling

### LINKED ISSUE #4228: School Administrator Email Value Needs To Clear After Save
I just added the school called `Use Later` and then clicked the Add New button to add another school. It is clear that the value in the Administrator Email field `dartajax@gmail.com` is still there from adding the previous school. After save, the value should be removed.

If anybody other than me ever finds this bug, that person will receive the DartAjax box set for free.

image

### LINKED ISSUE #4229: Only The Most Recently Added School Should Remain in the Celebratory Message
This is another "free DartAjax box set" bug if anyone other than HeWhoShouldNotBeNamed finds it. I just added the school `Use Now` - previously added was `Use Later`

image

stopfstedt:
fixes #4228 
fixes #4229

### DIFF
--- a/app/components/school-list.js
+++ b/app/components/school-list.js
@@ -29,13 +29,12 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
   store: service(),
   init(){
     this._super(...arguments);
-    this.set('newSchools', []);
     this.set('sortSchoolsBy', ['title']);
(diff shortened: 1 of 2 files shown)


dartajax:
Both tests (manual style) passed.

PR #4272, opened 2019-01-27
Turn off code coverage in deploys

### LINKED ISSUE #4271: Error when visiting single user page
The change in #4094 is not working and is causing an issue when attempting to load the single user page like https://ilios3-demo.ucsf.edu/users/1000

jrjohnson:
There appears to be a conflict between code coverage and dynamic auto
imports that is causing our deployed frontend to be missing some files.
This should fix it.

Fixes #4271


jrjohnson:
Going to merge this myself to verify that a staging job built with travis will run.

PR #4285, opened 2019-01-30
async/await throughout. access component props directly.

stopfstedt:
forgot to check this in earlier.

### DIFF
--- a/app/components/school-competencies-expanded.js
+++ b/app/components/school-competencies-expanded.js
@@ -39,6 +39,9 @@ export default Component.extend({
     return competencyIds.length && ! isManaging;
   }),
 
+  /**
+   * @todo rewrite the component so we don't deal with promises in this lifecycle hook. [ST 2019/01/30]
+   */
   async didReceiveAttrs(){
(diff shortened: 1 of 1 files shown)


app/components/school-competencies-expanded.js
```diff
@@ -39,34 +39,32 @@ export default Component.extend({
     return competencyIds.length && ! isManaging;
   }),
 
-  didReceiveAttrs(){
+  async didReceiveAttrs(){
```
jrjohnson: We're overriding something on EmberObject here and I'm not sure it can be async. I guess javascript lets us, but nothing is going to resolve this promise so I'm not sure this is a good idea.

stopfstedt: i agree, but we're using promises already in this method already. switching to async/await syntax makes it more visible, imo. will throw a `@todo` comment in there. the whole thing will need to be re-done at some point, i'm afraid.

jrjohnson: Agreed.

PR #4333, opened 2019-02-17
Improve our error reporting

jrjohnson:
The default of VERSION-tag isn't working for our sentry github
integration so instead this just sends the tag name.  I've also
configured travis to upload source maps and commit data to sentry to
better attach it to the release itself.  Since our production sourcemaps
are often behind an SSO firewall this ensures the sourcemap is attached
to the error.


jrjohnson:
Hard to test this stuff since it only happens in production. @stopfstedt can you scan for any obvious errors in my travis bash?

PR #4343, opened 2019-02-20
Move from yarn to NPM

jrjohnson:
This is the ember default and now that NPM supports a lock file it's a
good choice for us. It has a larger overall user base, is installed by
default with node, and is getting improved all the time.


jrjohnson:
Greenkeeper failure here is expected. Once it's merged I'll need to change our setup to build with npm.

PR #4359, opened 2019-02-22
replace regexp pattern matching with simple in-string search.

### LINKED ISSUE #4351: replace regex matching with simple string containment checks where possible
follow-up item to #4348, see conversation thread for details there.

stopfstedt:
fixes #4351 

affects:

- user search in admin interface
- user search in "pending updates" page of admin interface
- user search in "unassigned students page of admin interface
- search in courses lists page
- search in instructor groups list page
- search in learner groups list page
- my materials search
- user search in learner groups assignment page

### DIFF
--- a/app/components/learnergroup-cohort-user-manager.js
+++ b/app/components/learnergroup-cohort-user-manager.js
@@ -4,7 +4,6 @@ import Component from '@ember/component';
 import { computed } from '@ember/object';
 import { isEmpty } from '@ember/utils';
 import { task, timeout } from 'ember-concurrency';
-import escapeRegExp from 'ilios-common/utils/escape-reg-exp';
 
 export default Component.extend({
   currentUser: service(),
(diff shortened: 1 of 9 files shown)


jrjohnson:
How does this work with spaces? Does `Jon Johnson` return results when it matches a bit of first and a bit of last? Did it return results before? It might be that it never worked or that it works now and I don't know which. Maybe a test for that?

stopfstedt:
> How does this work with spaces? Does `Jon Johnson` return results when it matches a bit of first and a bit of last? Did it return results before? It might be that it never worked or that it works now and I don't know which. Maybe a test for that?

checks out for me. i did not spot any regressions. example:

selection_317


@dartajax  can you poke at this for a bit?

jrjohnson:
That search box wasn't changed here it's over at https://github.com/ilios/frontend/blob/master/app/components/manage-users-summary.js#L20. I just checked the cohort filter and it doesn't work with this, but it also didn't work before so I'm good with this change.

Issue #3997, opened 2018-08-27
FA5 icon consistency


saschaben:
let's make sure the new icons maintain consistency wherever used (and clean up extant inconsistencies along the way!)
- [ ] show disabled trashcans in grey, not black, 
- [ ]  show disabled trashcans in all locations (cf. #3989 )
- [ ]  use blue consistently for edit icon across site
- [ ] copy icons should be green, not black

saschaben:
bumping this, cuz it is still outstanding....

![Ilios Learner Groups 2019-02-19 13-37-51.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5551233e5601aa8f0097f50b/125516c66e82c728ace21e0d46c198932c91c3858ae8ad4da14fb0b33b4932633a44f17b21dcae78ed0f2a584f5e4db1425f4241fdeb9539e6f4632adf0f4fffdb7f5a71c23413faec801cb901e923884ed6b6bc821dc5b96e24f5979d0d6f405147d4a61760e28b41da59ff41e04850a4153761e375367b47900c38.png)

stopfstedt:
start here:

```
$ grep -r 'trash' app
app/templates/components/school-competencies-manager.hbs:        {{fa-icon "trash" class="remove clickable" click=(action remove obj.domain)}}
app/templates/components/school-competencies-manager.hbs:            {{fa-icon "trash" class="remove clickable" click=(action remove competency)}}
app/templates/components/instructorgroup-list.hbs:            <span class="clickable remove" {{action "confirmRemove" instructorGroup}}>{{fa-icon "trash"}}</span>
app/templates/components/school-session-types-list.hbs:                "trash"
app/templates/components/school-vocabulary-term-manager.hbs:        {{fa-icon "trash" class="clickable remove" click=(action "deleteTerm")}}
app/templates/components/curriculum-inventory-report-list.hbs:            <span class="clickable remove" {{action "confirmRemove" report}}>{{fa-icon "trash"}}</span>
app/templates/components/learnergroup-list.hbs:              {{fa-icon "trash" class="disabled"}}
app/templates/components/learnergroup-list.hbs:              <span class="clickable remove" {{action "confirmRemove" learnerGroup}}>{{fa-icon "trash"}}</span>
app/templates/components/school-vocabularies-list.hbs:                    "trash"
[...]

PR #4361, opened 2019-02-22
enforce iconographic consistency

stopfstedt:
refs #3997 
selection_326

selection_327
selection_330

selection_328

selection_329

selection_331

selection_332

selection_334

selection_333


saschaben:
Looking good, but please check learner groups -- not the landing page, but the internal group pages, cf. https://deploy-preview-4361--ilios-frontend.netlify.com/learnergroups/10996.

stopfstedt:
> Looking good, but please check learner groups -- not the landing page, but the internal group pages, cf. https://deploy-preview-4361--ilios-frontend.netlify.com/learnergroups/10996.

thanks for checking and finding this one. fixed it

selection_335

stopfstedt:
> Looks great. This is such a nice UI improvement, makes things look _way_ more polished. Makes me wonder if I should throw a class on every icon upstream in the fontawesome addon. Would that have helped here? Or is it just as easy to list the classes that are being targeted?

i think it's easy enough. but let's talk tomorrow, perhaps i'm not understanding your proposal.

PR #4370, opened 2019-02-25
 Update navigation to be accessible 

jrjohnson:
Removed an unnecessary role and change the colors to meet contrast guidelines.


jrjohnson:
@saschaben please check out this grey main navigation text and see how you feel about it vs our old orange.

PR #4387, opened 2019-03-01
Update to the latest code coverage version

jrjohnson:
This includes a lot of improvements to the way code coverage is
instrumented and allows us to output in new formats.


jrjohnson:
OK to merge, slight change in coverage stats to be expected as it more accurately processes our app.

PR #4423, opened 2019-03-13
adds preferred email to user admin form.

### LINKED ISSUE #4376: Provide for a "Preferred Email" field
Add a 2ndary email field for preferred email. This will be editable via the MyProfile page by the user, as well as from the admin profile page. When the preferred email is populated, it will be used for all communication from the system, and will not be updated by the EDS sync

stopfstedt:
fixes #4376

### DIFF
--- a/app/components/user-profile-bio.js
+++ b/app/components/user-profile-bio.js
@@ -47,6 +47,14 @@ const Validations = buildValidations({
       type: 'email'
     }),
   ],
+  preferredEmail: [
+    validator('length', {
+      max: 100
+    }),
+    validator('format', {
+      type: 'email'
+    }),
+  ],
   phone: [
     validator('length', {
       max: 20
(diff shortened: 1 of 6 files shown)


app/components/user-profile-bio.js
```diff
@@ -204,6 +226,7 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
       const firstName = this.firstName;
       const lastName = this.lastName;
       const email = this.email;
+      const preferredEmail = this.preferredEmail;
```
jrjohnson: I don't think this property will every by part of the sync. I'd have to look further, but I don't think we return a full user record from the sync run, it's an object based on what is in LDAP.

stopfstedt: thanks! taking it out.

stopfstedt:
> Can you add a test for removing / empty preferred email? I think the validation for email accounts for blank, but I don't _know_ that it does.

good catch. fixed and backed up with a test.

Issue #4416, opened 2019-03-11
Remove Scheduled As an Option For Program 


dartajax:
related to #3675 

same issue but at the program level

dartajax:
This one can be closed one #4443 is merged.

Issue #3675, opened 2018-04-13
Remove Publishing of Program Years


dartajax:
Since the concept of whether or not a Program Year is not enforced in the Ilios application. Program Years for a published Program are not being verified for publication status. The application treats an unpublished Program Year in the same manner that it treats a published Program Year.

We can just get rid of the drop-down and iconography in the interface entirely regarding the concept of publication of Program Years.

Here is how it currently appears ...

![image](https://user-images.githubusercontent.com/7493331/38760263-2ac4c958-3f2f-11e8-9f6f-72eef98b45a0.png)

The `Status` could be renamed `Action` or `Actions` since it will be possible to lock the Program Year or even delete it if it has not been used yet.

![image](https://user-images.githubusercontent.com/7493331/38760537-b1d7ef5a-3f30-11e8-98e8-c77226a8f831.png)

The drop-down in the upper right corner of this Program Year detail can also be removed since it will no longer be used.

PR #4430, opened 2019-03-15
yanks publication status from programs and program-years.

stopfstedt:
refs #3675 
refs #4416


dartajax:
LGTM - the whole concept of publishing program and program year is gone

stopfstedt:
> We need to have a button available to "publish" or "activate" the legacy "not published" years. "Activate" would work well.
> 
> image

i thought the whole notion of "published" goes away?

stopfstedt:
@dartajax how about a modified publish status dropdown to only include a non-reversible "publish" option, when the current status is in draft/tdb?

stopfstedt:
scrapped.

Issue #3632, opened 2018-03-22
Cleanup didReceiveAttrs in course and school leadership


jrjohnson:
We're using `didReceiveAttrs` to build async relationships on the `program-leadership-expanded` and `course-leadership-expanded` components.  This is a bad idea. If the async returns after the component is destroyed it will cause an error. They should be refactored to match `school-leadership-expanded`

PR #4435, opened 2019-03-18
Refactor school leadership

jrjohnson:
Removes the async actions from `didUpdateAttrs` and puts them in computed
properties.

Refs #3632


jrjohnson:
This does need better tests, but it's going to cause conflicts with changes in common so it will have to wait.

PR #4443, opened 2019-03-22
gracefully deprecate publishing workflow for programs and program-years

### LINKED ISSUE #3675: Remove Publishing of Program Years

### LINKED ISSUE #4416: Remove Scheduled As an Option For Program

stopfstedt:
fixes #4416 
fixes #3675 

Selection_399
Selection_400
Selection_397
Selection_398


dartajax:
When a new Program Year is added, the trash can should be enabled - which it is but only after a page refresh.

stopfstedt:
> When a new Program Year is added, the trash can should be enabled - which it is but only after a page refresh.

you are correct, this restriction needs to go. as discussed, i put that change in.

Issue #4402, opened 2019-03-06
Put Filters and Sorting on Learning Materials Summary - Course Level


dartajax:
The logic behind this is that there are usually _**way**_ more Learning Materials on Sessions and not Courses; however, it would be nice to be able to filter and sort based on the available values at the Course level as well.

Course Level ...

No filtering or sorting ...

![image](https://user-images.githubusercontent.com/7493331/53921919-39d9fb80-4027-11e9-9d62-78a54b98d2f4.png)

Session Level ...

Filtering and sorting ...

![image](https://user-images.githubusercontent.com/7493331/53921955-5aa25100-4027-11e9-9bdc-5d14429fa07e.png)

dartajax:
steps ...

* go to Courses and Sessions 
* Filter as shown below

![image](https://user-images.githubusercontent.com/7493331/54841237-1603f000-4c8c-11e9-8150-acdec5d143e5.png)

* click on the course title "Renal ..."

![image](https://user-images.githubusercontent.com/7493331/54841364-667b4d80-4c8c-11e9-9bee-75eb72b35335.png)

* Below is the top part of the screen showing Course materials - no filtering or sorting is provided - this is what we need.

![image](https://user-images.githubusercontent.com/7493331/54841413-7c890e00-4c8c-11e9-8da6-8c6c99dc3106.png)

* The example of how it should work is shown below (Session materials) - scroll down to see ...

![image](https://user-images.githubusercontent.com/7493331/54841656-02a55480-4c8d-11e9-896b-c2f822c30904.png)

PR #4450, opened 2019-03-25
course-materials component query param flow update

djvoa12:
**Summary:**
Since query params are used for session learning materials sorting, we need to update the query params to handle course learning materials as well.

Accompanies https://github.com/ilios/common/pull/596
Fixes https://github.com/ilios/frontend/issues/4402


jrjohnson:
Throwing a no merge label on this until it is merged and release in common.

Issue #4194, opened 2018-12-03
Method Of Pagination or Longer Scroll Needed in Admin Console Student Assignment


dartajax:
![image](https://user-images.githubusercontent.com/7493331/49406262-ee534100-f709-11e8-8b3d-2f324f3c0f3b.png)

After clicking Manage with 95 students possibly needing assignment ...

![image](https://user-images.githubusercontent.com/7493331/49406326-2490c080-f70a-11e8-8f23-6fca5ba0fc34.png)

There is no way to do a "Show All" at the top and there is no way to navigate to the next page at the bottom of the screen.

stopfstedt:
this is a bug. the list only contains the first 25 records.

fix: display all records.

let's deal with pagination as a separate, follow-up item.

stopfstedt:
to reproduce: log in as a user with elevated privileges, then pick "Admin" from the main navigation. click the "Manage" button in the "Students Requiring Cohort Assignment" pane to get to that screen.

PR #4470, opened 2019-04-04
can see the entire list of students requiring cohort assignment

djvoa12:
**Summary:**
Admins can see the entire list of students requiring cohort assignment, instead of an arbitrary limit of 25 that was not used in any meaningful way. I also refactored many other connecting parts.

**UI:**

Fixes https://github.com/ilios/frontend/issues/4194

### DIFF
--- a/app/components/assign-students.js
+++ b/app/components/assign-students.js
@@ -1,42 +1,30 @@
-/* eslint ember/order-in-components: 0 */
-import { inject as service } from '@ember/service';
 import Component from '@ember/component';
 import { computed } from '@ember/object';
+import { reads } from '@ember/object/computed';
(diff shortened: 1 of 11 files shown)


saschaben:
> if i switch schools, it never loads the new program-years/cohorts for assignment. the loading spinner just says up. assignable users are reloading tho.
> 
> Selection_415

this is where we are looking for an active Program _with an active & current program year_. If there is no active program year, the spinner will not resolve; if the only active program year is in the past, the  spinner will not resolve. This is independent of the bucket of potential students for assignment.

We should probably handle this a bit better.

djvoa12:
@stopfstedt it's the same bug on master as well. It's a different issue.

Issue #4479, opened 2019-04-08
Need empty list and empty search result handlers (various parts)


djvoa12:
Look through the app and create empty list and search result handlers to better notify the users what is going on.

Areas for improvement:
- Courses and Sessions (/courses)
- Instructor Groups (/instructorgroups)
- Programs (/programs)
- Curriculum Inventory (/curriculum-inventory-reports)

PR #4484, opened 2019-04-09
added empty list and search results handlers

djvoa12:
**Summary:**
Added empty list and search results handlers for ...
- Courses and Sessions
- Instructor Groups
- Programs
- Curriculum Inventory

**UI:**
_Empty List:_
Screen Shot 2019-04-08 at 5 25 18 PM

_Empty Search List:_
Screen Shot 2019-04-08 at 5 26 27 PM

Fixes https://github.com/ilios/frontend/issues/4479

### DIFF
--- a/app/components/curriculum-inventory-report-list.js
+++ b/app/components/curriculum-inventory-report-list.js
@@ -1,84 +1,94 @@
-/* eslint ember/order-in-components: 0 */
-import { inject as service } from '@ember/service';
-import { computed } from '@ember/object';
-import ObjectProxy from '@ember/object/proxy';
 import Component from '@ember/component';
-const { alias, not } = computed;
+import ObjectProxy from '@ember/object/proxy';
+import { computed } from '@ember/object';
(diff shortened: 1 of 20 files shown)


dartajax:
image

dartajax:
image

dartajax:
The same is true at least while searching on Courses and receiving the `No Results Found` message - it should be under `Course Title` and not under `Level`

dartajax:
so left justify instead of center alignment would be better I think

djvoa12:
I think it looks way better centered. Left-aligned looks weird. Center would stay true as the screen size collapses.

dartajax:
?
image

dartajax:
However ...

image

It should look more like this in that case ...

image

The label No Results Found should not affect the column headers.

djvoa12:
@dartajax I fixed the responsive columns issue.

stopfstedt:
superseded by #4492

PR #4485, opened 2019-04-09
fixed hanging spinner for assign-students component

djvoa12:
**Summary:**
Fixed hanging loading spinner for cohorts dropdown.

### DIFF
--- a/app/components/assign-students.js
+++ b/app/components/assign-students.js
@@ -24,7 +24,7 @@ export default Component.extend({
   cohorts: reads('loadCohorts.lastSuccessful.value'),
   isLoading: reads('loadCohorts.isRunning'),
 
-  bestSelectedCohort: computed('cohorts', 'primaryCohortId', function() {
+  bestSelectedCohort: computed('cohorts.[]', 'primaryCohortId', function() {
     const cohorts = this.cohorts;
     const primaryCohortId = this.primaryCohortId;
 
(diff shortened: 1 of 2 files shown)


app/components/assign-students.js
```diff
@@ -22,20 +22,22 @@ export default Component.extend({
...
-      if (primaryCohortId) {
-        let currentCohort = cohorts.find(cohort => cohort.id === primaryCohortId);
-        if (currentCohort) {
-          resolve(currentCohort);
-          return;
-        }
-      }
-      resolve(cohorts.get('lastObject'));
-    });
+  bestSelectedCohort: computed('cohorts', 'primaryCohortId', function() {
```
jrjohnson: Should be `cohorts.[]` right?

djvoa12: The promise never resolves as cohorts are empty initially so I put a check to see if cohorts are there or not. I was thinking just `cohorts` because the newly ran task will constantly replace it, instead of adding/removing to an array.

https://github.com/ilios/frontend/blob/master/app/components/assign-students.js#L29

jrjohnson: The next person who touches this code may make changes to this though, and the person after that. They may not realized they need to add `.[]` here. Is there some gain in not identifying the array elements to be tracked?

djvoa12: I don't have a strong opinion at all on this so I am adding it back. 👍

PR #4492, opened 2019-04-12
added empty list and search results handlers

djvoa12:
**Summary:**
Added empty list and search results handlers for ...
- Courses and Sessions
- Instructor Groups
- Programs
- Curriculum Inventory

**UI:**
_Empty List:_
Screen Shot 2019-04-08 at 5 25 18 PM

_Empty Search List:_
Screen Shot 2019-04-08 at 5 26 27 PM

Fixes https://github.com/ilios/frontend/issues/4479

### DIFF
--- a/app/breakpoints.js
+++ b/app/breakpoints.js
@@ -1,0 +1,6 @@
+export default {
+  mobile:  '(max-width: 767px)',
+  tablet:  '(min-width: 768px) and (max-width: 991px)',
+  desktop: '(min-width: 992px) and (max-width: 1200px)',
+  smallScreen: '(max-width: 1200px)'
+};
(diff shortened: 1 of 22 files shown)


app/breakpoints.js
```diff
@@ -0,0 +1,6 @@
+export default {
```
jrjohnson: Please align the names and sizes with https://github.com/ilios/common/blob/master/app/styles/ilios-common/mixins/media.scss and add a comment to this file that the two are linked.  I wonder if we can add this file to common and have it exported. We can investigate that in a different place though.

Looks like a bit more active voice here so `for-laptop-and-up` could just be `laptopAndUp`.

djvoa12: 👍

PR #4540, opened 2019-05-08
removed mymaterials and its contents

djvoa12:
**Summary:**
- Removed `/mymaterials` route and its contents
- ~Removed `ember-page-title` addon dependecy~ needs to remain

Accompanies https://github.com/ilios/common/pull/650


jrjohnson:
Just putting a warning on this that is requires a new version of common before it is merged.

Issue #4541, opened 2019-05-08
My Reports Needs To Maintain State After Following a Report Link


dartajax:
How to reproduce ...

* Login as admin user 
* Go to `My Reports`
* Create a Report for All Courses in School of Medicine
* Click the report title to run it
* Select a Year from the drop-down (2018-1019 will do)
* Scroll down to a Course to review it (about half way down)
* Click on the Course title 
* Click the back button on the browser to return (Don't follow the "Back to Courses" link)
* Notice that the Report has now closed and your selection in the drop-down has been reset along with the scroll position.

PR #4553, opened 2019-05-14
dashboard report menu to preserve state and scroll

djvoa12:
**Summary:**
Dashboard report list now keeps its report-setting state via query params and also preserves its scroll.
- Report ID - `/dashboard?report=1`
- Report Year - `/dashboard?reportYear=2016`

**UI:**
Screen Shot 2019-05-14 at 11 24 44 AM

Fixes https://github.com/ilios/frontend/issues/4541

### DIFF
--- a/app/components/dashboard-myreports.js
+++ b/app/components/dashboard-myreports.js
@@ -42,7 +42,6 @@ export default Component.extend(DomMixin, ReportTitleMixin, {
     const yPos = this.preserveScroll[this.scrollKey];
 
     if (yPos > 0) {
-      const element = document.querySelector('.dashboard-block-body');
       element.scrollTop = yPos;
     }
 
@@ -79,7 +78,7 @@ export default Component.extend(DomMixin, ReportTitleMixin, {
   }),
 
   reportResultsList: computed('selectedReport', 'selectedYear', async function() {
-    const report = await this.selectedReport;
+    const report = this.selectedReport;
     const year = this.selectedYear;
     return report ? await this.reporting.getResults(report, year) : [];
   }),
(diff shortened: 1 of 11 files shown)


app/components/dashboard-myreports.js
```diff
@@ -45,34 +77,34 @@ export default Component.extend(ReportTitleMixin, {
       });
     });
   }),
-  reportResultsList: computed('selectedReport', 'selectedYear', function(){
-    const report = this.selectedReport;
+
+  reportResultsList: computed('selectedReport', 'selectedYear', async function() {
+    const report = await this.selectedReport;
```
jrjohnson: `selectedReport` looks like a value and not a promise, any reason for the `await` here?

tests/integration/components/myreports-list-item-test.js
```diff
@@ -14,13 +14,15 @@ module('Integration | Component | myreports list item', function(hooks) {
       title: 'Lorem Ipsum'
     });
 
-    const reportModel = await this.owner.lookup('service:store').find('report', report.id);
+    const reportModel = await this.owner
```
jrjohnson: This should remain a single line, multi-line is unnecessary here and makes it harder to see where the promise is that is being `await`ed.

jrjohnson:
This looks great! Preserving the scroll state here seems particularly useful. Couple of minor tweaks in code, otherwise looks ready if @dartajax gives it a 👍

app/components/dashboard-myreports.js
djvoa12: Probably an error.

tests/integration/components/myreports-list-item-test.js
djvoa12: Ok

PR #4583, opened 2019-06-02
added placeholder option for course year select

djvoa12:
**Summary:**
New course creation year dropdown has a placeholder.

**UI:**
Screen Shot 2019-06-03 at 12 53 55 PM

Screen Shot 2019-06-03 at 12 54 03 PM

Fixes https://github.com/ilios/common/issues/697

### DIFF
--- a/app/components/new-course.js
+++ b/app/components/new-course.js
@@ -34,21 +34,11 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
   currentYear: null,
   selectedYear: null,
   title: null,
-  years: null,
   cancel() {},
   save() {},
+  years: Object.freeze(YEARS),
 
   isSaving: reads('saveCourse.isRunning'),
-
-  didReceiveAttrs() {
-    this._super(...arguments);
-    this.set('years', YEARS);
-    const currentYear = this.currentYear;
-    const selectedYear = currentYear && YEARS.includes(currentYear.title)
-      ? currentYear.title
-      : THIS_YEAR;
-    this.set('selectedYear', selectedYear);
-  },
 
   actions: {
     setYear(year) {
--- a/app/templates/components/new-course.hbs
+++ b/app/templates/components/new-course.hbs
@@ -23,6 +23,9 @@
     <label>{{t "general.academicYear"}}:</label>
 
     <select data-test-year onchange={{action "setYear" value="target.value"}}>
+      <option disabled hidden selected value="">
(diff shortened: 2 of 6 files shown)


translations/en-us.yaml
```diff
@@ -261,6 +261,7 @@ general:
   permissionsInOtherSchools: "Permissions in Other Schools"
   perPage: "Per Page"
   phone: Phone
+  pleaseSelect: "Please Select"
```
jrjohnson: I like the generic, but everywhere else we have `Select School` or `Select Course` I think this one should be `Select Year`.

app/templates/components/new-course.hbs
```diff
@@ -23,6 +23,9 @@
     <label>{{t "general.academicYear"}}:</label>
 
     <select data-test-year onchange={{action "setYear" value="target.value"}}>
+      <option disabled hidden selected value="">
```
jrjohnson: My reading of the `hidden` attribute says this is invalid: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden thoughts?

It's also invalid to have more than one element selected this causes issues with keyboard navigation it should probably be `selected={{is-empty this.selectedYear}}` though I'm not sure about `is-empty`.

app/components/new-course.js
```diff
@@ -34,22 +34,12 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
   currentYear: null,
   selectedYear: null,
   title: null,
-  years: null,
   cancel() {},
   save() {},
+  years: Object.freeze(YEARS),
```
jrjohnson: ❤️ this. So much better than doing it in a life cycle hook.

translations/en-us.yaml
djvoa12: Fixed.

app/templates/components/new-course.hbs
djvoa12: Fixed.

dartajax:
Functionally, I think this is fine. Is "Select Educational Year" correct as the caption? Should it be "Select Academic Year" instead? Question for @saschaben

stopfstedt:
> Caption should be "Select Academic Year" rather than "Select Educational Year".

fixed.

dartajax:
Save New Course is failing on demo3 right now as well so it may have nothing to do with this PR. If that is the case, my review can be labelled as "stale" once the save event is working again.

Issue #4591, opened 2019-06-04
Need To Restrict Adding Users to Two Different Sub Groups At Same Level


dartajax:
Currently it is possible to upload to something like the following ...

Branches >> Los Angeles >> 
Branches >> Denver >>
Branches >> Atlanta >>
Branches >> Seattle >>

All four cities are sub groups underneath the "Branches" parent group. Each city has an A Team and a B Team underneath the city (the parent group).

Using the .tsv file with the contents shown below and assuming all four users are real Ilios users with the correct CampusID's assigned, these users can be added to the A and B teams for each city. That should not be possible given the hierarchical nature of Ilios user groups.

An individual cannot be in Los Angeles and Atlanta at the same time any more than they can be in the Tuesday Small Group >> Team 1 and the Thursday Small Group >> Team 1.

The Learner Groups themselves become unmanageable in this situation as well. The user will click Manage and then Learner Group will all of a sudden appear to not have any members (or it might work). Either way, a learner cannot be in two places in the sub group structure in Ilios at the same time. 
 
![image](https://user-images.githubusercontent.com/7493331/58916410-77a6f880-86d8-11e9-9332-1f6aca203e6b.png)

stopfstedt:
fixed via #4593

PR #4593, opened 2019-06-04
reject upload if user is already present in group hierarchy.

stopfstedt:
refs #4591 

this expands the scope of the "is user already in group" validation check to entail the entire group-hierarchy (not just the current group and below), and moves the check up from the upload finalization step to the initial upload review step.

Selection_542


dartajax:
Looks good to me. 👍

PR #4594, opened 2019-06-05
Hookup Search to the API

jrjohnson:
The bare minimum to get some real search results.


jrjohnson:
Yes, and our current thinking is courses (sessions really) is all this will return for quite a while. We may add other things in the distant future.

PR #4597, opened 2019-06-05
Ember 3.10.1 upgrade

stopfstedt:


config/environment.js
```diff
@@ -59,12 +59,6 @@ module.exports = function (environment) {
       completeExistingMessages: true,
       showFileInfo: true,
     },
-    'ember-a11y-testing': {
```
jrjohnson: Oh no. If this had to get pulled as well I'm not sure I'm ok with that. The tests we have were difficult to get working and removing them is taking a big step back as some components which are currently accessible will probably get broken.

package.json
```diff
@@ -108,7 +107,7 @@
     "zxcvbn": "^4.4.2"
   },
   "engines": {
-    "node": ">= 10.*"
+    "node": "8.* || >= 10.*"
```
jrjohnson: Frontend doesn't support or test against node 8, 10 is required.

tests/integration/components/locale-chooser-test.js
```diff
@@ -3,20 +3,16 @@ import { setupRenderingTest } from 'ember-qunit';
 import { render } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
 import component from 'ilios/tests/pages/components/locale-chooser';
-import a11yAudit from 'ember-a11y-testing/test-support/audit';
 
 module('Integration | Component | locale-chooser', function(hooks) {
   setupRenderingTest(hooks);
 
-  test('it renders and is accessible', async function(assert) {
+  test('it renders', async function(assert) {
```
jrjohnson: 😢 yea. I really don't like this. I think it's worth waiting until this is fixed upstream.

.eslintrc.js
```diff
@@ -12,5 +12,46 @@ module.exports = {
...
+        "broccoli-file-creator",
+        "broccoli-merge-trees",
+        "browserslist",
+        "caniuse-db",
+        "dotenv",
+        "postcss-scss"
+      ]
+    }]
   },
+  overrides: [
```
jrjohnson: A lot of this is imported from https://github.com/ilios/eslint-config-ember/blob/master/index.js#L27. I would vote for changes that are needed here probably be needed in other ilios ember apps and they can just get added to `ilios/eslint-config-ember`

config/environment.js
stopfstedt: alright, i backed out of that.

package.json
stopfstedt: fixed.

.eslintrc.js
stopfstedt: i yanked those. turns out we don't need that new rule anyway.

PR #4598, opened 2019-06-05
Prevent subgroups with learners from being copied

### LINKED ISSUE #4592: Copy Learner Group Should Not Include Learners at Sub Group Level
Currently, a user can navigate down into the structure of a nested sub group and "copy" that Learner Group with its associated learners. The new group then is created right next to the original sub-group with the Learners. This should not be able to occur. A learner is then in multiple sub-groups under a single Parent Group - violation of Ilios hierarchical Learner Group structure.

To reproduce ....

* Go to the lowest level (where the Learners are attached) of a group that has sub-groups.
* Click the copy icon.
* Select "Copy with Learners".

"Copy With Learners" should not be an option on lower level sub-groups. It only available at the highest Learner Group level so that a new separate tree structure can be created for the newly created copy of the original learner group. This does not violate the Ilios hierarchical tree structure for Learner Groups.

So ...

Learner Groups >> Everybody (copy available with students)
Learner Groups >> Everybody >> Golden Gate Team >> Presidio (copy available - only without students)

jrjohnson:
This isn't allowed in our group model, users can only be in one branch
of a top level group, allowing subgroups to be copied with learners is
in violation of that.

Fixes #4592

### DIFF
--- a/app/components/learnergroup-list.js
+++ b/app/components/learnergroup-list.js
@@ -7,6 +7,9 @@ export default Component.extend({
   canDelete: false,
   learnerGroups: null,
   query: null,
+
+  'data-test-learnergroup-list': true,
+
   copy() {},
   remove() {},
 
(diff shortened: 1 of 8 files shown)


dartajax:
Functionally, it looks good to me.

Issue #4544, opened 2019-05-10
Visualization Call Out Information Is Not Viewable In Current State


dartajax:
Go see for yourself ... Here's a snapshot.

It looks like the callout information is somehow underneath the sidebar navigation menu - the Ilios flyout menu.

![image](https://user-images.githubusercontent.com/7493331/57560063-74b72300-7339-11e9-9d92-79eb6c2b6aec.png)

dartajax:
Steps ...

* Pull up a Course 
https://ilios3-demo.ucsf.edu/courses/1505

* Click to display the Visualizations 
![image](https://user-images.githubusercontent.com/7493331/58992413-a045f580-879f-11e9-9e74-8b51dd43bc4b.png)

* Review the Visualizations (graphs) especially as shown above "Session Types".

The same thing happens in any of them. The text values of the call out are underneath the Ilios flyout menu.

PR #4607, opened 2019-06-06
improved simple-chart tooltip display

djvoa12:
**Summary:**
Improved simple-chart tooltip display. Works for all four visualizations.

**UI:**
_Desktop_
Screen Shot 2019-06-06 at 1 13 30 PM

_Smaller Than Desktop_
Screen Shot 2019-06-06 at 1 11 40 PM

Fixes https://github.com/ilios/frontend/issues/4544

### DIFF
--- a/app/styles/components/simple-chart-tooltip.scss
+++ b/app/styles/components/simple-chart-tooltip.scss
@@ -1,0 +1,10 @@
+.simple-chart-tooltip {
+  left: 13rem !important;
+  width: 45%;
+
+  @include for-smaller-than-laptop {
+    left: 0 !important;
+    margin-left: 2.5%;
+    width: 90%;
+  }
+}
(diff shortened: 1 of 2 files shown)


app/styles/components/simple-chart-tooltip.scss
```diff
@@ -0,0 +1,10 @@
+.simple-chart-tooltip {
+  left: 13rem !important;
```
jrjohnson: Why is `!important` necessary here? It's my impression for posts like [this one](https://css-tricks.com/when-using-important-is-the-right-choice/) this this isn't a good idea.

djvoa12: I tried several things, including being more explicit, and I couldn't get it to override for now.

jrjohnson: Ok, can you create a ticket as a reminder for us to get into this at some point. It's not critical, but it's probably worth doing.

djvoa12: Done.

Issue #4611, opened 2019-06-07
global search results: only show a few sessions in a course


jrjohnson:
For each course the number of sessions should be capped at a small number (3 maybe?). The rest should be hidden, but readily accessible via a `show all sessions` link underneath each list. This should toggle open all of the sessions for a single course. There is no need to be able to toggle it closed again.

Issue #4610, opened 2019-06-06
global search: paginate search results


stopfstedt:
paginate into sets of ten courses each.

Issue #4608, opened 2019-06-06
global search: display matching attributes


stopfstedt:
each item (courses as well sessions) in the global search payload contains a `matchedIn` property which is a list of attributes-names that matched the given search term.

display this information as part of the search results list interface.

djvoa12:
@stopfstedt there's an inconsistency for the `matchIn` values.

It comes in as an array of strings.
```
"matchedIn": ["description", "title", "learningmaterials"]
```
It also comes in as an array of objects as well.
```
				"matchedIn": {
					"0": "description",
					"2": "objectives",
					"3": "title",
					"4": "learningmaterials"
				}
```
Should we pick one?

jrjohnson:
Whoops. Can you file this as an issue in ilios/ilios and assign it to me?  Should be strings.

djvoa12:
https://github.com/ilios/ilios/issues/2529

PR #4625, opened 2019-06-13
search UI additions

djvoa12:
**Summary:**
- Shows course and session tags
- Initially show up to three session (show more button to show the rest)
- Added frontend pagination

**UI:**
search

Fixes https://github.com/ilios/frontend/issues/4608
Fixes https://github.com/ilios/frontend/issues/4610
Fixes https://github.com/ilios/frontend/issues/4611

### DIFF
--- a/app/components/course-search-result.js
+++ b/app/components/course-search-result.js
@@ -10,9 +10,9 @@ export default Component.extend({
   showMore: false,
 
   courseTags: reads('course.matchedIn'),
-  sessions: reads('course.sessions'),
 
-  computedSessions: computed('showMore', function() {
-    return this.showMore ? this.sessions : this.sessions.slice(0, 3);
+  sessions: computed('showMore', function() {
+    const sessions = this.course.sessions;
+    return this.showMore ? sessions : sessions.slice(0, 3);
   })
 });
--- a/app/components/course-search-result.js
+++ b/app/components/course-search-result.js
@@ -1,0 +1,18 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { reads } from '@ember/object/computed';
+
+export default Component.extend({
+  classNames: ['course-search-result', 'results-row'],
+  tagName: 'li',
+
+  course: null,
+  showMore: false,
+
+  courseTags: reads('course.matchedIn'),
+  sessions: reads('course.sessions'),
+
+  computedSessions: computed('showMore', function() {
+    return this.showMore ? this.sessions : this.sessions.slice(0, 3);
+  })
+});
--- a/app/components/course-search-result.js
+++ b/app/components/course-search-result.js
@@ -3,7 +3,7 @@ import { computed } from '@ember/object';
 import { reads } from '@ember/object/computed';
 
 export default Component.extend({
-  classNames: ['course-search-result', 'results-row'],
+  classNames: ['course-search-result'],
   tagName: 'li',
 
   course: null,
--- a/app/components/pagination-links.js
+++ b/app/components/pagination-links.js
@@ -22,31 +22,36 @@ export default Component.extend({
 
   pages: computed('page', 'totalPages', function() {
     const { lastPage, page } = this.getProperties('lastPage', 'page');
-
-    if (lastPage <= 7) {
-      const pageNumbers = [];
-
-      for (let i = 1; i <= lastPage; i++) {
-        pageNumbers.push(i);
-      }
-
-      return pageNumbers;
-    } else {
-      if (page <= 4) {
-        return [1, 2, 3, 4, 5, '...', lastPage];
-      }
-
-      if (page >= (lastPage - 3)) {
-        return [1, '...', lastPage - 4, lastPage - 3, lastPage - 2, lastPage - 1, lastPage];
-      }
-
-      return [1, '...', page - 1, page, page + 1, page + 2, '...', lastPage];
-    }
+    return lastPage <= 7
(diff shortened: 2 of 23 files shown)


app/components/course-search-result.js
```diff
@@ -0,0 +1,18 @@
...
+
+export default Component.extend({
+  classNames: ['course-search-result', 'results-row'],
+  tagName: 'li',
+
+  course: null,
+  showMore: false,
+
+  courseTags: reads('course.matchedIn'),
+  sessions: reads('course.sessions'),
```
jrjohnson: Remove this property and just call the `computedSessions` property `sessions`.

app/components/pagination-links.js
```diff
@@ -0,0 +1,48 @@
...
+
+  disableNext: computed('lastPage', 'page', function() {
+    return this.lastPage === this.page;
+  }),
+
+  lastPage: computed('results', 'size', function() {
+    return Math.ceil(this.results.length / this.size);
+  }),
+
+  computedPages: computed('page', 'totalPages', function() {
```
jrjohnson: Just `pages` please. calling it `computedPages` is redundant.

app/controllers/search.js
```diff
@@ -1,9 +1,18 @@
 import Controller from '@ember/controller';
 
 export default Controller.extend({
-  queryParams: {
-    query: 'q'
-  },
+  queryParams: ['page', { query: 'q' }],
```
jrjohnson: This is much more confusing than
```
queryParams: {
    query: 'q',
    page: 'page'
  },
```

app/controllers/search.js
```diff
@@ -1,9 +1,18 @@
...
 
-  query: ''
+  page: 1,
+  query: '',
+
+  actions: {
+    prevPage() {
+      this.set('page', this.page - 1);
```
jrjohnson: Can you click on a page number directly? In that case it seems like both of these actions should just be replaced with `setPage` and the computation can be done in the component itself.

app/styles/components/course-search-result.scss
```diff
@@ -0,0 +1,31 @@
+.course-search-result {
+  margin-bottom: 1rem;
+  max-width: 40rem;
+
+  .matched-in {
```
jrjohnson: is `matched-in` different from `global-search-tags`?

app/styles/components/course-search-result.scss
```diff
@@ -0,0 +1,31 @@
...
+
+  .global-search-tags {
+    margin: .125rem 0;
+    overflow: hidden;
+
+    .global-search-tag {
+      background: lighten($grey, 5%);
+      border: 1px solid lighten($dark-grey, 10%);
+      border-radius: .125rem;
+      color: $darker-grey;
```
jrjohnson: There isn't enough contrast here to be legible. The background needs to be lighter or else the color should be darker.

app/styles/components/pagination-links.scss
```diff
@@ -0,0 +1,79 @@
...
+    outline: none;
+    padding: .125rem 0;
+
+    &:hover {
+      background: $ilios-orange;
+      border: 1px solid $ilios-orange;
+      color: $white;
+    }
+
+    @media only screen and (max-width: 550px) {
```
jrjohnson: This looks amazing! Love how it collapses down. Really nice.

app/templates/components/global-search-tags.hbs
```diff
@@ -0,0 +1,17 @@
+{{#if @tags}}
+  <div class="global-search-tags">
+    {{#each @tags as |tag|}}
+      <span class="global-search-tag">
+        {{#if (eq tag "id")}}
```
jrjohnson: isn't this the same as the last `concat`ed option?

app/components/course-search-result.js
djvoa12: 👍

app/components/pagination-links.js
djvoa12: 👍

app/controllers/search.js
djvoa12: 👍

app/controllers/search.js
djvoa12: 👍

app/styles/components/course-search-result.scss
djvoa12: This needs to be removed.

app/templates/components/global-search-tags.hbs
djvoa12: Yes.

app/styles/components/course-search-result.scss
djvoa12: I will lighten the background a little more.

app/components/course-search-result.js
```diff
@@ -0,0 +1,18 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { reads } from '@ember/object/computed';
+
+export default Component.extend({
+  classNames: ['course-search-result', 'results-row'],
```
jrjohnson: Is results-row still used? Looks like it's gone form the css.

app/styles/components/course-search-result.scss
jrjohnson: Still not great, but I'm ok with it for now.

app/components/course-search-result.js
djvoa12: Will remove.

PR #4663, opened 2019-06-27
new Promise refactor to async/await

djvoa12:
**Summary:**
`new Promise` refactor to `async/await`.

### DIFF
--- a/app/components/competency-title-editor.js
+++ b/app/components/competency-title-editor.js
@@ -1,5 +1,7 @@
 import Component from '@ember/component';
 import { isPresent } from '@ember/utils';
+import { reject } from 'rsvp';
+import { task } from 'ember-concurrency';
 import { validator, buildValidations } from 'ember-cp-validations';
 import ValidationErrorDisplay from 'ilios-common/mixins/validation-error-display';
 
@@ -35,20 +37,23 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
       if (isPresent(competency)) {
         this.set('title', competency.get('title'));
       }
-    },
-
-    save() {
-      this.send('addErrorDisplayFor', 'title');
-
-      if (this.validations.isValid) {
-        const { competency, title } = this.getProperties('competency', 'title');
-
-        if (isPresent(competency)) {
-          competency.set('title', title);
-        }
-
-        this.send('clearErrorDisplay');
-      }
     }
-  }
+  },
+
+  save: task(function* () {
+    this.send('addErrorDisplayFor', 'title');
+    const { validations } = yield this.validate();
+
+    if (validations.isValid) {
+      const { competency, title } = this.getProperties('competency', 'title');
+
+      if (isPresent(competency)) {
+        competency.set('title', title);
+      }
+
+      this.send('clearErrorDisplay');
+    } else {
+      yield reject();
+    }
+  })
 });
--- a/app/components/competency-title-editor.js
+++ b/app/components/competency-title-editor.js
@@ -53,7 +53,7 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
 
       this.send('clearErrorDisplay');
     } else {
-      yield reject();
+      return reject();
     }
   })
 });
--- a/app/components/competency-title-editor.js
+++ b/app/components/competency-title-editor.js
@@ -1,9 +1,7 @@
 import Component from '@ember/component';
 import { isPresent } from '@ember/utils';
-import { Promise } from 'rsvp';
 import { validator, buildValidations } from 'ember-cp-validations';
 import ValidationErrorDisplay from 'ilios-common/mixins/validation-error-display';
-import { task } from 'ember-concurrency';
 
 const Validations = buildValidations({
   title: [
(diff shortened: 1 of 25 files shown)


app/components/competency-title-editor.js
```diff
@@ -37,24 +35,20 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
...
         this.set('title', competency.get('title'));
       }
-    }
-  },
+    },
+
+    save() {
+      this.send('addErrorDisplayFor', 'title');
+
+      if (this.validations.isValid) {
```
jrjohnson: I think you need to call `await this.validate();` before checking `this.validations.isValid`.

djvoa12: Isn't that the old way? It should be a synchronous check now, no?

http://offirgolan.github.io/ember-cp-validations/docs/modules/Accessing%20Validations.html

jrjohnson: While isValid is not async the validations themselves must be run. If a field hasn't been touched they may not have been. I think that's where `this.validate()` comes into play.

app/components/competency-title-editor.js
```diff
@@ -42,19 +42,18 @@ export default Component.extend(Validations, ValidationErrorDisplay, {
...
+      const { competency, title } = this.getProperties('competency', 'title');
+
+      if (isPresent(competency)) {
+        competency.set('title', title);
       }
-    });
+
+      this.send('clearErrorDisplay');
+    } else {
+      yield reject();
```
jrjohnson: I feel like this should be `return reject()` and not `yield`.

app/components/dashboard-myreports.js
```diff
@@ -35,12 +34,9 @@ export default Component.extend(DomMixin, ReportTitleMixin, {
...
    * @public
    */
-  sortedReports: computed('user.reports.[]', function() {
-    return new Promise(resolve => {
-      this.user.get('reports').then(reports => {
-        resolve(reports.sortBy('title'));
-      });
-    });
+  sortedReports: computed('user.reports.[]', async function() {
+    const reports = await this.user.reports;
```
jrjohnson: I don't think you can chain getters in ember-data yet (according to something I saw somewhere i don't remember) so this should be `await this.user.get('reports');

app/components/my-profile.js
```diff
@@ -25,15 +24,6 @@ export default Component.extend({
   host: reads('iliosConfig.apiHost'),
   namespace: reads('iliosConfig.apiNameSpace'),
 
-  roles: computed('user.roles.[]', function() {
```
jrjohnson: ?? removed?

app/components/new-myreport.js
```diff
@@ -23,26 +23,18 @@ const PrepositionObject = EmberObject.extend({
...
-        resolve(model.get('title'));
-      }
-    });
+  label: computed('model', 'type', async function() {
+    const { model, type } = this.getProperties('model', 'type');
+
+    if (type === 'mesh term') {
+      return model.name;
+    } else if (type === 'term') {
+      const vocabulary = await model.vocabulary;
```
jrjohnson: Think this is ED as well so should be `model.get('vocabulary')`

app/components/my-profile.js
djvoa12: It wasn't being used. I could put it back.

app/components/dashboard-myreports.js
djvoa12: I think I checked all updates and they were coming back the same.

```
// From
something.get('something');

// To
something.something
```

app/components/new-myreport.js
djvoa12: I think I checked all updates and they were coming back the same.

```
// From
something.get('something');

// To
something.something
```

app/components/competency-title-editor.js
djvoa12: I think I tried to return it but the promise just hung and the spinner wouldn't resolve.

app/components/my-profile.js
jrjohnson: Ok, cool. No leave it out.

app/components/new-myreport.js
jrjohnson: I'm worried by https://blog.emberjs.com/2018/02/16/ember-3-1-beta-released.html which says 

> Ember Data returns promise proxy objects when you read an async relationship and from other API. Ember proxy objects, including promise proxies, still require that you call get to read values.

And this is one of those. It may only be working because network is fast or the data is preloaded so it's async but not really.

app/components/competency-title-editor.js
jrjohnson: Ok, go ahead and revert this one then back to the old format. There is probably something else wrong with it and I'll take a look at some future point.

app/components/new-myreport.js
djvoa12: Okay. We could stay on the safe side then. 👍

app/components/competency-title-editor.js
djvoa12: Checked again. It must have been with `await` instead. `return reject()` works under `task`.

app/components/dashboard-myreports.js
djvoa12: Reverted and addressed.

app/components/new-myreport.js
djvoa12: Reverted and addressed.

stopfstedt: @thomas, still seeing `model.vocabulary` here. did you forget to push after reverting?

djvoa12: I guess I did forget to push.

PR #4689, opened 2019-07-07
Different API Version Notice

jrjohnson:
Common now provides an improved component for this purpose, translations and some styles have moved there.

Requires: ilios/common#758


stopfstedt:
hey, just thought about this - wouldn't we want this in the LTIs as well? if so - let's put this into common.

PR #4739, opened 2019-08-05
prevent rollover of report into the same year.

### LINKED ISSUE #4049: Current Year Should Not Be Available in CI Report Rollover
It is possible as shown in the screen shot below to select the current year as the year to create the new CI Report. Only one CI Report is allowed per year so this should not be possible.

image

stopfstedt:
fixes #4049


dartajax:
@stopfstedt - I reviewed just now with @saschaben - I noticed that you can roll over a CI report into another year that already contains a ***finalized*** CI report. This should not be the case. If a year already has a ***finalized*** CI report, that year should not be available in the rollover drop-down.

stopfstedt:
ok. please file this as a follow-up ticket.

PR #4741, opened 2019-08-05
adds input options for start/end-date to ci rollover form.

### LINKED ISSUE #4042: CI Report Enhancement Request - Be Able to Specify Start / End Dates for New Report
Below is the interface used in rolling over CI Reports. The start and end dates should be included here with the default being one year later (or a different amount of years if a different target year was chosen) for each of the values. They can then be changed to reflect accuracy. This way when the rolled over CI Report is created, it will not have the previous report's start and end dates.

image

stopfstedt:
fixes #4042 

this tightens up the rollover form while at it.

Selection_670

### DIFF
--- a/app/components/curriculum-inventory-report-rollover.js
+++ b/app/components/curriculum-inventory-report-rollover.js
@@ -14,6 +14,18 @@ const Validations = buildValidations({
       min: 3,
       max: 200
     })
+  ],
+  startDate: [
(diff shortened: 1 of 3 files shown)


stopfstedt:
currently blocked by https://github.com/ilios/ilios/issues/2583

stopfstedt:
not needed anymore.

PR #4774, opened 2019-08-15
Enforce minimum search length

### LINKED ISSUE #4769: Search: Need To Provide A Message if Search Attempted with Fewer Than 3 Characters
In this case, I had just run a successful search on the term `brain`; but I decided to run a really wide search on only `br`. It states "No Results" but in reality, not enough characters had been entered to run the search. We should provide that information like we do in other places in Ilios when a minimum field length value has not been entered. We don't advise there is a minimum anywhere and the message is incorrect since clearly many results would have been returned by searching on `br` - too many actually.

image

jrjohnson:
Added a keep typing prompt when more input is needed, don't do anything
if the search is too short.

Fixes #4769

### DIFF
--- a/app/components/course-search-result.js
+++ b/app/components/course-search-result.js
@@ -5,6 +5,7 @@ import { reads } from '@ember/object/computed';
 export default Component.extend({
   classNames: ['course-search-result'],
   tagName: 'li',
+  'data-test-course-search-result': true,
 
   course: null,
   showMore: false,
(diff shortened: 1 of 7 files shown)


jrjohnson:
Fixed. Thanks @dartajax

dartajax:
LGTM 💯 merge it later when you can or we can do it in the morning @jrjohnson

PR #4799, opened 2019-08-29
Remove redundant package

jrjohnson:
This is provided by common.


jrjohnson:
This breaks in a very weird way. More investigation required.

stopfstedt:
it is known. suggesting leaving as-is.

PR #4855, opened 2019-09-27
updates ember-cli-inject-live-reload dependency.

stopfstedt:
dependabot didn't pick up that major version change, so rolling this by hand.

https://github.com/ember-cli/ember-cli-inject-live-reload/compare/v1.10.2...v2.0.1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11002,9 +11002,9 @@
       "dev": true
     },
     "ember-cli-inject-live-reload": {
-      "version": "1.10.2",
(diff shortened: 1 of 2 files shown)


jrjohnson:
We ignore this in https://github.com/ilios/frontend/blob/master/.dependabot/config.yaml#L31 it is supposed to get updated by `ember-cli-update`, but it looks like it's updated in the app blueprints at https://github.com/ember-cli/ember-cli/blob/v3.13.1/blueprints/app/files/package.json#L31 already so.... YEAH!

PR #4891, opened 2019-10-24
Add school name to search results

### LINKED ISSUE #4847: Add School Identifier to Course-level search results
When a course title is returned in the search results, it should be identified by owning school and year. This will provide additional user context and support the solution included in #4754 .
School label should use School short name.

Ilios Search 2019-09-24 11-22-33

jrjohnson:
The school a course belongs to appears as a tag next to each course.

Fixes #4847

### DIFF
--- a/app/styles/components/course-search-result.scss
+++ b/app/styles/components/course-search-result.scss
@@ -13,6 +13,17 @@
     position: relative;
   }
 
+  .school-flag {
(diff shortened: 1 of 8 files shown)


jrjohnson:
@saschaben can you check on this color and see if it works or if we need to add something new?
Screen Shot 2019-10-24 at 11 10 31 AM

PR #4937, opened 2019-11-17
Update to Ember 3.14

jrjohnson:
Also updated blueprints which includes a move to some new ES Class
syntax... this Octane thing is getting very serious now!


dartajax:
Click Test approved - gonna merge.

PR #5044, opened 2020-01-11
Improve Locale Chooser

jrjohnson:
Converted to glimmer component and octane patterns
Fixed a11y where the menubar and menuitem were incorrectly applied to the toggle.

### DIFF
--- a/app/components/locale-chooser.js
+++ b/app/components/locale-chooser.js
@@ -1,81 +1,86 @@
-import Component from '@ember/component';
-import { schedule } from '@ember/runloop';
-import { computed } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 
-export default Component.extend({
-  intl: service(),
-  moment: service(),
-  classNameBindings: [':locale-chooser', 'isOpen'],
-  isOpen: false,
-  ariaRole: 'menubar',
-  'data-test-locale-chooser': true,
+export default class LocaleChooserComponent extends Component {
+  @service intl;
+  @service moment;
+  @tracked isOpen = false;
 
-  locale: computed('locales.[]', 'intl.locale', function () {
+  get locale() {
     const locale = this.intl.get('locale');
     return this.locales.findBy('id', locale[0]);
-  }),
+  }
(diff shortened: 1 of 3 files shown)


app/components/locale-chooser.js
```diff
@@ -1,81 +1,86 @@
...
       break;
     case 'ArrowUp':
-      if (originalEvent.target.previousElementSibling) {
-        originalEvent.target.previousElementSibling.focus();
+      event.preventDefault();
+      if (target.previousElementSibling) {
+        target.previousElementSibling.focus();
       } else {
-        this.element.querySelector('.menu button:last-of-type').focus();
+        this.menuElement.querySelector('button:last-of-type').focus();
```
stopfstedt: should be `this.element...`

app/components/locale-chooser.js
```diff
@@ -1,81 +1,86 @@
...
+        target.nextElementSibling.focus();
       } else {
-        if (originalEvent.target.nextElementSibling) {
-          originalEvent.target.nextElementSibling.focus();
-        } else {
-          schedule('afterRender', () => {
-            this.element.querySelector('.menu button:nth-of-type(1)').focus();
-          });
-        }
+        this.menuElement.querySelector('button:nth-of-type(1)').focus();
```
stopfstedt: should be `this.element...`

app/components/locale-chooser.js
```diff
@@ -1,81 +1,86 @@
...
     case 'ArrowRight':
     case 'ArrowLeft':
-      this.set('isOpen', false);
+      this.isOpen = false;
+      break;
+    }
+  }
+  @action
+  clearFocus() {
+    const buttons = this.menuElement.querySelectorAll('button');
```
stopfstedt: should be `this.element...`

app/components/locale-chooser.js
jrjohnson: `this.element` doesn't exist for glimmer components, instead this uses the `{{ref}} modifier in the template to assign an element to `menuElement` https://guides.emberjs.com/release/components/template-lifecycle-dom-and-modifiers/#toc_communicating-between-elements-in-a-component

stopfstedt: i stand corrected.

PR #5063, opened 2020-01-18
Move noscript tag into the footer of the body tag

jrjohnson:
Putting this anywhere in the header messes with a11y automated scanning
when we replace the title tag.


jrjohnson:
Merging this myself to be in before our next siteimprove scan.

PR #5082, opened 2020-01-24
Test and Deploy with Github Actions

jrjohnson:


.github/workflows/deploy-production.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js 12
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12
+    - name: install dependencies
+      run: npm ci
+    - name: Ember CLI Deploy
+      run: npm run deploy:staging
```
stopfstedt: shouldn't this be `deploy:production` then?

PR #5199, opened 2020-03-19
Created custom publications status component for CI

jrjohnson:
This was overriding language options that aren't in the new version,
just created a new one without as many options and use it only here.


stopfstedt:
> LGTM

LIES! re-reviewing.

app/components/curriculum-inventory/publication-status.js
```diff
@@ -0,0 +1,24 @@
+import Component from '@glimmer/component';
+
+export default class CuriculumInventoryPublicationStatusComponent extends Component {
+  get textKey(){
+    if (this.args.item.isScheduled) {
```
stopfstedt: inventory reports really only have one status, that's `Finalized`.

should therefore be something like: 

```javascript
if (this.args.item.isScheduled) {
  return 'general.finalized';
}  else {
  return 'general.draft';
}
```

PR #5201, opened 2020-03-19
Use explicit this in route templates

jrjohnson:
The only thing that can really get passed into these is the `@model` so
everything else get's a this.

### DIFF
--- a/app/templates/head.hbs
+++ b/app/templates/head.hbs
@@ -5,5 +5,5 @@ setting values on the 'head-data' service.
 }}
 <title>
   {{t "general.ilios"}}
-  {{model.title}}
+  {{@model.title}}
 </title>
--- a/app/templates/learner-group.hbs
+++ b/app/templates/learner-group.hbs
@@ -1,13 +1,13 @@
 {{title (t "general.learnerGroups")}}
 <LearnergroupSummary
-  @setIsEditing={{action (mut isEditing)}}
-  @canCreate={{canCreate}}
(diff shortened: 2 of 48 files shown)


app/templates/head.hbs
```diff
@@ -5,5 +5,5 @@ setting values on the 'head-data' service.
 }}
 <title>
   {{t "general.ilios"}}
-  {{model.title}}
+  {{@model.title}}
```
stopfstedt: i don't think this does anything, but that's for another time.

app/templates/learner-group.hbs
```diff
@@ -1,13 +1,13 @@
...
-  @sortUsersBy={{sortUsersBy}}
-  @setSortUsersBy={{action (mut sortUsersBy)}}
+  @setIsEditing={{action (mut this.isEditing)}}
+  @canCreate={{this.canCreate}}
+  @canUpdate={{this.canUpdate}}
+  @canDelete={{this.canDelete}}
+  @setIsBulkAssigning={{action (mut this.isBulkAssigning)}}
+  @sortUsersBy={{this.sortUsersBy}}
+  @setSortUsersBy={{action (mut this.sortUsersBy)}}
   @learnerGroup={{model}}
```
stopfstedt: should be `@model`.

stopfstedt: on second thought - this is out of scope. will issue a PR  that addresses this later today.

PR #5221, opened 2020-03-25
Update to Common v38

jrjohnson:
Breaking change in common removed controller mixins which is handled here.


dartajax:
After selecting a Vocabulary >> Topic, the LCME callouts don't go away.

image

dartajax:
another fun screen shot ...

image

dartajax:
image

dartajax:
A hard refresh is required to get it to go away.

jrjohnson:
I had trouble getting the bottom of this then went a little too far and stripped out these tooltips in favor of building our own in https://github.com/ilios/common/pull/1271 

I think this update PR can be merged as is with this problem (which I don't think is caused by this change, probably just revealed by it) and then the next common release will have the new tooltips and fix it.

PR #5283, opened 2020-04-20
rm redundant dependency.

stopfstedt:
we're getting `ember-composable-helpers` from `common` already, let's remove the explicit dependency here since it's redundant.


stopfstedt:
refs https://github.com/ilios/frontend/pull/5280

PR #5333, opened 2020-05-13
Stop compressing font files

jrjohnson:
Woff and Woff2 are specialized formats that are highly
compressed and tend to get bigger when gzipped. They shouldn't be
included in the defaults.

### DIFF
--- a/config/deploy.js
+++ b/config/deploy.js
@@ -59,12 +59,12 @@ module.exports = function (deployTarget) {
       }
     },
     gzip: {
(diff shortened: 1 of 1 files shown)


dartajax:
I put DO NOT MERGE on there for now. The code looks good and logical but I received an API mismatch trying to test it so if that is expected or can be resolved, I will merge it in the morning.

jrjohnson:
Good catch, but mismatch is expected right now and not a result of this change.

PR #5337, opened 2020-05-15
octanifies ci-sequence-block-header component.

stopfstedt:


app/components/curriculum-inventory-sequence-block-header.js
```diff
@@ -1,56 +1,36 @@
...
-        this.set('blockTitle', newBlock.get('title'));
-        this.set('sequenceBlock', newBlock);
-      } else {
-        await reject();
-      }
-    },
+  @action
+  revertTitleChanges() {
+    const block = this.args.sequenceBlock;
+    this.title = block.title;
```
jrjohnson, line 21: `block` seems like an unnecessary local variable now that a `.get` is no longer needed.

tests/integration/components/curriculum-inventory-sequence-block-header-test.js
```diff
@@ -1,140 +1,97 @@
...
   });
 
   test('read-only mode for block in when it can not be updated', async function(assert) {
-    const block = EmberObject.create({
-      title: 'Block title',
-      report: {
-      }
-    });
+    const block = this.server.create('curriculum-inventory-sequence-block', { title: 'Block title' });
     this.set('sequenceBlock', block);
```
jrjohnson, line 25: This works for now, but `block` here isn't actually an EmberData model, it's a mirage value object. Should be `const blockModel = await this.owner.lookup('service:store').find('curriculum-inventory-sequence-block', block.id);` to be test everything as it is for a user.

app/components/curriculum-inventory-sequence-block-header.js
stopfstedt: good catch. handled.

tests/integration/components/curriculum-inventory-sequence-block-header-test.js
stopfstedt: handled. :+1:

PR #5426, opened 2020-06-30
Load less data on learner group list

### LINKED ISSUE #5394: Change Learner Group Detail - Badges for Associated Courses
Here is how it currently appears after clicking on a Learner Group title to go to its detail screen ...

image

Although this is an enhancement and not a requirement, it would be nice if we could convert the text strings of the courses associated with a given learner group into badges and provide a counter in this context of how many courses have used this learner group in their learning events. Consider moving this higher up on the screen as well.

Something like this maybe ...

image

jrjohnson:
I've removed the course count and moved the check for linked courses
into the delete dialog. This way we don't need to load the entire
curriculum when loading the learner groups list.

To make the course information easy to get to it has been highlighted
and re-styled in the single group view.

Fixes #5394

Todo:
- [x] Feedback on removal warning when courses are attached
- [x] Translations
- [x] Feedback on new single group summary

### DIFF
--- a/app/components/learnergroup-list.js
+++ b/app/components/learnergroup-list.js
@@ -25,12 +25,10 @@ export default class LearnerGroupListComponent extends Component {
 
   @task
   *confirmRemove(learnerGroup) {
-    if (this.args.canDelete) {
-      this.preparingToRemove = [...this.preparingToRemove, learnerGroup];
-      const deletableGroup = yield this.createDeletableGroup(learnerGroup);
-      this.toRemove = [...this.toRemove, deletableGroup];
-      this.preparingToRemove = this.preparingToRemove.filter(lg => lg !== learnerGroup);
-    }
+    this.preparingToRemove = [...this.preparingToRemove, learnerGroup];
+    const deletableGroup = yield this.createDeletableGroup(learnerGroup);
+    this.toRemove = [...this.toRemove, deletableGroup];
+    this.preparingToRemove = this.preparingToRemove.filter(lg => lg !== learnerGroup);
   }
 
   @action
--- a/app/components/learnergroup-list.js
+++ b/app/components/learnergroup-list.js
@@ -2,10 +2,12 @@ import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { task } from 'ember-concurrency-decorators';
+import { map } from 'rsvp';
 
 export default class LearnerGroupListComponent extends Component {
-  @tracked learnerGroupsForCopy = [];
-  @tracked learnerGroupsForRemovalConfirmation = [];
+  @tracked toCopy = [];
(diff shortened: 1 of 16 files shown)


dartajax:
I kind of miss not seeing the count of courses but I did ask about it and don't think people rely on it that much. I would suggest the button to not say `Cancel` but `OK` so the user just has to acknowledge the fact that the learner group cannot be deleted. One question, when you are logged in as Nicholas, there is one LG in Class of 2024 that cannot be deleted. Is that because Nicholas created that group? This one ... https://deploy-preview-5426--ilios-frontend.netlify.app/learnergroups/17257

jrjohnson:
I replaced with confirmation with `OK`.

The group that can't be deleted has learner in it (or in one of the subgroups). That was existing behavior and I didn't change it here.

dartajax:
image

Showing associated courses (0) ...
image

Now try and delete this Learner Group ...
image

dartajax:
Further testing ... try to delete one Learner Group, endless spinner - no immediate window - sorry

image

I clicked on the offending learner group and got a Zoinks message.

Never mind this is an API mis-match issue.

jrjohnson:
@dartajax latest version now displays all courses attached to the group and its sug-groups in the `Associated Courses` section. This is the same list that we're using on the delete page. I wasn't able to come up with a good way to represent what groups a course was attached to so I just added a counter for now, maybe we can refine later? Or you have an idea?

jrjohnson:
Good catch @dartajax, that should be fixed now and correctly indicate the number of subgroups no matter what level the course is attached at.

dartajax:
Thanks for the fix @jrjohnson - it looks much better to me.

app/components/learnergroup-list.js
```diff
@@ -1,84 +1,82 @@
...
+    const offerings = (await learnerGroup.offerings).toArray();
+    const ilms = (await learnerGroup.ilmSessions).toArray();
+    const arr = [].concat(offerings, ilms);
 
-    async copy(withLearners, learnerGroup) {
-      const copy = this.copy;
-      await copy(withLearners, learnerGroup);
-      this.learnerGroupsForCopy.removeObject(learnerGroup);
-    },
+    const sessions = await Promise.all(arr.mapBy('session'));
```
stopfstedt, line 66: ok, so native `Promise.all()` over `RSVP.all()`. noted.

app/components/learnergroup-list.js
```diff
@@ -1,84 +1,82 @@
...
+    const filteredSessions = sessions.filter(Boolean).uniq();
```
stopfstedt, line 67: what is this ork mischief?!

jrjohnson: Not my fault, I learned it from @Trott 😆. Allegedly `Boolean` is a function that returns a bool. Fancy right?

PR #5451, opened 2020-07-13
Remove course-loading route

jrjohnson:
This is now provided by common as of https://github.com/ilios/common/pull/1481


dartajax:
Sometimes when scrolling through months, Sunday gets highlighted as shown below. I also believe the year is fixed in position and the fonts are not the same between the month and year. 

image

PR #5455, opened 2020-07-14
Octanefiy DashboardMyreports component

### LINKED ISSUE #2977: Occasionally The First Report Entered Does Not Show Up / Requires Refresh
image

image

Still showing None 

I am logged in as Nicholas Gutierrez right now.

After refreshing the screen `F5`, the report appears. See below.

image

jrjohnson:
Removed the report title component because it wasn't necessary anymore,
we have to generate all of the titles during the initial display in
order to sort correctly. The original sorting wasn't actually working as
reports with a default empty title would be pushed to the top, it just
looked ok because most of them start with `All...`.

I think this fixes #2977


dartajax:
it stalled out on me ...

image

dartajax:
image

dartajax:
My "stall out" screen shots are indicative of a different problem that also exists in production and will be ticketed shortly. I'm gonna merge this one since that is the case.

PR #5501, opened 2020-08-05
Learner Group Management: display learner full-name instead of first/last name.

stopfstedt:
Replace the first and last-name columns with one "name" column that considers user preferred names, if given.

Full name will use display-name ("preferred name") if given, otherwise it will fall back to firstname + middlename + lastname.

### DIFF
--- a/app/components/learnergroup-cohort-user-manager.js
+++ b/app/components/learnergroup-cohort-user-manager.js
@@ -24,8 +24,7 @@ export default class LearnergroupCohortUserManagerComponent extends Component {
     }
 
     return this.args.users.filter((user) => {
-      return user.firstName.toLowerCase().includes(filter) ||
-        user.lastName.toLowerCase().includes(filter) ||
+      return user.fullName.toLowerCase().includes(filter) ||
         user.email.toLowerCase().includes(filter);
     });
   }
--- a/app/components/learnergroup-cohort-user-manager.js
+++ b/app/components/learnergroup-cohort-user-manager.js
@@ -24,7 +24,9 @@ export default class LearnergroupCohortUserManagerComponent extends Component {
     }
 
     return this.args.users.filter((user) => {
-      return user.fullName.toLowerCase().includes(filter) ||
+      return user.firstName.toLowerCase().includes(filter) ||
+        user.lastName.toLowerCase().includes(filter) ||
+        user.fullName.toLowerCase().includes(filter) ||
         user.email.toLowerCase().includes(filter);
     });
   }
--- a/app/components/learnergroup-user-manager.hbs
+++ b/app/components/learnergroup-user-manager.hbs
@@ -31,23 +31,15 @@
                 {{/if}}
               </th>
               <SortableTh
-                @colspan={{2}}
-                @click={{fn this.setSortBy "firstName"}}
+                @colspan={{4}}
+                @click={{fn this.setSortBy "fullName"}}
                 @sortedBy={{or
-                  (eq sortBy "firstName")
-                  (eq sortBy "firstName:desc")
+                  (eq sortBy "fullName")
+                  (eq sortBy "fullName:desc")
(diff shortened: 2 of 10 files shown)


app/components/learnergroup-cohort-user-manager.js
```diff
@@ -24,8 +24,7 @@ export default class LearnergroupCohortUserManagerComponent extends Component {
     }
 
     return this.args.users.filter((user) => {
-      return user.firstName.toLowerCase().includes(filter) ||
-        user.lastName.toLowerCase().includes(filter) ||
+      return user.fullName.toLowerCase().includes(filter) ||
```
jrjohnson, line 27: I know it might be confusing to see results on the screen that don't match the filter you entered, but I think we should continue to filter on first / last in addition to full name. This ensures that a match will show up if someone has changed their name and the admin doesn't know or has forgotten.

tests/integration/components/learnergroup-cohort-user-manager-test.js
```diff
@@ -1,132 +1,123 @@
...
-    const user2FirstName = 'tbody tr:nth-of-type(2) td:nth-of-type(2)';
-    const user2LastName = 'tbody tr:nth-of-type(2) td:nth-of-type(3)';
-    const user2CampusId = 'tbody tr:nth-of-type(2) td:nth-of-type(4)';
-    const user2Email = 'tbody tr:nth-of-type(2) td:nth-of-type(5)';
+    const user2FullName = 'tbody tr:nth-of-type(2) td:nth-of-type(2)';
+    const user2CampusId = 'tbody tr:nth-of-type(2) td:nth-of-type(3)';
+    const user2Email = 'tbody tr:nth-of-type(2) td:nth-of-type(4)';
 
-    const user1 = EmberObject.create({
+    const user1 = this.server.create('user', {
```
jrjohnson, line 23: Every time we update one of these it makes me 😄

tests/integration/components/learnergroup-user-manager-test.js
```diff
@@ -1,220 +1,255 @@
...
-    const user2 = EmberObject.create({
+    const user2 = this.server.create('user', {
       firstName: 'Jackson',
-      lowestGroupInTree: EmberObject.create({
-        id: 1
-      }),
+      learnerGroups: [ learnerGroup ],
+    });
+    const user3 = this.server.create('user', {
+      firstName: 'Jayden',
```
jrjohnson, line 173: 🤩

app/components/learnergroup-user-manager.hbs
```diff
@@ -31,23 +31,15 @@
                 {{/if}}
               </th>
               <SortableTh
-                @colspan={{2}}
-                @click={{fn this.setSortBy "firstName"}}
+                @colspan={{4}}
```
jrjohnson, line 34: First click here doesn't do anything other than update the URL. It should reverse the sort.  Also the icon here doesn't match the one on users in cohort not in this group. Should be AZ in both places.

stopfstedt: > First click here doesn't do anything other than update the URL. It should reverse the sort. Also the icon here doesn't match the one on users in cohort not in this group. Should be AZ in both places.

i fixed the sort order bug by updating the default sort order in the controller.

app/components/learnergroup-cohort-user-manager.js
stopfstedt: updated to consider first/last names in filter again.

app/components/learnergroup-user-manager.hbs
stopfstedt: fixed, the correct icon now gets applied. there is a bug in the `<SortableTh>` component where the sort order icon does not get updated if sort order is reversed, but this will need to be corrected in common.

PR #5507, opened 2020-08-10
Octanify CI components

stopfstedt:


app/components/curriculum-inventory-sequence-block-overview.js
```diff
@@ -121,255 +127,206 @@ export default Component.extend({
...
+  }
+
+  @action
+  revertRequiredChanges() {
+    this.required = this.args.sequenceBlock.get('required').toString();
+    this.minimum = this.args.sequenceBlock.get('minimum');
+  }
+
+  @action
+  async saveCourse() {
```
jrjohnson, line 179: Async actions should always be tasks to avoid trying to call `this.args.sequenceBlock.set('sessions', []);` if the component is removed from the DOM between the promise starting and resolving.

app/components/curriculum-inventory-sequence-block-overview.js
```diff
@@ -1,111 +1,117 @@
...
+    this.description = sequenceBlock.description;
+    this.course = yield sequenceBlock.course;
+    this.minimum = sequenceBlock.minimum;
+    this.maximum = sequenceBlock.maximum;
+    this.sessions = yield this.getSessions(this.course);
+    this.linkableCourses = yield this.getLinkableCourses(this.report, this.course);
+  }
+
+  @restartableTask
+  *reload() {
```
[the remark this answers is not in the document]
stopfstedt: load/reload:format(jpeg):mode_rgb():quality(90)/discogs-images/R-10032307-1490451654-5013.jpeg.jpg)

app/components/curriculum-inventory-sequence-block-overview.js
stopfstedt: i updated all async actions to droppable task, this works out just fine. 

however. when saving linked courses in the ui, the console now shows a new warning. 

> ember-concurrency detected a potentially hazardous "self-cancel loop" between parent task `saveData` and child task `saveCourse`. If you want child task `saveCourse` to be canceled when parent task `saveData` is canceled, please change `.perform()` to `.linked().perform()`. If you want child task `saveCourse` to keep running after parent task `saveData` is canceled, change it to `.unlinked().perform()`

this _only_ happens on `saveCourse()`, the other edit-in-place controls do not throw this warning. seems to be working without problems, but may be something that needs to be handled in the `<EditableField> component, which lives in `common`.

jrjohnson: I vaguely recall seeing this before and not fixing it create https://github.com/ilios/common/issues/1577 so we're tracking it this time.

PR #5520, opened 2020-08-19
copy title attribute when copying py objectives.

stopfstedt:


stopfstedt:
follow-up item: https://github.com/ilios/frontend/issues/5521

PR #5549, opened 2020-09-03
Use sessionObjectives in session report

jrjohnson:
Was still using the old session.objectives relationship which doesn't
exist any more.

Refs ilios/ilios#3097


stopfstedt (approved):
"the one that got away". LGTM

Issue #5563, opened 2020-09-10
add input controls and output for learner group "needs accommodation" flag


stopfstedt:
companion ticket to https://github.com/ilios/ilios/issues/3083

saschaben:
really bad mockup of what we need here. mouseover on icons would display "members of this group require accommodation". 
[Question: do we display icon for the entire tree, or just the subgroup it is set to (when set to a subgroup)? I'm thinking only the subgroup.]

![access1](https://user-images.githubusercontent.com/1753431/92790604-c8d47c00-f360-11ea-8e4d-305a4a647c93.png)
![access2](https://user-images.githubusercontent.com/1753431/92790609-c96d1280-f360-11ea-84d7-4d3d83b1fba0.png)

here we could do a check box or a toggle, and may want to provide some additional visual cue for when selection is true (eg the label goes bold, or something):
![access3](https://user-images.githubusercontent.com/1753431/92790611-c96d1280-f360-11ea-8522-847d590fbced.png)

saschaben:
One more. Also, regarding the question above: maybe we can add an icon to the subgroup count, if there is a group in the subtree that requires, but the parent doesn't...?
![access4](https://user-images.githubusercontent.com/1753431/92792245-39c86380-f362-11ea-9601-f587935eb85e.png)

stopfstedt:
@saschaben thanks for the mocks, these are great.

PR #5576, opened 2020-09-14
Adds input controls for learner group url

stopfstedt:
Selection_169

fixes https://github.com/ilios/frontend/issues/5563

### DIFF
--- a/app/components/learnergroup-summary.hbs
+++ b/app/components/learnergroup-summary.hbs
@@ -79,6 +79,44 @@
         {{/if}}
       </div>
     </div>
+    <div class="block defaulturl">
+      <label>
+        {{t "general.defaultUrl"}}:
+      </label>
+      <span>
+        {{#if this.canUpdate}}
+          <EditableField
+            @value={{if this.url this.url (t "general.clickToEdit")}}
+            @save={{action "changeUrl"}}
+            @close={{action "revertUrlChanges"}}
+            @saveOnEnter={{true}}
+            @closeOnEscape={{true}} as |isSaving|
+          >
+            <input
+              type="text"
+              value={{this.url}}
+              oninput={{action (mut this.url) value="target.value"}}
+              disabled={{this.isSaving}}
+              onkeyup={{action "addErrorDisplayFor" "url"}}
+            >
+            {{#if
+              (and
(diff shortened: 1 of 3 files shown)


jrjohnson (approved):

saschaben:
Looks like the input form field is not auto-refreshing to text after hitting the green check to save; a click out of the form field allows the form input to refresh to text link.

stopfstedt:
> Looks like the input form field is not auto-refreshing to text after hitting the green check to save; a click out of the form field allows the form input to refresh to text link.

good find, although caused by something different. turns out that the old form validations library (`ember-cp-validations`) we're still using here is accepting URLs without protocol, but the backend is rejecting this. weak!

Selection_165

will see if i can configure us out of this calamity, and if that doesn't work then check into other options.

stopfstedt:
i'll need to octanify this component b/c of the flaw in the currently used validation library.

stopfstedt:
needs re-review and regression testing.

stopfstedt:
Selection_170

the link validation issue has been fixed.

jrjohnson (approved):

saschaben:
ok. the spinner issue with user management is noted. we should clean that up someday.

On point though: If we are going to require fully formed urls with protocols explicitly entered, we need to provide that information in the validation prompt, eg. "`a valid url is required ('https://mysmellysocks.com')`", or provide the pre-filled "`https://`" prompt in the form field as we do with learning materials. Otherwise it's a little confusing (if it confused me, it's guaranteed to confuse users when they plug in "www.thisplace.com" and they cannot save).

stopfstedt:
> ok. the spinner issue with user management is noted. we should clean that up someday.
> 
> On point though: If we are going to require fully formed urls with protocols explicitly entered, we need to provide that information in the validation prompt, eg. "`a valid url is required ('https://mysmellysocks.com')`", or provide the pre-filled "`https://`" prompt in the form field as we do with learning materials. Otherwise it's a little confusing (if it confused me, it's guaranteed to confuse users when they plug in "www.thisplace.com" and they cannot save).

fair point. i'm going to pre-fill with `https://` then, as we do with URLs in the offering form. 
Selection_177

saschaben:
FTW!

Issue #5579, opened 2020-09-15
Relabel Learner Group Default URL input field


stopfstedt:
name this "Virtual Learning Link".

follows https://github.com/ilios/frontend/pull/5576

stopfstedt:
done.

PR #5581, opened 2020-09-15
added translations - default virtual learning link

dartajax:
As the title states - I added translations for default virtual learning link - this will be used for small groups so they can have a pre-set value for this very important new field.

refs #5579

### DIFF
--- a/translations/en-us.yaml
+++ b/translations/en-us.yaml
@@ -93,6 +93,7 @@ general:
   dashboard: Dashboard
   deactivate: Deactivate
   defaultInstructors: "Default Instructors"
+  defaultVirtualLearningLink: "Default Virtual Learning Link"
   defaultLocation: "Default Location"
   departments: Departments
   details: "Course Details"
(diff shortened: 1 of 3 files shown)


translations/en-us.yaml
```diff
@@ -93,6 +93,7 @@ general:
   dashboard: Dashboard
   deactivate: Deactivate
   defaultInstructors: "Default Instructors"
+  defaultVirtualLearningLink: "Default Virtual Learning Link"
```
stopfstedt, line 96: `defaultVirtualLearningLink` should be below `defaultLocation`.

translations/es.yaml
```diff
@@ -93,6 +93,7 @@ general:
   dashboard: Tablero
   deactivate: Desactivar
   defaultInstructors: "Instructores Predeterminados"
+  defaultVirtualLearningLink: "Enlace de Aprendizaje Virtual Predeterminado"
```
stopfstedt, line 96: see comment above.

stopfstedt (changes requested):

translations/es.yaml
dartajax: oh oops - I changed it from a different title and then it got out of alphabetization - I will fix this.

stopfstedt (approved):
LGTM

PR #5616, opened 2020-09-30
[Test]: flakey test with save

snewcomer:
on master.

Screen Shot 2020-09-29 at 10 08 26 PM

Seems to be flakey b/c recent commits do not suggest this was failing in the CI runs.  But pulling and running locally will fail (most of the time).

https://github.com/emberjs/data/pull/7337

### DIFF
--- a/tests/integration/components/user-profile-ics-test.js
+++ b/tests/integration/components/user-profile-ics-test.js
@@ -1,6 +1,6 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { render, click } from '@ember/test-helpers';
+import { render, click, settled } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
 import {
   triggerSuccess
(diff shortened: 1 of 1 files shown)


tests/integration/components/user-profile-ics-test.js
```diff
@@ -43,6 +43,7 @@ module('Integration | Component | user profile ics', function(hooks) {
 
     await render(hbs`<UserProfileIcs @isManaging={{true}} @user={{this.user}} @setIsManaging={{noop}} />`);
     await click('.refresh-key');
+    await settled();
```
snewcomer, line 46: I'm guessing our test harness awaits this chain of async operations.  But we must be missing something 🤔 

```
  @dropTask
  *refreshKey() {
    const token = yield this.randomToken(this.args.user.id);
    this.args.user.set('icsFeedKey', token);
    yield this.args.user.save();
    this.args.setIsManaging(false);
    this.hasSavedRecently = true;
    yield timeout(500);
    this.hasSavedRecently = false;
  }
```

jrjohnson: IIRC I pulled this `settled()` out a week or so ago since it  didn't seem to be needed anymore and tests ran, but I'm not fussed about it being back. Thanks!

PR #5625, opened 2020-10-06
octanifies CI reporting table components.

stopfstedt:


dartajax (changes requested):
Would it be possible to have it navigate so the "Back to Table of Contents" link appears at the top of the page rather than in the middle of the page?

dartajax:
I noticed when testing this on demo3, the "Back to Table of Contents" link is not initially visible at all. The user needs to scroll up slightly to find it.

stopfstedt:
> Would it be possible to have it navigate so the "Back to Table of Contents" link appears at the top of the page rather than in the middle of the page?

negative. we're using a different method for scrolling those tables into view now, getting rid of hard-wired element ids. mid-page scroll is the trade-off for this improvement.

dartajax (approved):

app/components/curriculum-inventory-verification-preview.hbs
```diff
@@ -1,111 +1,128 @@
...
+    <span class="hidden" {{did-insert this.setScrollTarget "table3a"}}></span>
     <CurriculumInventoryVerificationPreviewTable3a
       @data={{this.tables.non_clerkship_sequence_block_instructional_time}}
     />
-    <a class="back-to-toc" href="#{{this.tocId}}">
+    <span class="link clickable back-to-toc" role="button" {{on "click" (fn this.scrollTo "toc")}}>
       {{t "general.backToTableOfContents"}}
-    </a>
+    </span>
+    <span class="hidden" {{did-insert this.setScrollTarget "table3b"}}></span>
```
jrjohnson, line 83: I think (but am not certain) that you could put this `did-insert` onto the component as:
```
<CurriculumInventoryVerificationPreviewTable4
   @data={{this.tables.instructional_method_counts}}
  {{did-insert this.setScrollTarget "table3b"}}
/>
```
as long as the CurriculumInventoryVerificationPreviewTable4 has an `...attributes` it will apply `did-insert` where `...attributes` is in the component.

jrjohnson (commented):
Maybe a way to clean up the markup a bit, or maybe a bad idea on my part...

app/components/curriculum-inventory-verification-preview.hbs
stopfstedt: this works. brilliant!

jrjohnson (approved):

PR #5629, opened 2020-10-09
Update common 51.0.0

jrjohnson:
Includes a major version bump for `ember-page-title` which eliminates the need for `ember-cli-head` in favor of controlling the title directly.


stopfstedt (approved):
LGTM

PR #5644, opened 2020-10-13
improves state handling on "select all" toggle.

### LINKED ISSUE #5631: learner group cohort members all selected when filtering.
image

_Originally posted by @dartajax in https://github.com/ilios/frontend/pull/5630#issuecomment-706409040_

This applies to the learner group user manager as well

Selection_004

stopfstedt:
fixes #5631

### DIFF
--- a/app/components/learnergroup-cohort-user-manager.js
+++ b/app/components/learnergroup-cohort-user-manager.js
@@ -16,6 +16,15 @@ export default class LearnergroupCohortUserManagerComponent extends Component {
     return this.args.sortBy.search(/desc/) === -1;
   }
 
+  get selectableUsers() {
+    if (this.currentUser.isRoot) {
+      return this.args.users;
+    }
+    return this.args.users.filter((user) => {
+      return user.enabled;
+    });
+  }
+
   get filteredUsers() {
     const filter = this.filter.trim().toLowerCase();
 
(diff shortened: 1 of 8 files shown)


dartajax:
I like it. Great work. 👍

dartajax (approved):

app/components/learnergroup-cohort-user-manager.js
```diff
@@ -49,13 +58,11 @@ export default class LearnergroupCohortUserManagerComponent extends Component {
 
   @action
   toggleUserSelectionAllOrNone() {
-    const selectedUsers = this.selectedUsers.length;
-    const filteredUsers = this.filteredUsers.length;
-
-    if (selectedUsers >= filteredUsers) {
-      this.selectedUsers = [];
+    const unselectedFilteredUsers = this.filteredUsers.filter(user => { return ! this.selectedUsers.includes(user); });
```
jrjohnson, line 61: As a one liner this should be `const unselectedFilteredUsers = this.filteredUsers.filter(user => !this.selectedUsers.includes(user));`
Otherwise it should be split as:
```
const unselectedFilteredUsers = this.filteredUsers.filter(user => { 
  return ! this.selectedUsers.includes(user); 
});
```

app/components/learnergroup-user-manager.js
```diff
@@ -16,6 +16,15 @@ export default class LearnergroupCohortUserManagerComponent extends Component {
     return this.args.sortBy.search(/desc/) === -1;
   }
 
+  get selectableUsers() {
+    if (this.currentUser.isRoot) {
+      return this.args.users.mapBy("content");
+    }
+    return this.args.users.filter((user) => {
+      return user.get("enabled");
```
jrjohnson, line 24: This `get` isn't necessary, can use be `user.enabled`.

jrjohnson (changes requested):

app/components/learnergroup-cohort-user-manager.js
stopfstedt: handled.

app/components/learnergroup-user-manager.js
stopfstedt: nyet. `user` is an ember proxy object, so we still have to use the getter.
```
Error: Assertion Failed: You attempted to access the `enabled` property (of <(unknown):ember28001>).
Since Ember 3.1, this is usually fine as you no longer need to use `.get()`
to access computed properties. However, in this case, the object in question
is a special kind of Ember object (a proxy). Therefore, it is still necessary
to use `.get('enabled')` in this case.
```

jrjohnson: Ah ok. Sorry!

jrjohnson (approved):

PR #5645, opened 2020-10-14
a few more octanified components

stopfstedt:


app/components/curriculum-inventory-report-list.js
```diff
@@ -16,74 +17,69 @@ const ReportProxy = ObjectProxy.extend({
...
-   */
-  proxiedReports: computed('program.curriculumInventoryReports.[]', async function () {
-    const currentUser = this.currentUser;
-    const intl = this.intl;
-    const permissionChecker = this.permissionChecker;
-    const program = this.program;
-    const reports = await program.get('curriculumInventoryReports');
-    return reports.map(report => {
+  @action
+  load(element, [ reports ]) {
```
jrjohnson, line 35: Since this is sync and only setting the one property couldn't it just be a getter?  `get proxiedReports()`?

jrjohnson (changes requested):

app/components/curriculum-inventory-report-list.js
stopfstedt: handled. i rewired the component to take the reports as input directly.

app/templates/curriculum-inventory-reports.hbs
```diff
@@ -81,7 +81,7 @@
     <div class="list">
       {{#if (is-fulfilled this.reports)}}
         <CurriculumInventoryReportList
-          @program={{await this.selectedProgram}}
+          @reports={{get (await this.selectedProgram) "curriculumInventoryReports"}}
```
jrjohnson, line 84: I'm not sure this away is in the right place. Shouldn't it be `@reports={{await(get this.selectedProgram "curriculumInventoryReports")}}`?

stopfstedt: that doesn't work. to be on the safe side, i added another `await` for the linked reports.

```
@reports={{await (get (await this.selectedProgram) "curriculumInventoryReports")}}
```

jrjohnson (approved):

PR #5656, opened 2020-10-20
removes explicit ember-click-outside dependency.

stopfstedt:
we're getting that package via ilios-common.

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,6 +125,5 @@
   "pre-commit": [
     "lint"
   ],
-  "private": true,
-  "dependencies": {}
+  "private": true
 }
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -126,5 +125,6 @@
   "pre-commit": [
     "lint"
   ],
-  "private": true
+  "private": true,
+  "dependencies": {}
```
jrjohnson, line 129: On purpose?

stopfstedt: no, `npm uninstall ember-click-outside` bolted this one. remove?

stopfstedt: removed it.

jrjohnson: Yea, it's done that to me, maybe just leave it this time.

jrjohnson: or not 😁

jrjohnson (approved):

PR #5684, opened 2020-11-04
removes edit button from instructor-groups list.

### LINKED ISSUE #2632: Get Rid Of Edit Action Icon in Instructor Groups
Four reasons for this ...

1. The horizontal alignment is slightly off from the trash icon
2. In Learner Groups, it has already been removed and we should be consistent
3. The editing of Instructor Groups is accomplished by clicking on the Instructor Group title
4. It's way too close to the trash can.

image

stopfstedt:
fixes #2632

### DIFF
--- a/app/components/instructorgroup-list.hbs
+++ b/app/components/instructorgroup-list.hbs
@@ -38,13 +38,6 @@
             {{/if}}
           </td>
           <td class="text-right">
-            <LinkTo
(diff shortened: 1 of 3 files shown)


dartajax:
I am cool with that. 👍

dartajax (approved):
This works for me.

PR #5723, opened 2020-11-26
octanifies IliosUsers component.

### LINKED ISSUE #5722: Bulk New User cancel button not working
it should close the component.

Selection_068

stopfstedt:
fixes #5722

### DIFF
--- a/app/components/new-user.js
+++ b/app/components/new-user.js
@@ -146,7 +146,7 @@ export default class NewUserComponent extends Component {
     user = yield user.save();
     const authentication = this.store.createRecord('authentication', {
       user,
-      username: this.userName,
+      username: this.username,
       password: this.password,
     });
     yield authentication.save();
--- a/app/components/new-user.js
+++ b/app/components/new-user.js
@@ -75,10 +75,11 @@ export default class NewUserComponent extends Component {
   @restartableTask
   *loadSchools() {
     const store = this.store;
-    const schools = yield store.findAll('school', { reload: true });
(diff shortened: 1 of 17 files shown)


app/components/new-user.js
```diff
@@ -1,78 +1,166 @@
...
+  @restartableTask
+  *reload() {
+    this.currentSchoolCohorts = yield this.bestSelectedSchool?.cohorts;
+    this.cohorts = yield this.loadCohorts.perform(this.bestSelectedSchool);
+  }
+
+  @restartableTask
+  *loadSchools() {
+    const store = this.store;
+    const schools = yield store.findAll('school', { reload: true });
```
jrjohnson, line 78: An improvement may be possible here, because we always load all schools in https://github.com/ilios/frontend/blob/master/app/routes/application.js#L45 `peekRecord` could be used here synchronously replacing this with a simple `get schools` and possible making all async in this component a bit easier.

jrjohnson (commented):
One maybe improvement, looks good to me. I scanned the tests and didn't dig into them too thoroughly, anything I should be looking out for there?

app/components/new-user.js
stopfstedt: i changed it. the method will still have to be declared as either a task or an async function, since the perms check that follow the fetching of schools are async in nature.

jrjohnson: Oh, yea. I see that now. Ok. Accepted.

jrjohnson (approved):

PR #5787, opened 2021-01-07
give the my-courses table some room inside its container.

### LINKED ISSUE #5783: Make Padding on My Courses Equal to that of My Reports
image

stopfstedt:
Selection_154

~~that should be close enough (not exactly tho). my-reports uses a list to lay out its items, whereas the my-courses uses a table to do so for its items. so there are subtle spacing/padding differences that i don't want to get into.~~
this required changing the my-courses layout from using a table to using a list.

fixes #5783


dartajax (changes requested):
Can you ensure that My Reports and My Courses have the same padding? Now My Courses has more than My Reports.

dartajax (approved):

Issue #5753, opened 2020-12-11
Accessing Visualizations Should Not Affect Ilios Menu


dartajax:
What I am saying here is that when a user accesses our Visualizations - which are accessed via Courses and Sessions, there is no reason for `Courses and Sessions` to not be highlighted on the Ilios flyout menu.

Here it is shown below ...

![image](https://user-images.githubusercontent.com/7493331/101962447-0b585f80-3bc1-11eb-9fd5-724de1749413.png)

It should still be like ...

![image](https://user-images.githubusercontent.com/7493331/101962477-21feb680-3bc1-11eb-9f3d-c579c9b9b390.png)

... since the user never actually left `Courses and Sessions`.

PR #5795, opened 2021-01-13
sets navigation state on routes previously not considered

stopfstedt:
i made a pass over all of our top-level routes and added them to the navigation with them where they were missing and could be mapped to a nav item.

fixes https://github.com/ilios/frontend/issues/5753


jrjohnson (approved):
This is fantastic. I didn't even know this worked at it's super clean. Love it.

PR #5810, opened 2021-01-20
adds term counts to vocab and term manager.

### LINKED ISSUE #5752: Provide Counts of Sub-Terms in Vocabulary Management
Terms can be nested below the highest level.

image

After clicking as shown above, the user receives the following information. As shown above, there are 7 terms / sub-terms in this vocabulary.

image

There is no way a user would know where the nested terms are nested. There should be numerical indicators to help with this task. 

Something like ...

image

Here it is drilled down another level - this should apply at every level of navigation.

image

We also need to verify the counts are correct - I was only seeing six sub-terms in this example. Anyway it would be helpful.

stopfstedt:
fixes #5752

### DIFF
--- a/app/components/school-vocabulary-manager.hbs
+++ b/app/components/school-vocabulary-manager.hbs
@@ -44,6 +44,7 @@
       {{else}}
         {{this.title}}
       {{/if}}
+      ({{t "general.countTotal" total=(get (await @vocabulary.terms) "length")}})
     </div>
     <h5>
       {{t "general.terms"}}:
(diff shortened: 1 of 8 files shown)


dartajax:
Is there any possibility of not showing the sub counts if the answer is zero like in these cases?

image

dartajax (changes requested):
image

stopfstedt:
@dartajax - how's this?

image

dartajax (approved):
looks good to me

Issue #5548, opened 2020-09-02
Make Calendar Events Clickable In Admin Console User Calendar


dartajax:
As shown below, it is handy to be able to mouse over and reveal details of an offering on the user profile page in the Admin Console. However, it would be useful to be able to click on the event to go to the Event Detail view right from this calendar review page in the Admin Console.

![image](https://user-images.githubusercontent.com/7493331/92012230-07cf5580-ed01-11ea-8283-4b897c90293b.png)

Refs #5078

PR #5814, opened 2021-01-21
removes ability to click through to event details from user admin calendar

### LINKED ISSUE #5813: Admin Calendar View of User's Events - Events Should Be Disabled or Allow Access to View
Below is a screen shot from the Admin Console. The user 17223 is not the user who is logged in. The logged in admin user is not directly associated with the Calendar Events and when the logged in admin user clicks on any of the events displayed, they will receive the coveted "Rats ..." error. It would be best if the admin user could see and view the Event Details screen in order to research the user's calendar if that was the intent. If this is not possible or is considered to be giving too much information to the admin user (only limited people should have this advanced level of permission), we can disable the clickable characteristics of these calendar events inside the Admin Console.

image

image

stopfstedt:
reverts #5705
refs #5548
fixes #5813

### DIFF
--- a/app/components/user-profile-calendar.hbs
+++ b/app/components/user-profile-calendar.hbs
@@ -21,9 +21,8 @@
     <IliosCalendarWeek
       @calendarEvents={{this.calendarEvents}}
       @date={{date}}
-      @areEventsSelectable={{true}}
+      @areEventsSelectable={{false}}
       @areDaysSelectable={{false}}
-      @selectEvent={{@selectEvent}}
     />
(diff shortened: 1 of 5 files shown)


jrjohnson (changes requested):
Code it good, but is there a way to add a test to prevent us going back and forth on this again?

stopfstedt:
> Code it good, but is there a way to add a test to prevent us going back and forth on this again?

handled.

jrjohnson (approved):

PR #5822, opened 2021-01-28
adds padding to myreport-title for better vertical alignment.

### LINKED ISSUE #5821: Line Up MyReports Better With Week at a Glance
image

It would look better more like this ...

image

This may not be the same if there are results and also since the space is so small when there are results, they may need all of the space available to them.

stopfstedt:
image

fixes #5821

### DIFF
--- a/app/components/dashboard-myreports.hbs
+++ b/app/components/dashboard-myreports.hbs
@@ -75,34 +75,32 @@
           </div>
         {{/if}}
         {{#if (is-array this.reportResultsList)}}
-          {{#if this.reportResultsList.length}}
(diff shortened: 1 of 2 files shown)


dartajax (changes requested):
When the result set is empty the "None" label should align in a similar manner to when the results. 

Sorry to be a bother about this. With results, it looks great.

image

stopfstedt:
> When the result set is empty the "None" label should align in a similar manner to when there are results.
> 
> Sorry to be a bother about this. With results, it looks great.

good find, and no bother at all. i made some improvements, check now.

dartajax (approved):
much better 💯

Issue #5844, opened 2021-02-05
replace ember-simple-auth mixins


stopfstedt:
```
WARN: DEPRECATION: Ember Simple Auth: The session service's events API is deprecated; to add custom behavior to the authentication or invalidation handling, override the handleAuthentication or handleInvalidation methods. [deprecation id: ember-simple-auth.events.session-service]
```

companion ticket to https://github.com/ilios/common/issues/1915, also see https://github.com/ilios/common/pull/1916

stopfstedt:
closed via #5845

PR #5845, opened 2021-02-06
Replaces deprecated ember-simple-auth mixins.

stopfstedt:
refs #5844

### DIFF
--- a/app/routes/instructor-group.js
+++ b/app/routes/instructor-group.js
@@ -1,7 +1,7 @@
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
 
-export default class DashboardRoute extends Route {
+export default class InstructorGroupRoute extends Route {
   @service permissionChecker;
   @service session;
 
--- a/app/routes/instructor-group.js
+++ b/app/routes/instructor-group.js
@@ -1,17 +1,24 @@
 import Route from '@ember/routing/route';
-import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
 import { inject as service } from '@ember/service';
 
-export default Route.extend(AuthenticatedRouteMixin, {
-  permissionChecker: service(),
-  canUpdate: false,
+export default class DashboardRoute extends Route {
+  @service permissionChecker;
+  @service session;
+
+  canUpdate = false;
+
+  beforeModel(transition) {
+    this.session.requireAuthentication(transition, 'login');
+  }
+
   async afterModel(instructorGroup) {
     const permissionChecker = this.permissionChecker;
     const canUpdate = await permissionChecker.canUpdateInstructorGroup(instructorGroup);
     this.set('canUpdate', canUpdate);
-  },
+  }
+
   setupController(controller, model) {
-    this._super(controller, model);
+    super.setupController(controller, model);
     controller.set('canUpdate', this.canUpdate);
   }
-});
+}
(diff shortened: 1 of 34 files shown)


app/routes/instructor-group.js
```diff
@@ -1,17 +1,24 @@
 import Route from '@ember/routing/route';
-import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
 import { inject as service } from '@ember/service';
 
-export default Route.extend(AuthenticatedRouteMixin, {
-  permissionChecker: service(),
-  canUpdate: false,
+export default class DashboardRoute extends Route {
```
jrjohnson, line 4: Should be `instructorGroupRoute`

app/routes/login.js
```diff
@@ -3,26 +3,25 @@ import { inject as service } from '@ember/service';
...
+  @service iliosConfig;
+  @service session;
 
-  noAccountExistsAccount: null,
-  noAccountExistsError: false,
+  noAccountExistsAccount = null;
+  noAccountExistsError = false;
 
-  beforeModel(transition){
-    this._super(transition);
```
jrjohnson, line 18: I'm not sure what this did, but wanted to make sure you dropped it on purpose.

app/routes/program-year-visualizations.js
```diff
@@ -1,17 +1,25 @@
...
 import Route from '@ember/routing/route';
-import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
 
-export default Route.extend(AuthenticatedRouteMixin, {
-  store: service(),
+export default class ProgramYearVisualizationsRoute extends Route {
+  @service store;
+  @service session;
+
   // eslint-disable-next-line ember/avoid-leaking-state-in-ember-objects
```
jrjohnson, line 8: I'm betting this eslint modifier can go away now that this is a class.

app/routes/session-type-visualize-terms.js
```diff
@@ -1,17 +1,25 @@
 /* eslint-disable ember/avoid-leaking-state-in-ember-objects */
```
jrjohnson, line 1: Probably removable now.

jrjohnson (changes requested):

app/routes/program-year-visualizations.js
stopfstedt: :heavy_check_mark:

app/routes/instructor-group.js
stopfstedt: :heavy_check_mark:

app/routes/login.js
stopfstedt: hmm, according to the docs, it was necessary to invoke the parent method in order to prevent infinite redirects.

```
Checks whether the session is authenticated and if it is aborts the current
    transition and instead transitions to the
    {{#crossLink "UnauthenticatedRouteMixin/routeIfAlreadyAuthenticated:property"}}{{/crossLink}}.

   __If `beforeModel` is overridden in a route that uses this mixin, the route's
   implementation must call `this._super(...arguments)`__ so that the mixin's
   `beforeModel` method is actually executed.
```

supposedly, this can be ditched now without replacement. in click-testing, i was unable to trigger an infinite redirect.

app/routes/session-type-visualize-terms.js
stopfstedt: :heavy_check_mark:

jrjohnson (approved):

PR #5849, opened 2021-02-10
Cleanup Application Route

### LINKED ISSUE #5848: replace deprecated ember-simple-auth ApplicationRouteMixin
this is a real bugbear. i took a stab at this already but couldn't pull it off so far.

we'll need to find replacements for the [`sessionInvalidated()`](https://github.com/ilios/frontend/blob/master/app/routes/application.js#L97) hook, and the event callback for [`authenticationSucceeded`](https://github.com/ilios/frontend/blob/master/app/routes/application.js#L63) event.

we _may_ be able to get rid of  `sessionInvalidated()` by overriding the `session` service with our own and implementing [`handleInvalidation()`](http://ember-simple-auth.com/api/classes/SessionService.html#method_handleInvalidation), but that's untested. 

what really stumps me is how  replace `authenticationSucceeded`, as we attach the event handler within a route-handler lifecycle hook[ `activate()`](https://api.emberjs.com/ember/3.24/classes/Route/methods/activate?anchor=activate). i don't see a clean way to transition this flow of control, so service worker registration may have to be handled differently altogether, perhaps de-coupled form authentication.

refs https://github.com/ilios/frontend/pull/5845

jrjohnson:
Override ESA session service with out own which allows us to ditch events which are now deprecated by ESA and
instead handle our custom auth/setup/teardown logic directly here.

Modernize and cleanup the application route.

Fixes #5848

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -1,111 +1,70 @@
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
-import ApplicationRouteMixin from 'ember-simple-auth/mixins/application-route-mixin';
-import config from 'ilios/config/environment';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
 import { all } from 'rsvp';
 import * as Sentry from '@sentry/browser';
 import { loadPolyfills } from 'ilios-common/utils/load-polyfills';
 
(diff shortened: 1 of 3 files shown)


stopfstedt (approved):
LGTM

PR #5843, opened 2021-02-05
replace obsolete academicYearTitle with title attr.

stopfstedt:
refs https://github.com/ilios/ilios/issues/3272

PR #5857, opened 2021-02-11
ilios-common version bump.

stopfstedt:


jrjohnson:
I think this may need to go inside #5843 so both can work.

app/controllers/courses.js
```diff
@@ -100,7 +100,7 @@ export default Controller.extend({
   selectedYear: computed('model.years.[]', 'yearTitle', function() {
     const years = this.get('model.years');
     if(isPresent(this.yearTitle)){
-      return years.find(year => year.get('title') === parseInt(this.yearTitle, 10));
+      return years.find(year => year.get('id') === this.yearTitle);
```
jrjohnson, line 103: Soon we'll have linting for it, but for now can you make this `year.id` and remove the `get`?

jrjohnson (changes requested):

app/controllers/courses.js
stopfstedt: :heavy_check_mark:

app/components/new-course.js
```diff
@@ -1,80 +1,76 @@
...
-  currentSchool: null,
-  currentYear: null,
-  selectedYear: null,
-  title: null,
-  cancel() {},
-  save() {},
-  years: Object.freeze(YEARS),
+  @tracked @NotBlank() selectedYear;
+  @tracked @NotBlank() @Length(3, 200) title;
+  @tracked years = Object.freeze(YEARS);
```
jrjohnson, line 26: This `Object.freeze` shouldn't be necessary anymore since this component is now a JS class and they don't share state the way ember object based components did.

app/controllers/courses.js
```diff
@@ -40,18 +40,18 @@ export default Controller.extend({
       return [];
     }
 
-    const yearTitle = selectedYear.title;
+    const year = parseInt(selectedYear.id, 10);
```
jrjohnson, line 43: Isn't this already an `int`?

app/templates/courses.hbs
```diff
@@ -33,8 +33,8 @@
         data-test-year-filter
       >
         {{#each (sort-by "title:desc" @model.years) as |year|}}
-          <option value={{year.year}} selected={{is-equal year this.selectedYear}}>
-            {{year.academicYearTitle}}
+          <option value={{year.id}} selected={{is-equal year this.selectedYear}}>
```
jrjohnson, line 36: This sets a variable on the controller called `yearTitle`. I think using the ID is better, but that variable name should change.

tests/integration/components/new-course-test.js
```diff
@@ -0,0 +1,177 @@
...
+    assert.ok(component.years(0).selected);
+    assert.notOk(component.years(1).selected);
+    assert.notOk(component.years(2).selected);
+    assert.notOk(component.years(3).selected);
+    assert.notOk(component.years(4).selected);
+    assert.notOk(component.years(5).selected);
+  });
+
+  // currently doesn't work since the injected ilios-config mock service doesn't seem to get picked up [ST 2021/02/12]
+  skip('year options show range if applicable', async function (assert) {
```
jrjohnson, line 71: The best way to inject config is to mock that API call instead of the service like [this](https://github.com/ilios/frontend/blob/6c6c7e5eccb582b400fa92d0c12593c2205d1d26/tests/acceptance/search-test.js#L14)

jrjohnson (changes requested):
Great test coverage!

app/controllers/courses.js
stopfstedt: it's not, it's a string.

app/components/new-course.js
stopfstedt: :heavy_check_mark:

app/controllers/courses.js
jrjohnson: ✔️

app/templates/courses.hbs
stopfstedt: :heavy_check_mark:

tests/integration/components/new-course-test.js
stopfstedt: this saved the day! :heavy_check_mark:

tests/acceptance/courses-test.js
```diff
@@ -5,6 +5,7 @@ import moment from 'moment';
 import page from 'ilios/tests/pages/courses';
 import { setupApplicationTest } from 'ember-qunit';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+import { freezeDateAt, unfreezeDate } from 'ember-mockdate-shim';
```
jrjohnson, line 8: 🔥 !!!!!

jrjohnson (approved):

package.json
```diff
@@ -81,6 +81,7 @@
     "ember-load-initializers": "^2.1.2",
     "ember-maybe-import-regenerator": "^0.1.6",
     "ember-metrics": "1.2.0",
+    "ember-mockdate-shim": "^0.1.0",
```
jrjohnson, line 84: I can't figure out if this adds anything interesting on top of https://github.com/boblauer/MockDate which is probably directly usable.

stopfstedt: i don't think it does, but i like the freeze/unfreeze helpers.

jrjohnson: 🆗 🆒

Issue #5866, opened 2021-02-16
make remaining academic-year toggles configuration-aware


stopfstedt:
![image](https://user-images.githubusercontent.com/1410427/108134765-a3644080-706b-11eb-8e26-c638a2fbafde.png)
![image](https://user-images.githubusercontent.com/1410427/108134887-deff0a80-706b-11eb-992c-da19e0ad8498.png)
![image](https://user-images.githubusercontent.com/1410427/108781865-6d1d3a00-751f-11eb-9b51-286548a1ad7b.png)

![image](https://user-images.githubusercontent.com/1410427/108134794-b0812f80-706b-11eb-9230-bfb100a491db.png)
![image](https://user-images.githubusercontent.com/1410427/108743127-55c55900-74ed-11eb-8fe3-491132009a1a.png)
![image](https://user-images.githubusercontent.com/1410427/108743212-6970bf80-74ed-11eb-98c3-ca5137b5b487.png)
![image](https://user-images.githubusercontent.com/1410427/108743043-3dedd500-74ed-11eb-924f-ef4e38ee9bfb.png)

refs https://github.com/ilios/ilios/issues/3272

jrjohnson:
Adding course rollover to this list and suggesting that maybe we throw this in a service to make it easier to keep track of?

PR #5868, opened 2021-02-18
consider academic year config for program-years and ci reports

stopfstedt:
fixes https://github.com/ilios/frontend/issues/5866
fixes https://github.com/ilios/frontend/issues/5878


app/components/learnergroup-header.js
```diff
@@ -1,51 +1,33 @@
...
-  title: [
-    validator('presence', true),
-    validator('length', {
-      min: 3,
-      max: 60
-    })
-  ]
-});
+@validatable
+export default class ProgramyearHeaderComponent extends Component {
```
jrjohnson, line 8: Should be LearnergroupHeaderComponent

app/routes/curriculum-inventory-reports.js
```diff
@@ -14,9 +14,4 @@ export default class CurriculumInventoryReportsRoute extends Route {
     const store = this.store;
     return store.findAll('school');
   }
-
-  setupController(controller, model) {
-    super.setupController(controller, model);
-    controller.set('sortByTitle', ['title']);
```
jrjohnson, line 20: Love getting rid of these. They always bothered me.

app/templates/curriculum-inventory-reports.hbs
```diff
@@ -1,95 +1,12 @@
...
-          @remove={{action "removeCurriculumInventoryReport"}}
-          @setSortBy={{action (mut this.sortReportsBy)}}
-        />
-      {{else}}
-        <PulseLoader />
-      {{/if}}
-    </div>
-  </section>
-</section>
+<CurriculumInventoryReports
```
jrjohnson, line 2: ❤️

jrjohnson (changes requested):
One super minor thing. I'm also going back in forth over the structure of the years. It's repeated so many times I wonder if a component to display year that loaded the config internally would be better than repeating the config load and display logic in so many places. I'm ok with this being a future improvement though.

app/components/learnergroup-header.js
stopfstedt: :heavy_check_mark:

stopfstedt:
> I'm also going back in forth over the structure of the years. It's repeated so many times I wonder if a component to display year that loaded the config internally would be better than repeating the config load and display logic in so many places. I'm ok with this being a future improvement though.

we could make another pass over this, should be easier this time around since we can now grep for the app config value's name. i've been seeing some common functionality that could be repurposed for the academic-year drop downs. outside of that, the labeling seems to be all over the place.

jrjohnson (approved):

PR #5869, opened 2021-02-18
Remove preloading for application route

jrjohnson:
This isn't needed anymore as the values we need for the navigation are
available synchronously in the token.


jrjohnson:
Requires #5870

stopfstedt (approved):

Issue #5886, opened 2021-02-25
Ensure New Program Year Does Not Contain Previously Used Years


dartajax:
![image](https://user-images.githubusercontent.com/7493331/109229665-79e89a80-7778-11eb-9fdf-ab5f08fd7058.png)

We already have used 2018-2019 - it should not be in the list for selection as a "new" program year. Any number of duplicates can be created. This appears to be some sort of regression bug. We should only have future and unused years available for selection. Multiple program years in a program of the same year can and could possibly cause severe confusion if this were to occur and be allowed.

![image](https://user-images.githubusercontent.com/7493331/109229777-a8667580-7778-11eb-850e-23df3b86f29f.png)

Issue #5521, opened 2020-08-19
provide test coverage on creating a new program year


stopfstedt:
all kinds of things going on in that process - copying of objectives, etc.

cover this with tests.

PR #5887, opened 2021-02-26
reload component if given list of program years changes.

stopfstedt:
fixes https://github.com/ilios/frontend/issues/5886

### DIFF
--- a/app/components/new-programyear.js
+++ b/app/components/new-programyear.js
@@ -4,6 +4,7 @@ import { action } from '@ember/object';
 import { dropTask } from 'ember-concurrency';
 
 export default class NewProgramyearComponent extends Component {
+  @tracked allYears;
   @tracked year;
   @tracked availableAcademicYears = [];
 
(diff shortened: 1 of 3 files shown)


stopfstedt:
_side note:_ we need more/better test coverage surrounding program-year creation.  refs https://github.com/ilios/frontend/issues/5521

app/components/new-programyear.js
```diff
@@ -4,6 +4,7 @@ import { action } from '@ember/object';
 import { dropTask } from 'ember-concurrency';
 
 export default class NewProgramyearComponent extends Component {
+  @tracked allYears;
```
jrjohnson, line 7: Doesn't matter in any way, but since this value never changes it doesn't actually need to be `@tracked`.

jrjohnson (approved):

app/components/new-programyear.js
stopfstedt: :heavy_check_mark:

dartajax (approved):
It looks good to me.

PR #5897, opened 2021-03-03
Back on the Octane Train

jrjohnson:
Requires https://github.com/ilios/common/pull/1994 to fix an issue where the `@use` resource couldn't be passed down into the `UserSearch` component because it used a computed property proxy.


app/components/program-year/list-item.js
```diff
@@ -0,0 +1,89 @@
...
+import { inject as service } from '@ember/service';
+import { dropTask } from 'ember-concurrency';
+
+export default class ProgramYearListItemComponent extends Component {
+  @service permissionChecker;
+  @service currentUser;
+
+  @tracked showRemoveConfirmation = false;
+
+  @use program = new ResolveAsyncValue(() => [this.args.programYear.program]);
```
stopfstedt, line 15: yes!!!!

stopfstedt (approved):
fixes https://github.com/ilios/frontend/issues/5521

LGTM!!!

app/components/program-year/list-item.js
jrjohnson: Right? It's so pretty!!!!

PR #5906, opened 2021-03-07
Fix flakey upload test

jrjohnson:
Waiting for settled wasn't cutting it in some cases. Instead we should
wait for the table with valid users to appear and then move on.


stopfstedt (approved):
LGTM.

Issue #5900, opened 2021-03-05
Remove usage of ember-modal-dialog


jrjohnson:
This addon isn't maintained much and it has trouble with the new Embroider build system. We're only using it in a few places, so let's transition away. Most of these cases are in `WaitSaving` which may not be necessary anymore with options in our API to save multiple items at the same time.

stopfstedt:
remaining instances of <WaitSaving> usage:

```
app/components/learner-groups/root.hbs:  <WaitSaving
app/components/curriculum-inventory/report-details.hbs:    <WaitSaving>{{t "general.finalizingReport"}}</WaitSaving>
app/components/bulk-new-users.hbs:        <WaitSaving
app/components/assign-students.hbs:    <WaitSaving
app/components/learnergroup-subgroup-list.hbs:    <WaitSaving

```

stopfstedt:
non-issue at this point - `<WaitSaving />` has shed its dependency on ember-modal-dialog and is good to use again.

PR #5907, opened 2021-03-08
Save Program Year Objectives in One Go

jrjohnson:
As part of the quest for #5900 I've removed the wait saving indicator and saved all new program year objectives in a single chunk.


stopfstedt (approved):
click tested and reviewed. LGTM.

PR #5922, opened 2021-03-12
replaces sequence block object proxy with list-item subcomponent

stopfstedt:


app/components/curriculum-inventory/sequence-block-list.js
```diff
@@ -1,109 +1,39 @@
...
+  *save(block) {
+    this.editorOn = false;
+    this.savedBlock = yield block.save();
+    // adding/updating a sequence block will have side-effects on its siblings if the given block is nested
+    // inside an "ordered" sequence block. they all get re-sorted server-side.
+    // therefore, we must reload them here in order to get those updated sort order values.
+    // [ST 2021/03/16]
+    if (this.args.parent) {
+      const blocks = yield this.args.parent.children;
+      yield all(blocks.invoke('reload'));
```
jrjohnson, line 36: This will load them one at a time in a bunch of requests, which is maybe ok depending on how many there are. Could also leverage JSON:API side loading to get all children in the same request with:

```
this.store.findRecord('curriculum_inventory_sequence_block', this.args.parent.id, {
   include: 'children',
   reload: true,
});
```

tests/integration/components/curriculum-inventory/sequence-block-list-test.js
```diff
@@ -1,550 +1,257 @@
...
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+import { component } from 'ilios/tests/pages/components/curriculum-inventory/sequence-block-list';
 
 module('Integration | Component | curriculum-inventory/sequence-block-list', function (hooks) {
   setupRenderingTest(hooks);
-
-  test('it renders with top-level sequence blocks', async function (assert) {
-    assert.expect(25);
-
-    const school = EmberObject.create({
```
jrjohnson, line 17: 😂

jrjohnson (approved):
Good stuff here, one suggestion for a possible future change.

jrjohnson:
I reviewed code, but did not click test. @dartajax please give CI stuff a run through.

dartajax (changes requested):
The report titles are not correctly formatted as links. See screen shots below for reference.

from the PR ...

image

from Production ...





image

stopfstedt:
> The report titles are not correctly formatted as links. See screen shots below for reference.

Thanks Dave. I fixed this now.

app/components/curriculum-inventory/sequence-block-list.js
stopfstedt: :heavy_check_mark:

dartajax (approved):
LGTM 🥇

app/components/curriculum-inventory/report-list-item.hbs
```diff
@@ -51,12 +52,13 @@
     <CurriculumInventory::PublicationStatus @item={{@report}} />
   </td>
   <td class="text-right report-actions" colspan="2" data-test-actions>
-    <FaIcon
+    <span role="button" {{on "click" (fn @edit @report)}}>
```
jrjohnson, line 55: I'd like to see this either be a real button or else a link with the icon. Adding a span that acts like a button is a workaround that is a bit of an anti-pattern to avoid doing it the right way and styling with CSS.

jrjohnson (changes requested):

PR #5936, opened 2021-03-19
Build with embroider

jrjohnson:
Unlocks splitting at routes so users are not required to load the entire
application every time. This is especially impactful for students who
don't need or have access to any of the administrative functions.

- [x] https://www.npmjs.com/package/webpack-retry-chunk-load-plugin

Before Route Splitting:
before

After:
after

Saves download time and significant time processing javascript as we're shipping way less javascript to parse.


dartajax (approved):
It seems snappy and responsive - no issues to report.

stopfstedt (approved):
LGTM

dartajax:
Do we want the more pronounced vertical bars as shown in the first screen shot from demo or the fainter ones in the PR? I think I like the thicker ones like we currently have in place on demo.

image

image

jrjohnson:
@dartajax this is actually a bug in production. In many places where we have a light grey border (here, and on the login form for example) that border isn't showing up and instead we're getting dark black. I have no idea why this is happening, but the fix is to merge this change. Then, if you want, we could create a ticket to modify the calendar day border to be darker.

jrjohnson:
Merging this myself as I've removed the node 14.x tests, I'll update the requirements after I merge.

Issue #5931, opened 2021-03-17
Navigation Indicator Needed On Smaller Screens


dartajax:
Here is how it currently is ...

The first screen shot is of the user being on Instructor Groups.

![image](https://user-images.githubusercontent.com/7493331/111548142-791fa480-8737-11eb-9648-8d3257515af2.png)

The Instructor Group icon is shown with the mouse over and callout ...

![image](https://user-images.githubusercontent.com/7493331/111548278-b421d800-8737-11eb-98c7-b969e8504dd9.png)

That shading that occurs in this second situation should also apply to the first scenario. The "Instructor Group" icon should be highlighted. This can be shown in any of the other functional areas in Ilios but since the behavior is similar, this should suffice.

PR #5952, opened 2021-03-25
indicate active route with grey backgroun in mobile view.

stopfstedt:
image

fixes https://github.com/ilios/frontend/issues/5931


dartajax (approved):
I like it.

PR #5961, opened 2021-03-31
Test for memory leaks

jrjohnson:
Adding chrome tests is a nice addition to our CI process anyway, but
adding this memory leak detection will help us find any issues that
could cause huge problems for our users.


jrjohnson:
Someday.

PR #5962, opened 2021-03-31
consolidate sequence block components

stopfstedt:
- [x] use page objects throughout
- [x] rolls the various editor sub-components for sequence block attributes into the overview components, these are one-offs that outlived their pre-octane-era usefulness


jrjohnson (approved):
👋

dartajax (changes requested):
Button appearing where it should not be ...

image

stopfstedt:
> Button appearing where it should not be ...

good find. fixed it, please re-review. thanks!

:heavy_check_mark:

dartajax (approved):
ein besserung - nice

PR #5967, opened 2021-04-05
octanify new-sequence-block component.

stopfstedt:


dartajax (approved):
I did a bunch of click testing on CI sequence block. Looks good to me.

Issue #5972, opened 2021-04-09
Learner Group Default Location Error Msg No Longer Needed


dartajax:
Since NULL "empty string" value is allowable in this context, there is no reason to not allow a 1 or 2 char value from being saved here either. Whether anyone ever chooses to do this or not, the error msg is not helpful here anymore.

![image](https://user-images.githubusercontent.com/7493331/114212335-91d64100-9916-11eb-8a2d-7a04741d6933.png)

PR #5973, opened 2021-04-09
remove length constraint from input validation on default location.

stopfstedt:
fixes https://github.com/ilios/frontend/issues/5972

### DIFF
--- a/app/components/learnergroup-summary.js
+++ b/app/components/learnergroup-summary.js
@@ -15,7 +15,7 @@ export default class LearnergroupSummaryComponent extends Component {
   @tracked currentGroupsSaved = 0;
   @tracked learnerGroupId = null;
   @tracked learnerGroupTitle = null;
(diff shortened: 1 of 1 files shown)


dartajax (approved):
works for me - DartAjax approves

app/components/learnergroup-summary.js
```diff
@@ -15,7 +15,7 @@ export default class LearnergroupSummaryComponent extends Component {
   @tracked currentGroupsSaved = 0;
   @tracked learnerGroupId = null;
   @tracked learnerGroupTitle = null;
-  @Length(2, 100) @tracked location = null;
```
jrjohnson, line 18: I'm worried about his one, doesn't this allow users to specify a location the will not be valid [in the API](https://github.com/ilios/ilios/blob/master/src/Entity/LearnerGroup.php#L79)

PR #5987, opened 2021-04-19
octanify ilios-navigation component.

stopfstedt:


dartajax (approved):
I clicked around all over - looks good to me.

PR #5992, opened 2021-04-21
re-layouts instructorgroup header bar

### LINKED ISSUE #5990: instructorgroup header - align styles with rest of header component
make it look like the course header, for example

current look:
image

should look more like this the learner groups header bar:

image

input should span most of the width.

stopfstedt:
removes the school-prefix from the header bar, and adds a clickable breadcrumb underneath instead.

fixes #5990 

image


dartajax (approved):
I like these changes.

PR #5997, opened 2021-04-22
Refactor Learner Groups List

jrjohnson:
Octanification plus consolidation of controller into a top level
component with only query params at the controller level.


dartajax (changes requested):
This is on demo3 ...

image

This is on netlify of this PR ...

image

stopfstedt (changes requested):
found a regression - sorting the learner groups is no longer "sticky", `sortBy` does not get set in the URL anymore.

jrjohnson:
Solid reviews, two excellent catches!

> This is on netlify of this PR ...
I fixed this centering issue and it should look better now.


> found a regression - sorting the learner groups is no longer "sticky", `sortBy` does not get set in the URL anymore.
I removed this behavior because I didn't think persisting sort order to the URL had user value since, but I'm 100% willing to be talked back into it, I just couldn't come up with a reason to keep it.

stopfstedt:
> I removed this behavior because I didn't think persisting sort order to the URL had user value, but I'm 100% willing to be talked back into it, I just couldn't come up with a reason to keep it.

I'm good either way on this.

dartajax (approved):

dartajax:
The issue I reported has been resolved.

stopfstedt (approved):

PR #6019, opened 2021-04-29
Class syntax controllers

stopfstedt:
just a couple of the easy ones.


app/controllers/admin-dashboard.js
```diff
@@ -1,3 +1,3 @@
 import Controller from '@ember/controller';
 
-export default Controller.extend({});
+export default class AdminDashboardController extends Controller {}
```
jrjohnson, line 3: Is there a reason this controller even exists? It's empty.

jrjohnson (approved):

PR #6030, opened 2021-05-03
octanifies global search box

stopfstedt:
ember-template-lint doesn't like the `aria-label` on the search input field and claims "duplicate labeling". that's bogus and also clashes with the a11y test helpers assessment. 
i added this false positive as a linter todo item so we can move on from this.

### DIFF
--- a/app/components/global-search-box.js
+++ b/app/components/global-search-box.js
@@ -1,58 +1,70 @@
-import Component from '@ember/component';
-import { computed } from '@ember/object';
-import { reads } from '@ember/object/computed';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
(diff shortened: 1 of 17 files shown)


app/components/global-search-box.js
```diff
@@ -1,58 +1,70 @@
...
     }
-    return this.query;
-  }),
+    return this.args.query;
+  }
 
-  init() {
-    this._super(...arguments);
+  constructor() {
+    super(...arguments);
```
jrjohnson, line 40: This isn't needed after moving to a class, autocompleteCache = [] can just be declared as the property.

jrjohnson (approved):

app/components/global-search-box.js
stopfstedt: :heavy_check_mark:

jrjohnson (approved):

dartajax:
I clicked this one a bit too. Looks fine - gonna merge.

PR #6046, opened 2021-05-07
octanifies new-directory-user component

stopfstedt:


app/components/new-directory-user.hbs
```diff
@@ -1,292 +1,320 @@
 {{! template-lint-disable no-attrs-in-components }}
```
jrjohnson, line 1: Are these still needed? Can the need for them be removed in this refactor?

app/components/new-directory-user.js
```diff
@@ -1,127 +1,208 @@
...
+      if (currentCohort) {
+        return currentCohort;
+      }
+    }
+
+    return this.currentSchoolCohorts.lastObject;
+  }
+
+  async loadSchools() {
+    const schools = await this.store.findAll('school', { reload: true });
```
jrjohnson, line 78: I don't think the `reload: true` is necessary anymore. This seems like a workaround we may have needed in the past.

app/mixins/newuser.js
```diff
@@ -1,190 +0,0 @@
-import Mixin from '@ember/object/mixin';
```
jrjohnson, line 1: 🎉

app/components/new-directory-user.js
```diff
@@ -1,127 +1,208 @@
...
+    this.schoolId = id;
+  }
+
+  @action
+  setPrimaryCohort(id) {
+    this.primaryCohortId = id;
+  }
+
+  @action
+  async keyboard(event) {
```
jrjohnson, line 173: I'm not understanding why this method is async (and why the calls to `perform()` need to be awaited here and not just proceed after the keypress.

app/components/new-directory-user.js
```diff
@@ -132,10 +213,57 @@ export default Component.extend(NewUser, Validations, {
...
+      'phone',
+      'username',
+      'password',
+    ]);
+    const isValid = yield this.isValid();
+    if (!isValid) {
+      return false;
+    }
+    const roles = yield this.store.findAll('user-role');
+    const primaryCohort = yield this.bestSelectedCohort;
```
jrjohnson, line 240: `this.bestSelectedCohort` is a getter and not async so doesn't need to be `yield`ed.

jrjohnson (changes requested):

app/components/new-directory-user.js
stopfstedt: :heavy_check_mark:

app/components/new-directory-user.js
stopfstedt: :heavy_check_mark:

app/components/new-directory-user.js
stopfstedt: :heavy_check_mark:

app/components/new-directory-user.hbs
stopfstedt: :heavy_check_mark:

jrjohnson (approved):

PR #6047, opened 2021-05-07
fix did-update modifier usage on SchoolManager

runspired:
activating the feature flags for custom-model-classes in EmberData results in us notifying unchanged relationships any time any relationship is change on a record. This is something we will address before release but while debugging the failures in ilios I noticed that one of them was because the update codepath was broken but not tested before (the extra notifications caused it to run).


stopfstedt (approved):
LGTM. Thank you @runspired !

PR #6058, opened 2021-05-12
minor improvements to NewUser component

stopfstedt:
- convert internal tasks to actions to prevent the tasks-calling-tasks issue
- adds missing tracked property `currentSchoolCohorts`
- adds convenience getter for checking if component is loading/reloading

refs #6046


dartajax (approved):
testing on demo so form based new user - works great though

jrjohnson (approved):

PR #6069, opened 2021-05-20
octanifies user-profile-bio component.

stopfstedt:
notable changes:
- "sync with directory" button is only visible if LDAP sync is enabled
- minor layout changes due to realignment of form styles
- new after-sync style transitions


app/components/user-profile-bio.hbs
```diff
@@ -1,428 +1,450 @@
...
-      }}
-        <span class="message error">
-          {{v-get this "lastName" "message"}}
-        </span>
-      {{/if}}
+<div
+  data-test-user-profile-bio
+  class="user-profile-bio small-component {{if this.hasRecentlySaved "has-saved" "has-not-saved"}}"
+  {{did-insert (perform this.load)}}
+  {{did-update (perform this.load) @user}}
```
jrjohnson, line 5: Since the `load` action depends on `@isManaging` it should be added here as well.

jrjohnson (changes requested):

app/components/user-profile-bio.hbs
stopfstedt: :heavy_check_mark:

jrjohnson (approved):

dartajax (changes requested):
I noticed something - hopefully the screen shots that get pasted in will tell the story ...

Admin Console in the PR ...

image

Admin Console on demo3 ...

image

PR #6074, opened 2021-05-24
Remove jQuery integration

jrjohnson:
The only place we seem to still need jQuery is in some tests and it is
provided there by ember-page-object. So This change removes the flag to
provide the Ember.$ option and move us forward into the future a bit.


dartajax (approved):
I clicked through and did some testing - passed. 💯

Issue #6095, opened 2021-05-27
Navigation Context Gets Lost In Event Detail


dartajax:
It has been pointed out, accurately enough, that this will not affect students who do not perform teaching activities. Many do however. This does affect all other Ilios users.

Here's what's happening ...

After selecting to view the Calendar, the user is brought to the following. The Calendar is part of the Dashboard. Dashboard is circled (actually an ellipse of course) on the side bar menu.

![image](https://user-images.githubusercontent.com/7493331/119903158-2283d300-befd-11eb-8772-d428d00fcff3.png)

Now I am going to view the details of an event - the one on May 24th at 9:15am to review the details. After clicking on the event ...

![image](https://user-images.githubusercontent.com/7493331/119903299-6545ab00-befd-11eb-88cf-35060c72c8b5.png)

The user never made a decision to leave the Dashboard and go elsewhere. Dashboard should still be **Dashboard**.

jrjohnson:
Closed via #6096

PR #6096, opened 2021-06-09
highlight the Dashboard nav-item on routes as applicable.

### LINKED ISSUE #6094: Navigation Context Gets Lost In My Materials >> All Materials
This one is related to ilios/ilios#3465 in the sense that the user has not made a decision to leave the Dashboard so the indication on the Ilios flyout menu should not change. Dashboard should still be selected.

Here is a quick sample shot of this screen ...

before ...
image

after ...

The Dashboard is no longer **Dashboard** - it is Dashboard.

image

stopfstedt:
fixes #6094 #6095

### DIFF
--- a/app/components/ilios-navigation.hbs
+++ b/app/components/ilios-navigation.hbs
@@ -19,11 +19,22 @@
     </ul>
     <ul class="navigation-links" data-test-navigation-links>
       <li>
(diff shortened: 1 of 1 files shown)


dartajax (approved):
It works for me. Great stuff.

PR #6097, opened 2021-06-09
fix template linter warnings

stopfstedt:
another round of this.


jrjohnson (approved):
Closer we creep to perfection
Every step harder than the last
Fittingly the reward is to remove pain
not to add pleasure
at the end.

PR #6120, opened 2021-06-22
fix: fetch async data before testing length

runspired:
There are other ways to test length as well using references, but testing the length of an unloaded async relationship is not reliable.

That said, this appears to surface an unexpected divergence in behavior between the 3.27 release and 3.26 beta+ of ember-data, will open an issue to look into it.


runspired:
cc @stopfstedt think you could review this? I think @jrjohnson is away.

stopfstedt:
> cc @stopfstedt think you could review this? I think @jrjohnson is away.

Thanks @runspired , I'll have a look this afternoon.

stopfstedt (approved):
LGTM

PR #6124, opened 2021-06-29
Automerge safe dependabot PRs

jrjohnson:
These are dependencies we control or else they are very well covered by
tests. Thus they can be automerged.


stopfstedt (approved):
LGTM

PR #6132, opened 2021-07-02
corrects query param names on root breadcrumb item.

### LINKED ISSUE #6131: Breadcrumb Issue in Learner Groups - Maintain Proper School Root
So I'm here - researching this learner group's school relationship.

image

So you may wonder -- why not follow the breadcrumbs up an bit and when you land at the root ("Learner Groups"), there the mystery will be revealed. When I click up the tree at the root, here's where you land currently on demo3 ...

image

This is not super helpful because ...

A. I believe I was looking at a School of Pharmacy group anyway.
B. Even if it was a School of Medicine learner group, it clearly wasn't in the Class of 2025 - Bridges since there are no learner groups in SOM Bridges Class of 2025. Even if there were, -Neuroscience and Therapeutics - Skills/Inquiry is not one of them.

stopfstedt:
fixes #6131

### DIFF
--- a/app/components/learnergroup-header.hbs
+++ b/app/components/learnergroup-header.hbs
@@ -54,11 +54,11 @@
         <LinkTo
           @route="learnerGroups"
           @query={{hash
-            schoolId=(get
+            school=(get
               (await @learnerGroup.cohort.programYear.program.school) "id"
             )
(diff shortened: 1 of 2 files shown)


jrjohnson (approved):

dartajax (approved):
Looks good to me.

PR #6159, opened 2021-07-31
fix: remove match in removeMatch()

Trott:
Refs: https://lgtm.com/projects/g/ilios/frontend/snapshot/6c1b50ecb13f0674eedd792e45312daded58e53b/files/app/components/learnergroup-bulk-assignment.js?sort=name&dir=ASC&mode=list#x198251bf390a9134:1

I'll note that lgtm.io does give this repo an A+ for code quality. 🎉


jrjohnson:
Looks good. This appears to be working because it is attached to a select dropodown currently so removing a group always wants to remove all groups. I'm going to add a test for switching groups and then this will be good to merge.

PR #6164, opened 2021-08-04
Fail tests builds when translations are missing

jrjohnson:
This is the best defense against this as recommended by the addon
author. Regular builds will show a warning, but test runs will not build
and die with an error message.


stopfstedt (approved):
:+1:

PR #6170, opened 2021-08-09
Manually Register Listener for Connection Status

jrjohnson:
Ember Lifeline's implementation of addEventListener is deprecated in
favor of a modifer based approach.


stopfstedt (approved):
elegant solution. LGTM.

PR #6174, opened 2021-08-11
Stop using PostCSS

jrjohnson:
This reverts our move to PostCSS and returns to ember-cli-sass to
compile our styles. PostCSS is great, but the sass plugin is abandoned
and has very outdated dependencies as well as not being compatible with
the latest PostCSS version. Until we're willing to re-write entirely in
non-sass this isn't a good path for us.


jrjohnson:
@dartajax this needs an in-depth click test to ensure there are no style issues anywhere. I've assigned it to you for that initial review because the code itself hasn't changed at all.

dartajax:
Got it. Gonna test now.

jrjohnson:
Thanks @dartajax I'm figuring out the build issue here and then I'll move this forward.

dartajax:
Been testing on Safari - seems ok.

stopfstedt (approved):

PR #6183, opened 2021-08-19
make ci report description mandatory.

### LINKED ISSUE #6182: Put Default Value in Curriculum Inventory Report Description Field
This causes problems with the .XML validation of CI Reports. When there is no value entered into the description field in a CI Report, the other entered values do not get validated at all. The error message received is confusing as well. In the future, 2022 or later, this field value will be optional anyway and won't cause this issue. To circumvent this, let's set a value there "Curriculum Inventory Report" will suffice.

Pre-save ...

image

Post-save ...

image

We can just pre-populate this so it can be edited later by the user.

stopfstedt:
fixes #6182

### DIFF
--- a/app/components/curriculum-inventory/new-report.hbs
+++ b/app/components/curriculum-inventory/new-report.hbs
@@ -37,11 +37,15 @@
           </label>
           <textarea
             id="description-{{templateId}}"
-            value={{this.description}}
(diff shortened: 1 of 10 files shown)


dartajax (approved):
I like it. I am stanning this change.

PR #6187, opened 2021-08-23
span entire width of screen when editing sequence block title.

### LINKED ISSUE #6186: CI Sequence Block Title Fields Should Span the Page
Visible from the very helpful breadcrumbs, it can be seen that we are now at the Curriculum Block level (top of it). In this case, the block needs to be edited. This process is started as shown below.

image

After clicking ...

image

stopfstedt:
fixes #6186


dartajax (approved):
Very nice - Dart_500 approved.

PR #6193, opened 2021-08-24
correctly sets selected school id on manage link.

### LINKED ISSUE #6192: Admin Console School Drop-Down Not Taking User to Selected School
start here - this is in production

image

change the school to Pharmacy

image

click "Manage"

now you are here ...

back on "Medicine" but it should be "Pharmacy"

image

stopfstedt:
fixes #6192


dartajax (approved):
It seems fine to me. It kept the correct value when going to add to cohort but since there are no accts needing maintenance on demo currently, I couldn't test that completely.  I approve in any case.

PR #6198, opened 2021-08-25
adds competency to PCRS mapper.

stopfstedt:
fixes #6197


jrjohnson (approved):
Code looks great, love the extraction of the ID from the URL as a util exposed through a helper. Very cool.

PR #6206, opened 2021-09-01
Update Input component arguments

jrjohnson:
Switching out keypress for keyup as well since the former is deprecated.


stopfstedt (approved):
LGTM

Issue #5719, opened 2020-11-24
Admin Console Year Selector Should Retain Its Value Upon Return via Back Button


dartajax:
![image](https://user-images.githubusercontent.com/7493331/100142628-0b2e3500-2e49-11eb-9b16-16b0e8d77f88.png)

![image](https://user-images.githubusercontent.com/7493331/100142977-7841ca80-2e49-11eb-87cc-01cd2eec647d.png)

![image](https://user-images.githubusercontent.com/7493331/100143153-bdfe9300-2e49-11eb-921b-471125234d2a.png)

stopfstedt:
i can't get this addition to work with this component in its current form, it's the George Costanza Wallet.

will have another go at it tomorrow, perhaps breaking this thing up into smaller components will yield the desired result.

stopfstedt:
btw, if we're doing sticky year selections, then we should also make the selected school sticky.

PR #6217, opened 2021-09-10
adds stickiness to school and year selections from permissions dropdowns

stopfstedt:
fixes https://github.com/ilios/frontend/issues/5719

this affect the permissions display in the user admin console as well as "my profile".

### DIFF
--- a/app/components/user-profile-permissions.js
+++ b/app/components/user-profile-permissions.js
@@ -56,7 +56,7 @@ export default class UserProfilePermissionsComponent extends Component {
     );
 
     this.schools = (yield this.store.findAll('school', { reload: true })).toArray();
-    this.academicYears = (yield this.store.findAll('academic-year', { reload: true })).toArray();
+    this.academicYears = (yield this.store.findAll('academic-year')).toArray();
     this.defaultSchool = yield this.args.user.school;
 
     this.selectedYearId = this.getBestSelectedYearId(
--- a/app/components/user-profile-permissions.js
+++ b/app/components/user-profile-permissions.js
@@ -55,7 +55,7 @@ export default class UserProfilePermissionsComponent extends Component {
       'academicYearCrossesCalendarYearBoundaries'
     );
 
-    this.schools = (yield this.store.findAll('school', { reload: true })).toArray();
+    this.schools = (yield this.store.findAll('school')).toArray();
     this.academicYears = (yield this.store.findAll('academic-year')).toArray();
     this.defaultSchool = yield this.args.user.school;
 
--- a/app/components/user-profile-permissions.js
+++ b/app/components/user-profile-permissions.js
@@ -46,7 +46,12 @@ export default class UserProfilePermissionsComponent extends Component {
     this.schools = map.schools;
     this.academicYears = map.academicYears;
 
-    let currentYear = Number(moment().format('YYYY'));
+    let currentYear;
+    if (this.args.selectedYear) {
+      currentYear = Number(this.args.selectedYear);
+    } else {
+      currentYear = Number(moment().format('YYYY'));
+    }
     const currentMonth = Number(moment().format('M'));
     if (this.academicYearCrossesCalendarYearBoundaries && currentMonth < 6) {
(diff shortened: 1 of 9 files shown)


app/components/user-profile-permissions.js
```diff
@@ -31,99 +32,84 @@ export default class UserProfilePermissionsComponent extends Component {
...
 
-    const map = yield hash({
-      schools: this.store.findAll('school'),
-      academicYears: this.store.findAll('academic-year'),
-      defaultSchool: this.args.user.school,
-    });
-
-    this.schools = map.schools;
-    this.academicYears = map.academicYears;
+    this.schools = (yield this.store.findAll('school', { reload: true })).toArray();
```
jrjohnson, line 58: Why the reload?

app/components/user-profile-permissions.js
```diff
@@ -31,99 +32,84 @@ export default class UserProfilePermissionsComponent extends Component {
...
+    this.academicYears = (yield this.store.findAll('academic-year', { reload: true })).toArray();
```
jrjohnson, line 59: Why the reload?

jrjohnson (changes requested):
Along with the reload questions I think this would benefit from moving to the `@use` syntax instead of the `did-insert` loader. That is the way ember is moving and as we re-write things they should follow that pattern if possible.

dartajax (approved):
This is an improvement to be certain.

dartajax:
Is there any way to add stickiness to the expanded state (or not) of Programs, Program Years, Courses, and Sessions so that when the user comes back to the Admin Console, those values would be in the same state as before the user followed the link?

app/components/user-profile-permissions.js
stopfstedt: good question - beats me - it yields an empty collection without.

app/components/user-profile-permissions.js
stopfstedt: > Is there any way to add stickiness to the expanded state (or not) of Programs, Program Years, Courses, and Sessions so that when the user comes back to the Admin Console, those values would be in the same state as before the user followed the link?

technically, that's all doable. the question is - how far do you want to spin that thread... i can't answer that. if this is important, then pls file a follow up ticket and i'll get to it asap.

stopfstedt:
> Is there any way to add stickiness to the expanded state (or not) of Programs, Program Years, Courses, and Sessions so that when the user comes back to the Admin Console, those values would be in the same state as before the user followed the link?

technically, that's all doable. i'd prefer a follow up ticket on that.

app/components/user-profile-permissions.js
jrjohnson: We preload schools in the application controller to ensure all subsequent loads are fast. Is it empty only in tests? Or when used in the app? I'll investigate.

dartajax:
If @jrjohnson reviews and approves this and his change request is approved as well, it would be my honor to merge this

app/components/user-profile-permissions.js
stopfstedt: you're right, this is not necessary here. i went overboard when dealing with the incomplete schools loading behaviour. i now removed that flag :heavy_check_mark:

app/components/user-profile-permissions.js
stopfstedt: check out `app/adapters/school.js`, i'm fairly certain that the defined loading behaviour there is causing the problem here. forcing a reload here, as an exception to the rule, seems acceptable to me.

jrjohnson: Ok, looks like I broke this earlier this year and I'm surprised it hasn't bitten us before this. Several places in the app only work now because users usually navigate directly from the dashboard or else never change from their default school. I added a fix in https://github.com/ilios/frontend/pull/6235 and once that is merged this `reload` can be removed.

stopfstedt: :heavy_check_mark:

jrjohnson (approved):

PR #6230, opened 2021-09-16
Expand input elements in learner group overview

### LINKED ISSUE #5867: More Space Needed to Edit Learner Group Defaults
It should span across the screen I believe - or something similar.

image

When you click in to edit either of those defaults, there is not enough space to edit the long string or URL contained in these fields. 

image

stopfstedt:
fixes #5867


dartajax (approved):
LGTM

PR #6231, opened 2021-09-16
replaces onclick event binding with modifier.

stopfstedt:


stopfstedt:
rolled-into and superseded by #6234

PR #6234, opened 2021-09-17
fix template linter errors

stopfstedt:


app/components/user-menu.hbs
```diff
@@ -24,20 +24,20 @@
     <ClickOutside @onClickOutside={{fn (set this.isOpen) false}}>
       <ul class="menu">
         <li tabindex="-1" data-test-item>
-          <LinkTo @route="myprofile" @invokeAction={{set this.isOpen false}}>
+          <LinkTo @route="myprofile" {{on "click" (set this.isOpen false)}}>
```
jrjohnson, line 27: Unfortunately this doesn't work see https://stackoverflow.com/questions/69021116/add-extra-actions-to-linkto-in-octane I have an idea for a component to solve this problem, but this should be reverted here.

app/components/user-menu.hbs
```diff
@@ -24,20 +24,20 @@
...
             {{t "general.myProfile"}}
           </LinkTo>
         </li>
         <li tabindex="-1" data-test-item>
           <LinkTo
             @route="mymaterials"
-            @invokeAction={{set this.isOpen false}}
+            {{on "click" (set this.isOpen false)}}
```
jrjohnson, line 34: see above, revert.

app/components/user-menu.hbs
```diff
@@ -24,20 +24,20 @@
...
           >
             {{t "general.myMaterials"}}
           </LinkTo>
         </li>
         <li tabindex="-1" data-test-item>
-          <LinkTo @route="logout" @invokeAction={{set this.isOpen false}}>
+          <LinkTo @route="logout" {{on "click" (set this.isOpen false)}}>
```
jrjohnson, line 40: see above, revert.

jrjohnson (changes requested):

app/components/user-menu.hbs
stopfstedt: :heavy_check_mark:

app/components/user-menu.hbs
stopfstedt: :heavy_check_mark:

app/components/user-menu.hbs
stopfstedt: :heavy_check_mark:

jrjohnson (approved):

PR #6237, opened 2021-09-20
Handle actions on LinkTo with a component

jrjohnson:
This builds an internal LinkToWithAction component that does what we
need in this case without overriding the internal component. It isn't a
great replacement, but it handles this case and removes the deprecation
and linting errors that our original solution has caused.


app/components/user-menu.hbs
```diff
@@ -24,22 +24,23 @@
...
-          </LinkTo>
+          </LinkToWithAction>
         </li>
         <li tabindex="-1" data-test-item>
-          <LinkTo
+          <LinkToWithAction
             @route="mymaterials"
-            @invokeAction={{set this.isOpen false}}
+            @action={{set this.isOpen false}}
+            @queryParams={{hash course=null filter=null sortBy=null firstOfferingDate=null}}
```
stopfstedt, line 35: this blows up when you set `sortBy` to `null` with a `TypeError: sortKey is null` error.

```suggestion
            @queryParams={{hash course=null filter=null firstOfferingDate=null}}
```

stopfstedt (changes requested):

app/components/user-menu.hbs
jrjohnson: This doesn't end up setting the default correctly as it populates it on the URL string. I'm gonna try changing this controller a bit in common to handle this. Good catch!

jrjohnson:
Requires https://github.com/ilios/common/pull/2350

stopfstedt (approved):
retested with ilios/common#2350, looks good to me.

stopfstedt:
flagged as `DO NOT MERGE` until the next common release has been cut and integrated.

PR #6312, opened 2021-11-02
updates ember-plugin-qunit dependency.

stopfstedt:


stopfstedt:
self approved and merged so i can move today.

PR #6320, opened 2021-11-04
Update common 61.2.0

jrjohnson:
And update our type and style to reflect new common standards.

### DIFF
--- a/app/styles/components/ilios-navigation.scss
+++ b/app/styles/components/ilios-navigation.scss
@@ -14,6 +14,7 @@
     padding: 0;
 
     li {
+      /* stylelint-disable property-disallowed-list */
       font-size: 4vmin;
       height: 100%;
       padding: .25rem;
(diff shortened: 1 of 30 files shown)


app/styles/components/ilios-navigation.scss
```diff
@@ -109,9 +112,10 @@
     height: 100%;
     text-decoration: none;
     width: 100%;
+    font-weight: 400;
 
     &.active {
-      font-weight: bold;
+      font-weight: 700;
```
stopfstedt, line 118: i'm curious about this change. why go to numbers here?

jrjohnson: Habit. We pre load 400,600, and 700 weights of our font so sticking to those numbers avoids needing to load new font files. I can never remember what weight each keyword represents.

stopfstedt (approved):
code reviewed and visually spot-checked. lgtm.

PR #6328, opened 2021-11-07
Octaneify remaining controllers

jrjohnson:
I got a bit distracted with data loading in the middle of this, but pulling it into two PRs was beyond me so it's slightly bigger change here, but hopefully manageable.


stopfstedt (approved):

PR #6339, opened 2021-11-14
Stop relying on the default model hook

jrjohnson:
This is deprecated and won't work very well in Ember v4 do to the
absence of a default store. Let's get ahead and remove these and
replace with explicit model loading.


stopfstedt (approved):
LGTM

PR #6344, opened 2021-11-15
Refactor DashboardMyReports component

### LINKED ISSUE #6343: MyReports Report deletion fails
On Stage:

 when trying to delete a MyReports report, this error is generated in the inspector:

`Error: Promised response from onMessage listener went out of scope`

and the report is not deleted. a second attempt will generate"

`Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://sentry.io/api/1311608/envelope/?sentry_key=ded7a44cf4084601a2fb468484bbe3ed&sentry_version=7. (Reason: CORS request did not succeed).` and 
`Uncaught Error: Attempted to handle event `deleteRecord` on <report:1837> while in state root.deleted.saved. ` and will print the "Browser not supported" hijinks across the screen.

Hard refresh clears the screen and completes the deletion.

jrjohnson:
Using resources instead of a loader ensures that new reports and delete
reports are reflected in the UI.

Fixes #6343

### DIFF
--- a/app/components/dashboard-myreports.hbs
+++ b/app/components/dashboard-myreports.hbs
@@ -1,152 +1,149 @@
 {{#let (unique-id) as |templateId|}}
   <div
     class="dashboard-myreports"
-    {{did-insert (perform this.load) (await this.user.reports) @selectedReport @selectedYear}}
-    {{did-update (perform this.load) (await this.user.reports) @selectedReport @selectedYear}}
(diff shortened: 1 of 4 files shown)


stopfstedt (approved):

dartajax (approved):
looks fine functionally - those font sizes I'm not totally sure of yet - but that's not this issue

PR #6356, opened 2021-11-23
Update ember-auto-import to v2

jrjohnson:
This requires setting up webpack ourselves and indicating where our
scripts go since we use defer there


stopfstedt (approved):
LGTM

stopfstedt (approved):
reviewed and click tested. LGTM.

PR #6359, opened 2021-11-24
Disable Token Refresh

jrjohnson:
Not sure what the long term value of this feature is, but it is broken
now so let's disable it.


stopfstedt (approved):
agreed.

PR #6368, opened 2021-12-02
Improve network loading for user profile data

jrjohnson:
Previously this required a *lot* of requests to load all the ways a user
is connected to courses and sessions. Putting this all in a service
reduces the requests and ensures everything we need it already loaded.


stopfstedt (approved):
i :heart:  it.

PR #6387, opened 2021-12-16
Remove simple-set-helper

jrjohnson:
We have this transform in common now. Shouldn't need this addon anymore.


dartajax (approved):
looks great

PR #6391, opened 2021-12-20
removes "Edit" button from list components

stopfstedt:
wherever this exist, it's a redundant design element - the title of the list item links to the same screen and serves the same purpose. 

it's also inconsistently implemented - present in some lists, but not in others.

this removes them across the board.

### DIFF
--- a/app/components/school-list.hbs
+++ b/app/components/school-list.hbs
@@ -90,25 +90,17 @@
         <table>
           <thead>
             <tr>
-              <th class="text-left" colspan="4">
+              <th class="text-left">
                 {{t "general.school"}}
-              </th>
-              <th class="text-left" colspan="1">
-                {{t "general.actions"}}
               </th>
             </tr>
           </thead>
           <tbody>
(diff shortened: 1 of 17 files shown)


jrjohnson:
Assigning to both @saschaben and @dartajax for review as the code is good, but I don't know if these buttons are a necessary design element for usability or not.

dartajax (approved):
From my perspective this is fine.

app/components/school-list.hbs
```diff
@@ -90,27 +90,19 @@
         <table>
           <thead>
             <tr>
-              <th class="text-left" colspan="4">
```
jrjohnson, line 93: Why the `colspan` change?

stopfstedt: no need to span anymore since it's just one column now.

jrjohnson (approved):

app/components/school-list.hbs
jrjohnson: Lol. Yeah. That'll do it!

PR #6402, opened 2021-12-21
adds click-to-select to user attributes in learner group management 

### LINKED ISSUE #6401: Make Entire Row Clickable - Another Learner Group Management Suggestion
It is necessary to click directly into the relatively small check box in order to select a user to be added to a learner group. It would be easier to use if you could click anywhere in the row and not have to focus only on that very small check box. I'm talking as a user here.

image

stopfstedt:
fixes #6401

the whole cannot be made clickable, so i made the various user attributes in that row clickable instead.


jrjohnson (changes requested):
If you style the  inline-button with `width: 100%` and `text-alight: left` it will fill the entire column so the whitespace between text will be clickable as well.

stopfstedt:
> `width: 100%` and `text-alight: left`

nice! done.

jrjohnson (approved):

Issue #6417, opened 2022-01-05
Course Filter Should Not Lose Focus After Search Character is Entered


dartajax:
I noticed, while of course trying to do something else, that the search box for courses loses focus after key press / key enter of a search criteria, well any singular character actually.

I am looking to find the course that has a title that starts with "Renal". So I wanted to enter the first three characters of this "Ren" and then review what was returned.

What happened ... I started typing "r" and then by the time I tried typing "e", the second character, nothing happened. I had to manually with the mouse go back and click into the search box again where I was then allowed to type the second character before the same thing happens again. The search box here is not working as it should. The user needs to be able to adjust what they type and judge that against the results received with the focus remaining on the search box.

<img width="1213" alt="Screen Shot 2022-01-05 at 1 53 01 PM" src="https://user-images.githubusercontent.com/7493331/148295287-d4541ddb-d3dc-49ee-a23b-cbbad9922561.png">

stopfstedt:
losing focus on search boxes each time the application state changes is not a feature imo, but a bug. despite what this plugin claims. 

suggesting we lose this a11y plugin. 

@jrjohnson thoughts?

dartajax:
I am not surprised it happens in the other locations as well. I'd say bug not feature. My $.02.

jrjohnson:
I'd really hate to get rid of this. It's apparently an absolutely critical thing for screen readers. I'm guessing the issue is that we're changing the query param on these filters. I'd actually be happier to dump saving filter state to the URL than removing this addon. I wonder if they'd be willing to see this as a bug.

stopfstedt:
quoting from the plugin's [readme file](https://github.com/ember-a11y/ember-a11y-refocus).

> 2. It moves the focus to that message for the screen reader user, effectively resetting focus in Ember apps (similar to how a native web page/site works).

i'm not 100% sure on how to interpret that. should the focus only be reset for screen readers _only_, or in general?

jrjohnson:
The focus should only move when the route transitions (a link is followed). Changing it on the query param seems like a bug to me. I opened this as an issue upstream. Let's wait to hear back and then figure out what we should do next.

PR #6419, opened 2022-01-06
pins ember-a11y-refocus dependency to version 2.0.0

stopfstedt:
the next version of this plugin, v2.1.0, introduces some
behavior that we don't want in ilios.
it is also not clear to us at this point whether those changes
consistute a bug or feature.
therefore reverting to a known, "good" state.

fixes https://github.com/ilios/frontend/issues/6417

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10841,9 +10841,9 @@
       }
     },
     "ember-a11y-refocus": {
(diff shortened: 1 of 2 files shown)


dartajax (approved):
This is more like it - back to normal.

PR #6441, opened 2022-01-22
turn off autocompletion on search inputs.

stopfstedt:
refs ilios/ilios#3849

### DIFF
--- a/app/components/global-search-box.hbs
+++ b/app/components/global-search-box.hbs
@@ -1,6 +1,7 @@
 <div class="global-search-box" data-test-global-search-box>
   <input
     aria-label={{t "general.searchTheCurriculum"}}
+    autocomplete="off"
     class="global-search-input"
     data-test-input
     type="search"
--- a/app/components/global-search-box.hbs
+++ b/app/components/global-search-box.hbs
@@ -1,7 +1,7 @@
 <div class="global-search-box" data-test-global-search-box>
   <input
     aria-label={{t "general.searchTheCurriculum"}}
-    autocomplete="off"
+    autocomplete="name"
     class="global-search-input"
     data-test-input
     type="search"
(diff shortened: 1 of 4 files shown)


app/components/global-search-box.hbs
```diff
@@ -1,6 +1,7 @@
 <div class="global-search-box" data-test-global-search-box>
   <input
     aria-label={{t "general.searchTheCurriculum"}}
+    autocomplete="off"
```
jrjohnson, line 4: I think this should be [name](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)

```suggestion
    autocomplete="name"
```

app/components/ilios-users.hbs
```diff
@@ -8,6 +8,7 @@
   <div class="filters" data-test-filters>
     <div class="filter user-search">
       <input
+        autocomplete="off"
```
jrjohnson, line 11:
```suggestion
        autocomplete="name"
```

app/components/manage-users-summary.hbs
```diff
@@ -34,6 +34,7 @@
   </div>
   <div class="user-search">
     <input
+      autocomplete="off"
```
jrjohnson, line 37:
```suggestion
      autocomplete="name"
```

app/components/new-directory-user.hbs
```diff
@@ -204,6 +204,7 @@
       >
         <input
           aria-label={{t "general.search"}}
+          autocomplete="off"
```
jrjohnson, line 207:
```suggestion
          autocomplete="name"
```

jrjohnson (changes requested):
This change seems correct, but I noticed it didn't work. I think that's because Safari doesn't really support it, it seems to work well in other browsers depending on a bunch of variables. However while I was looking it up I found that it could be set to `name` to indicate that the input expected some part of a name and that seems like the right value for most of these.

jrjohnson (approved):

PR #6457, opened 2022-01-26
Bump Node to v16 and pin @embroider/macros

jrjohnson:
This will unblock us from making other changes to the lock file as
upstream work on getting macros updated everywhere continues.


stopfstedt (approved):
click-tested and reviewed. LGTM.

PR #6497, opened 2022-02-11
Handle Changes to the way ember page object selects things in v2

jrjohnson:
In the process I also found and fixed a rendering issue with our user menu.


stopfstedt (approved):
reviewed and user menu click-tested. LGTM.

PR #6506, opened 2022-02-16
Add a deploy:test command

jrjohnson:
This builds everything, but doesn't push it to S3.

### DIFF
--- a/config/deploy.js
+++ b/config/deploy.js
@@ -90,6 +90,7 @@ module.exports = function (deployTarget) {
     ENV['cloudfront'].distribution = 'E1RJJYSB507IYA';
   }
   if (deployTarget === 'development') {
+    ENV.build.environment = 'production';
     ENV.pipeline.disabled = {
       's3-index': true,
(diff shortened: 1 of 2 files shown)


config/deploy.js
```diff
@@ -90,6 +90,7 @@ module.exports = function (deployTarget) {
     ENV['cloudfront'].distribution = 'E1RJJYSB507IYA';
   }
   if (deployTarget === 'development') {
+    ENV.build.environment = 'production';
```
stopfstedt, line 93: i'm confused by this line. why change the build env to production when the deploy target is development?

jrjohnson: Yeah, that's confusing I agree. The deployment environment is development, however the ember build will be run against production. This ensures that the output is the same as it would be for a production app (minimized, comments stripped out, no tests files) which makes it suitable for testing various things that happen in our build deploys on a semi-annual basis when built for production.

stopfstedt (approved):

PR #6507, opened 2022-02-16
Remove content security policy addon and config

jrjohnson:
This is causing issues with our build and isn't configured up to the
latest standards of the addon.


jrjohnson:
This is some part of the solution to our build issues and I'm going to go ahead and merge it. Will get more info with this in the mix.

PR #6532, opened 2022-02-25
When installing ignore scripts in CI/CD

jrjohnson:


jrjohnson:
Not an option here because we need `sharp` to process images.

PR #6549, opened 2022-03-10
Use ValidationError component

jrjohnson:
No longer relies on await in the template and much easier to read, this
component from common is the new way cool kids display validation
errors.


stopfstedt (approved):
that is a very nice improvement :+1:

PR #6550, opened 2022-03-10
Handle errors in an embroiderer compatible way

jrjohnson:
Embroiderer hates treeFor* methods so I've placed the error script
inline into the html instead. I'd like to put it in a separate JS file,
but we need access to come build info to create the browser list.


stopfstedt (approved):
LGTM

PR #6582, opened 2022-03-24
Replace test helpers

jrjohnson:
Most of these have been moved to common and can be imported from there. Others were part of old ember blueprints that don't exist anymore.


stopfstedt (approved):
Neat!

PR #6605, opened 2022-04-06
Use getPrimaryLocale to get locale

jrjohnson:
the locale property returns an array or possible locales which doesn't
match what we're expecting in these places. PrimaryLocale is the right
property for doing what we want.


stopfstedt (approved):
LGTM

PR #6610, opened 2022-04-07
replace task with async action.

### LINKED ISSUE #6609: Deal with self-cancel loops in editable field
As noted in https://github.com/ilios/frontend/pull/5507#discussion_r469572537 we're getting this console warning and we should address it.

> ember-concurrency detected a potentially hazardous "self-cancel loop" between parent task saveData and child task saveCourse. If you want child task saveCourse to be canceled when parent task saveData is canceled, please change .perform() to .linked().perform(). If you want child task saveCourse to keep running after parent task saveData is canceled, change it to .unlinked().perform()

stopfstedt:
currently, this is the only way to shake the self-cancel warning that's being emitted by ember-concurrency.
editable-field is not aware what kind of save callback is being passed - it could be a task or an action - so it just "yields".
using an action instead of a task appears to be an adequate solution for this problem to me.

fixes #6609

### DIFF
--- a/app/components/curriculum-inventory/sequence-block-overview.js
+++ b/app/components/curriculum-inventory/sequence-block-overview.js
@@ -222,13 +222,15 @@ export default class CurriculumInventorySequenceBlockOverviewComponent extends C
 
   @action
   async saveCourse() {
-    const oldCourse = await this.args.sequenceBlock.course;
-    if (oldCourse !== this.course) {
-      this.args.sequenceBlock.set('sessions', []);
-      this.args.sequenceBlock.set('excludedSessions', []);
+    if (!this.isDestroying) {
+      const oldCourse = await this.args.sequenceBlock.course;
+      if (oldCourse !== this.course) {
+        this.args.sequenceBlock.set('sessions', []);
+        this.args.sequenceBlock.set('excludedSessions', []);
+      }
+      this.args.sequenceBlock.set('course', this.course);
+      await this.args.sequenceBlock.save();
     }
-    this.args.sequenceBlock.set('course', this.course);
-    await this.args.sequenceBlock.save();
   }
 
   @action
--- a/app/components/curriculum-inventory/sequence-block-overview.js
+++ b/app/components/curriculum-inventory/sequence-block-overview.js
@@ -220,15 +220,15 @@ export default class CurriculumInventorySequenceBlockOverviewComponent extends C
     this.minimum = this.args.sequenceBlock.minimum;
   }
 
-  @dropTask
-  *saveCourse() {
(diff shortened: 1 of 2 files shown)


app/components/curriculum-inventory/sequence-block-overview.js
```diff
@@ -220,15 +220,15 @@ export default class CurriculumInventorySequenceBlockOverviewComponent extends C
...
-  *saveCourse() {
-    const oldCourse = yield this.args.sequenceBlock.course;
+  @action
+  async saveCourse() {
+    const oldCourse = await this.args.sequenceBlock.course;
     if (oldCourse !== this.course) {
       this.args.sequenceBlock.set('sessions', []);
       this.args.sequenceBlock.set('excludedSessions', []);
     }
     this.args.sequenceBlock.set('course', this.course);
```
jrjohnson, line 230: This is where we get into trouble with an action, as this `set` is called after the `await` it is possible for this component to get torn down and `this` to be invalid during the async work. To be safe this approach should protect against that with `if (!this.isDestroying) {`.

jrjohnson (changes requested):

app/components/curriculum-inventory/sequence-block-overview.js
stopfstedt: :heavy_check_mark:

app/components/curriculum-inventory/sequence-block-overview.js
```diff
@@ -220,15 +220,17 @@ export default class CurriculumInventorySequenceBlockOverviewComponent extends C
...
+  async saveCourse() {
+    if (!this.isDestroying) {
+      const oldCourse = await this.args.sequenceBlock.course;
+      if (oldCourse !== this.course) {
+        this.args.sequenceBlock.set('sessions', []);
+        this.args.sequenceBlock.set('excludedSessions', []);
+      }
+      this.args.sequenceBlock.set('course', this.course);
+      await this.args.sequenceBlock.save();
     }
```
jrjohnson, line 233:
```suggestion
    const oldCourse = await this.args.sequenceBlock.course;
    if (!this.isDestroying) {
      if (oldCourse !== this.course) {
        this.args.sequenceBlock.set('sessions', []);
        this.args.sequenceBlock.set('excludedSessions', []);
      }
      this.args.sequenceBlock.set('course', this.course);
      await this.args.sequenceBlock.save();
    }
```
Have to do the async work first so when it comes back the isDestroying will be checked.

jrjohnson (changes requested):

jrjohnson (approved):

PR #6611, opened 2022-04-07
Add the app version to the footer

jrjohnson:
Now that we can get the version of the backend from the API we can
display it in the footer. Fixes ilios/ilios#2043

Ilios footer post change


stopfstedt (approved):
:fire:

PR #6618, opened 2022-04-08
Fix Responsive Layout

jrjohnson:
The search box breaks out mobile layout, it has to be moved into the
next row on phone sized screens.


saschaben:
WFM

PR #6626, opened 2022-04-14
rm lang strings that we get from common already.

stopfstedt:
refs https://github.com/ilios/common/pull/2804
refs https://github.com/ilios/common/pull/2806


dartajax (approved):
looks good to me

PR #6627, opened 2022-04-15
adds deletion confirmation for school session type

### LINKED ISSUE #3231: Prompt and Confirmation Needed When Session Type Gets Deleted
If a Session Type can be deleted (ie. hasn't been used), and it gets deleted / removed in Ilios, there is no message asking the user if he or she is certain this is what is wanted AND there is no confirmation once the action has been completed.

See below.

image

And then ... (crickets)

This list appears to flutter but the user should be prompted and then should receive a confirmation with even a possibility of undoing that action just in case.

image

stopfstedt:
fixes #3231


dartajax (approved):
looks good to me

PR #6632, opened 2022-04-18
Remove capitals from route names

jrjohnson:
We've been ignoring this linting error for a while, but it causes big
problems with embroider so I've gone ahead and replaced all the capital
route names with dasherized.


stopfstedt (approved):
reviewed and routes click-tested. lgtm.

PR #6637, opened 2022-04-21
Fix annoying horizontal scroll

jrjohnson:
This appears to be a known browser bug, I'm not sure why it is popping
up for us now, but using vw units doesn't account for any scroll bars so
anytime there is a need for vertical scroll we were getting horizontal
bars as well. By using a 100% of the body we get the same effect without
the bug.

Fixes ilios/ilios#3999

### DIFF
--- a/app/styles/layout/_layout.scss
+++ b/app/styles/layout/_layout.scss
@@ -7,7 +7,7 @@
   'footer';
   grid-template-rows: auto 10px 1fr 20px;
   height: 100vh;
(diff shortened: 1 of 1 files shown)


dartajax (approved):
Click tested on Windows 10 - Chrome - great job - no horizontal scroll anymore - I like it and approve this PR.

stopfstedt (approved):
LGTM. visually reviewed with latest FF/Chrome on Linux.

PR #6638, opened 2022-04-21
Restructure header for a11y

jrjohnson:
This may only be an issue for our automated scanner, but it's having
trouble parsing out the route announcer because it is technically inside
the orange header. This restructuring moves the visual content into a
div with the background without changing the actual look of the page.


stopfstedt (approved):
reviewed code and visually. LGTM.

PR #6679, opened 2022-05-26
Re-thinking Navigation

jrjohnson:
Creating a bolder navigation and left hand drawer feels like a more
modern style and also gives less visual white space to distract from the
content.

wip:
- [x] Initial Thoughts and Review / Opinions
- [x] Need to constantly style `Back to X` links and figure out where they belong
- [x] Needs https://github.com/ilios/common/pull/2872


dartajax:
At first I found the blue background for the selected Dashboard to be a bit too much. After navigating though, it makes it very clear where you are. I'm not totally sold on that blue as the background color. Also, with the black background below the flyout menu, we will be provided the opportunity/responsibility to really line everything up better - left justify. The dark background makes these indentation issues more apparent.

jrjohnson:
Agree on all points. Could you add a list of everything that has indentation issues in this PR? I think it's mostly stuff in common, but having a definitive list will be super useful.

dartajax:
pretty neat - is there any way to slightly change the font color itself to be the same blue as the icon and arrow (which I like by the way)?

dartajax:
image

dartajax:
on demo ... all three of these "Session Attributes" fit - keep in mind there may be more than 3 of these - schools can customize them - as SOP has.

old

on PR ... Same screen size - the image has a diff width because I selected the range by hand. Only two fit here. 

new

dartajax (changes requested):
see my other comments for reference

jrjohnson:
Both of these are issues in common that were introduced in the last update, I've added new tickets there to fix them.

dartajax (approved):

Issue #6716, opened 2022-06-21
Reduce Size of Logo


dartajax:
This is on netlify and probably should not be pushed out as is.

I am logged in as a student so there is no access to the flyout menu. You would know and identify with "Ilios", or at least be forced to face it in its current context. See screen shot below.

<img width="1429" alt="Screen Shot 2022-06-21 at 2 34 06 PM" src="https://user-images.githubusercontent.com/7493331/174901292-4054fb16-887d-4a45-b46e-01e269903f1e.png">

PR #6717, opened 2022-06-21
corrects layout on login screen and student dashboard.

stopfstedt:
account for the fact that pages without navigation only have three rows.

fixes https://github.com/ilios/frontend/issues/6716

fixes https://github.com/ilios/frontend/issues/6718


dartajax (approved):
`unborked` = true

jrjohnson (approved):

Issue #6722, opened 2022-06-22
fix template linter warnings


stopfstedt:
on today's menu are:

```
$ npm run lint:hbs

> ilios@31.0.0 lint:hbs
> ember-template-lint .

app/components/course-search-result.hbs
  35:8  warning  <span> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children

app/components/global-search-box.hbs
  43:12  warning  <li> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children

app/components/learnergroup-calendar.hbs
  20:6  warning  <li> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children
  26:6  warning  <li> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children

app/components/learnergroup-instructor-manager.hbs
  41:16  warning  <li> has a role of button, it cannot have semantic descendants like <UserNameInfo>  require-presentational-children
  42:16  warning  <li> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children
  59:18  warning  <span> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children
  61:18  warning  <span> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children

app/components/learnergroup-list.hbs
  99:22  warning  <span> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children
  101:22  warning  <span> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children

app/components/login-form.hbs
  28:12  warning  No autofocus attribute allowed, as it reduces accessibility by moving users to an element without warning and context  no-autofocus-attribute

app/components/manage-users-summary.hbs
  87:18  warning  <li> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children

app/components/new-myreport.hbs
  115:18  warning  <li> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children

app/components/school-vocabularies-list.hbs
  56:20  warning  <span> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children

app/components/school-vocabulary-manager.hbs
  54:14  warning  <span> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children
  68:16  warning  <li> has a role of button, it cannot have semantic descendants like <FaIcon>  require-presentational-children
[...]

PR #6724, opened 2022-06-23
Fix template linter warnings

stopfstedt:
fixes https://github.com/ilios/frontend/issues/6722

also re-colors a lot of the trashbins in _red_ - i know we had a discussion on that before but can't find the ticket (anymore).


app/components/login-form.hbs
```diff
@@ -25,7 +25,6 @@
           <input
             autocapitalize="off"
             autocorrect="off"
-            autofocus=""
```
jrjohnson, line 28: I think we can keep this [the docs](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-autofocus-attribute.md#no-autofocus-attribute) says we can use it on elements that 
> serve as the main purpose of the page

jrjohnson (changes requested):

app/components/login-form.hbs
stopfstedt: :heavy_check_mark:

jrjohnson (approved):

PR #6727, opened 2022-06-23
Fix navigation on tablets

jrjohnson:
On a tablet sized screen in some cases for no particular reason that I
can tell the navigation will get huge when some items are selected.
Setting the body to 1fr seems to fix this so it may just be a browser
bug with double auto and our many hidden items in a collapsed navigation
view.

Bad:
Screen Shot 2022-06-22 at 10 28 30 PM

and now better:
Screen Shot 2022-06-22 at 10 28 16 PM


stopfstedt (approved):
LGTM

PR #6731, opened 2022-06-23
show totals on instructor- and learnergroup lists.

### LINKED ISSUE #6729: Include Record Count In Learner Groups and Instructor Groups
We do this in Courses and Sessions. I believe it would be helpful in Learner Groups as well - less so in Instructor Groups, but maybe there too - gonna tag `Team Discussion` so we can determine that.

see below ...

image

stopfstedt:
fixes #6729


dartajax (approved):
I like these changes.

Issue #6737, opened 2022-06-24
Move Member Count Up on Learner Group Management


dartajax:
Here is how it is ...

![image](https://user-images.githubusercontent.com/7493331/175653359-70641d0f-1983-464c-b258-e19687f90b1b.png)

This might be better ...

![image](https://user-images.githubusercontent.com/7493331/175654957-6698ada5-c4fe-4fba-a337-8f0744484c35.png)

We can discuss whether the `Hide Calendar` `Show Calendar` buttons can be up that high. They can be below and still maintain the height of the section.

Issue #6581, opened 2022-03-24
use page objects everywhere


stopfstedt:
these are the tests that are currently using Qunit DOM and/or ember test helper functions.

```
$ egrep -lr '(assert.dom| find\(| findAll\(| click\()' tests | sort
tests/acceptance/admin-test.js
tests/acceptance/header-test.js
tests/acceptance/programs-test.js
tests/acceptance/users-test.js
tests/acceptance/user-test.js
tests/integration/components/bulk-new-users-test.js
tests/integration/components/connection-status-test.js
tests/integration/components/curriculum-inventory/report-details-test.js
tests/integration/components/dashboard-loading-test.js
tests/integration/components/error-display-test.js
tests/integration/components/global-search-test.js
tests/integration/components/instructor-groups/loading-test.js
tests/integration/components/learnergroup-calendar-test.js
tests/integration/components/learnergroup-header-test.js
tests/integration/components/learnergroup-instructor-manager-test.js
tests/integration/components/learnergroup-summary-test.js
tests/integration/components/manage-users-summary-test.js
tests/integration/components/new-myreport-test.js
tests/integration/components/program-year/header-test.js
tests/integration/components/program-year/objective-list-loading-test.js
tests/integration/components/user-profile-calendar-test.js
tests/integration/components/user-profile-cohorts-test.js
tests/integration/components/user-profile-ics-test.js
tests/integration/components/user-profile-learnergroups-test.js
tests/integration/components/user-profile-roles-test.js
tests/integration/components/visualizer-program-year-competencies-test.js
tests/integration/components/visualizer-session-type-terms-test.js
tests/integration/components/visualizer-session-type-vocabularies-test.js
tests/integration/components/yes-no-test.js
```

replace their usage with ember cli page objects.

PR #6738, opened 2022-06-27
revamps learnergroup user management toolbar.

stopfstedt:
fixes https://github.com/ilios/frontend/issues/6737

this presented an opportunity to use page objects for test coverage here, so this also refs #6581 

while viewing:
image

while managing membership:
image

while bulk assigning:

image


dartajax:
While this ...

image

... is already a great improvement, would it be possible to have something like this ...

image

... instead?

dartajax:
One more .. sorry.

While this ...

image

... is a great improvement. Can we do something like this ...

image


... instead?

jrjohnson (approved):

PR #6789, opened 2022-08-03
use :hover pseudo class to style clicked button.

### LINKED ISSUE #6788: Click Event in Admin Console Highlights All Records Rather than the One Selected
To reproduce ...

 * login as an admin user 
 * do a search for a common name in the Admin Console
 * determine that the third record appearing in the result set is the one you wish to review
 * click on the third record to select it 

Notice that all rows of the search results are highlighted. This led me to a bit of confusion since I was no longer confident that I had selected the correct record. Once the user profile appears, you can verify your selection; but after clicking on the desired record only the selected record should be highlighted - not all of them.

stopfstedt:
this scopes the visual indication of a user transition down to the user
record that was clicked.

fixes #6788

### DIFF
--- a/app/components/manage-users-summary.hbs
+++ b/app/components/manage-users-summary.hbs
@@ -55,9 +55,7 @@
             )
           )
           "hidden"
-        }}
-
-        {{if this.clickUser.isRunning "navigating"}}"
+        }}"
     >
       {{#if this.searchForUsers.isRunning}}
         <li>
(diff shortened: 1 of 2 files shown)


dartajax (commented):
I can't test this because search doesn't work on demo currently.

stopfstedt:
> I can't test this because search doesn't work on demo currently.

ah balls. well, @jrjohnson should be able to test this locally without elasticsearch. i'm un-assigning you from this @dartajax.

jrjohnson (approved):
Oooh, I love me some CSS only solutions!

PR #6796, opened 2022-08-04
improves left-hand alignment of components on program details screen.

### LINKED ISSUE #6793: Better Alignment Needed in Program Years
How it is ...

image

It should be more like this ...

image

stopfstedt:
fixes #6793 

please review the program-year detail screen as well while at it - i tried my hand in better (not perfect) alignment of components along their left/right margins as well.


dartajax (approved):
looks good to me

jrjohnson (approved):
LGTM, I wonder if we should make these constants, but then I think it's probably too much work and this is good!

PR #6800, opened 2022-08-05
reworks program year leadership component

### LINKED ISSUE #6799: Improve Format of Program Year Director Selection Area
Here it is ...

image

I am suggesting something more like this ...

image

I tagged this with `Needs Team Discussion` so we can decide about the positioning on this screen regarding the search box and selected item - this type or usage pattern exists in many places in Ilios. Primarily I don't think the search box was properly positioned and we could make better use of the space with the selected records as well.

stopfstedt:
fixes #6799


dartajax (approved):
I saw a weird flashing thing that @stopfstedt also noticed / verified that I wasn't seeing things. That'll be another ticket since it happens in more than one area and is low priority anyway. I'm going to merge this one.

Issue #6808, opened 2022-08-09
My Reports and My Courses Should Be the Same Size


dartajax:
Some examples shown below ...

![image](https://user-images.githubusercontent.com/7493331/183760530-a66b5885-45ad-4307-9bea-23882a09e38b.png)

![image](https://user-images.githubusercontent.com/7493331/183760867-2938a4c6-9ff2-47c8-8591-4c5d26d9197f.png)

![image](https://user-images.githubusercontent.com/7493331/183761450-f0331e8d-da86-43ed-8b1a-af3afeb4f1d4.png)

PR #6809, opened 2022-08-09
use a grid layout instead of flex to ensure equal width for blocks.

stopfstedt:
fixes https://github.com/ilios/frontend/issues/6808


jrjohnson (approved):
Grids FTW!

dartajax (approved):

Issue #6807, opened 2022-06-24
Suggested Update to Instructor Group Maintenance


dartajax:
Here is how it is ...

![image](https://user-images.githubusercontent.com/7493331/175650716-cab69516-2bda-4415-bdd4-e827ed256631.png)

Maybe this would be an improvement ...

![image](https://user-images.githubusercontent.com/7493331/175656826-962e8e31-13ba-496c-a655-b69629f0e717.png)

Cleaner look - new caption `Add Instructor(s)` or `Add Instructors` - we will discuss.

PR #6820, opened 2022-08-16
Complete rework of Instructor Group details page

stopfstedt:
fixes https://github.com/ilios/frontend/issues/6807

### DIFF
--- a/app/services/data-loader.js
+++ b/app/services/data-loader.js
@@ -78,7 +78,7 @@ export default class DataLoaderService extends CommonDataLoaderService {
         instructorGroupId,
         this.store.findRecord('instructor-group', instructorGroupId, {
           reload: true,
-          include: 'users',
+          include: 'users,ilmSessions.session.course,offerings.session.course',
         })
       );
     }
--- a/app/services/data-loader.js
+++ b/app/services/data-loader.js
@@ -6,6 +6,8 @@ export default class DataLoaderService extends CommonDataLoaderService {
   #loadedSchoolInstructorGroups = new Map();
   #loadedCohortLearnerGroups = new Map();
   #loadedUserProfiles = new Map();
+  #loadedInstructorGroup = new Map();
+
   async loadCoursesForLearnerGroup(learnerGroupId) {
     if (!this.#loadedLearnerGroupWithCourses.has(learnerGroupId)) {
       const sessionIncludes = 'offerings.session.course';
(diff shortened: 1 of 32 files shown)


dartajax (approved):
You talkin' to me? I like it - approved.

jrjohnson (changes requested):
This is a great change. May be better done in a followup, but the data loading parts can be improved by pre-loading all the data in the instructor group route in the same way we do for [learner-group](https://github.com/ilios/frontend/blob/master/app/routes/learner-group.js) so all the data is pre-loaded in a single fetch and then you don't need to check for `isFulfilled` or have data loading spinners. Since most users are coming from the list you can probably add loading all related courses to [the existing data loader](https://github.com/ilios/frontend/blob/master/app/services/data-loader.js#L73) and then simple re-use that in the instructor group route.

stopfstedt:
> This is a great change. May be better done in a followup, but the data loading parts can be improved by pre-loading all the data in the instructor group route in the same way we do for [learner-group](https://github.com/ilios/frontend/blob/master/app/routes/learner-group.js) so all the data is pre-loaded in a single fetch and then you don't need to check for `isFulfilled` or have data loading spinners. Since most users are coming from the list you can probably add loading all related courses to [the existing data loader](https://github.com/ilios/frontend/blob/master/app/services/data-loader.js#L73) and then simple re-use that in the instructor group route.

handled :heavy_check_mark:

app/services/data-loader.js
```diff
@@ -70,6 +72,19 @@ export default class DataLoaderService extends CommonDataLoaderService {
 
     return this.#loadedLearnerGroup.get(learnerGroupId);
   }
+  async loadInstructorGroup(instructorGroupId) {
+    if (!this.#loadedInstructorGroup.has(instructorGroupId)) {
+      this.#loadedInstructorGroup.set(
+        instructorGroupId,
+        this.store.findRecord('instructor-group', instructorGroupId, {
+          reload: true,
+          include: 'users',
```
jrjohnson, line 81: Should also include all the things to load the courses view. `offerings.session.course,ilmSessions.course, maybe others I'm not thinking of`.

jrjohnson (changes requested):

app/services/data-loader.js
stopfstedt: added. the users and courses associated with the given instructor groups are all the extra data points we will need here.

jrjohnson (approved):

PR #6832, opened 2022-08-26
Test in all supported browsers

jrjohnson:
Using the browserstack API along with installing and running firefox ESR
versions to test all of our supported browser versions so we can ensure
everything works for all of our users.

Fixes ilios/ilios#4273

So..... 

Give-it-a-shot-and-hope-for-the-best-Dory-Quote

Requires ilios/common#3033


app/components/user-profile-ics.js
```diff
@@ -35,17 +35,32 @@ export default class UserProfileIcsComponent extends Component {
   /**
    * Generate a random token from a combination of
    * the user id, a random string and the current time
-   * Implementation lifted from https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest
-   * @param userId
-   * @return String
+   *
+   * We use two different methods for making this work. Because some versions of safari don't allow
```
stopfstedt, line 39: regarding the Safari compatibility issue  - please add a link to https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle#browser_compatibility in the comments, since it shows the problem in the compat matrix.

stopfstedt (changes requested):
minor request on code docs, otherwise this looks good to me.

app/components/user-profile-ics.js
jrjohnson: Good call, handled ✅

stopfstedt (approved):
LGTM

PR #6877, opened 2022-09-26
converts ember-concurrency tasks to arrow notation and async/await

stopfstedt:
refs https://github.com/ilios/common/pull/3031


stopfstedt:
won't fix at this time.

PR #6878, opened 2022-09-28
trying to fix a capitalization - caption fix

dartajax:
see title - just the "with" and "without" labels don't need to be Capitalized


stopfstedt (approved):
LGMT

PR #6903, opened 2022-10-14
Replace moment with luxon

jrjohnson:
Luxon is the modern moment replacement and we need to move to it everywhere, this mainly focuses on tests except for some CI templates where the data formatting was being done by moment instead of our internationalization library. There is one small UI change in a sequence block course summary from `yyyy-mm-dd` format to localized format.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "file-saver": "^2.0.5",
     "ilios-common": "^71.1.1",
     "loader.js": "^4.7.0",
+    "luxon": "^3.0.4",
     "miragejs": "^0.1.45",
     "mockdate": "^3.0.5",
     "moment-timezone": "^0.5.37",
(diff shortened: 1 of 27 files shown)


package.json
```diff
@@ -96,6 +96,7 @@
     "file-saver": "^2.0.5",
     "ilios-common": "^71.1.1",
     "loader.js": "^4.7.0",
+    "luxon": "^3.0.4",
```
stopfstedt, line 99: wer'e getting Luxon out of common already, so declaring it here again as a dependency is not necessary.

stopfstedt (changes requested):
overall this LGTM, see my comment on `package.json` for one minor change request.

package.json
jrjohnson: ember-auto-import got mad at me if I didn't include this. I think it's because it will eventually be required with embroider.

stopfstedt: fair enough. moving on then, _to be revisited_.

stopfstedt (approved):

PR #6961, opened 2022-11-28
Remove dependency on ember-composable-helpers

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4491
refs https://github.com/ilios/ilios/issues/4492


app/helpers/intersection-count.js
```diff
@@ -1,42 +0,0 @@
-/* eslint ember/avoid-leaking-state-in-ember-objects: 0 */
```
jrjohnson, line 1: Ooh, happy to see this go. Were we just not using it anywhere?

jrjohnson (approved):

app/helpers/intersection-count.js
stopfstedt: yea, was surprised too. we must have left this behind some while ago.

dartajax (changes requested):
image

dartajax:
It looks identical if only for the Browser message behind.

dartajax (approved):
I found one weird browser message that could not be duplicated a second time. I click tested through the application - I find these changes to be fine. Planning on merging this soon.

PR #6989, opened 2022-12-19
Use latest ubuntu for CI

jrjohnson:
This issue is resolved, maybe we can revert to using the latest ubuntu for running our tests.


dartajax:
You may need to remove the `Draft` status before this can be merged.

stopfstedt (approved):

stopfstedt:
i took the lock off so this can be pushed through.

PR #7006, opened 2023-01-11
Skip Linting for Integration Tests

jrjohnson:
For both helpers and components doing linting results in less readable code. Additionally we sometimes have to jump through hoops to make a test pretty like production code and most of our template linting just doesn't apply to tests. Let's disable template linting in test files and make tests pretty again.


stopfstedt (approved):
i concur.

PR #7011, opened 2023-01-14
chore: update ilios/common and prettier

Trott:


jrjohnson:
I believe this got resolved in #7013 which auto updates transitive dependencies like these so I'm going to close, but if something else needs to get merged happy to re-open it.

PR #7015, opened 2023-01-18
GraphQL Subject Reports

jrjohnson:
Taking advantage of our GraphQL endpoints for our existing subject reports speeds up query and rendering for these reports.

Because we no longer need so much boilerplate for loading the report results I've been able to split each report into its own component and handler rendering in templates which should make it much easier to augment reports with more data or formatting in the future.

WIP:
- [ ] needs to be tested with lots of real reports

### DIFF
--- a/app/components/reports/subject.hbs
+++ b/app/components/reports/subject.hbs
@@ -56,35 +56,9 @@
         </select>
       </div>
     {{/if}}
-    {{#if (is-array this.reportResultsList)}}
-      <ul class="report-results" data-test-results>
-        {{#each this.reportResultsList as |item|}}
-          {{#if item.model2}}
-            <li>
-              <LinkTo
-                @route={{item.route}}
-                @models={{array item.model item.model2}}
-              >
-                {{item.value}}
-              </LinkTo>
-            </li>
-          {{else if item.model}}
-            <li>
-              <LinkTo @route={{item.route}} @model={{item.model}}>
-                {{item.value}}
-              </LinkTo>
-            </li>
-          {{else}}
-            <li>
-              {{item.value}}
-            </li>
-          {{/if}}
-        {{else}}
-            <li>{{t "general.none"}}</li>
-        {{/each}}
-      </ul>
-    {{else}}
-      <LoadingSpinner />
-    {{/if}}
+    <this.subjectComponent
+      @report={{@selectedReport}}
+      @year={{@selectedYear}}
+    />
   </div>
 </section>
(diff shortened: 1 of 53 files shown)


stopfstedt (approved):

dartajax (approved):
looks good to me

dartajax:
This report did not return data on the PR version but it did return data on demo. The report is session types based on instructor selected. Here it is from the PR version ...

image

Here it is from demo ...

image

dartajax:
It appears to be fine when querying for session types based on term (well I think it's term anyway).

dartajax (changes requested):
see my recent comments - Session Types by Instructor didn't return any data using this Netlify PR deploy build or whatever - but records were returned using demo

dartajax (approved):

dartajax:
This one was from the All Terms in All Schools report. Data is slightly different between demo and this PR also - not sure which is more accurate or appropriate.

image

dartajax:
All Session Types >> All Schools >> MeSH not returning in GraphQL but ok on Netlify front-end environment

image

image

jrjohnson:
@dartajax - ready for a last round of testing I think and can be merged if it meets with your approval.

dartajax:
In This PR, no results found for the following ...

image

On demo and Netlify, 4 results are found ...

image

dartajax:
Sort order – all LM’s by Session Type – Lecture – the first in the sort I suspect have leading blanks – which should be trimmed and then it goes to quotes, numbers, and finally regular A-Z – at least in the sort order we should ignore the leading blank spaces 
 
The screen shot below shows where it changes the sort from [blank space first or whatever is going on here to orphan these records] to [quotations] - this is if and only if there are blank spaces in the LM titles such as ` Addiction 101 Video` - I don't see it when looking it up on prod in the database.
 
image

dartajax:
From demo (same as netlify)

image

From the PR (same report)

image

dartajax:
Font is ever so slightly larger with the Graph QL PR than it is on demo (lower screen shot)

image

image

dartajax:
This is Session Types by MeSH -- on demo ...

image

From the PR netlify ...

image

There is a divergence in data here. I think the return of "Lecture" is correct. I can go attach that MeSH to a different session type to verify. Wow so I added this MeSH term to a session of type "Practical (Lab) Exam" and it got added to "Lecture" on the demo and netlify latest versions AND it got returned by Graph QL - maybe Graph QL is the most accurate - either way, diverging data gets me to report out on it.

PR #7026, opened 2023-01-25
rm layout-breaking calculated width from global search results.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4592


dartajax (approved):
fine with me - I may have a follow-up "space saving" ticket down the road but no hurry - scroll bar appears to move down - results clickable - great stuff

PR #7028, opened 2023-01-25
realigns program-year and school header styles.

stopfstedt:
image

fixes https://github.com/ilios/ilios/issues/4521


dartajax (approved):
looks good to me

PR #7032, opened 2023-01-30
adds sortable headers to school session type table.

stopfstedt:
refs https://github.com/ilios/ilios/issues/4494

image

couple of notes:
- the original sort order was by active status, descending - that ensures that inactive session types are pushed to the bottom. in order to maintain this functionality, i added a new "active" column.
- filtering is out of scope. this is an interface that is hardly used as it is, i'm kinda drawing the line here. sorting will have to do, at least for the time being (got more important work to do anyway).
- i threw in sort-by-color as this was trivial to implement

### DIFF
--- a/app/components/school-session-types-list-item.hbs
+++ b/app/components/school-session-types-list-item.hbs
@@ -7,11 +7,13 @@
     <button class="link-button" type="button" {{on "click" (fn @manageSessionType @sessionType.id)}}>
       {{@sessionType.title}}
     </button>
-    {{#unless @sessionType.active}}
-      <em data-test-inactive>
-        ({{t "general.inactive"}})
-      </em>
-    {{/unless}}
+  </td>
+  <td data-test-active>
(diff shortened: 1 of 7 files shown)


dartajax (commented):
I am not so sure about this and I hate to say that. Without persistent headers and considering the number of session types in a school, I'm not certain how useful this will be. There are some sizing issues too I will show soon.

dartajax:
Here is the current sort order as I see it ...

image

dartajax:
Another weird thing happens when you look one of those session terms attached to session types visualizations. This happens ...

image

dartajax:
I am going to re-test this stuff against our main Netlify frontend build to see if the issues shown in the above screen shots actually exist there as well and as such, have nothing to do with this PR directly.

stopfstedt:
i'm retracting this PR based on today's team meeting discussion.

PR #7037, opened 2023-02-03
Display Errors Messages in Output

jrjohnson:
This error display page is usually shown when a browser error prevents the application from booting, it's super helpful to display the actual error in the output because we often get a screen shot of it. Having the message itself displayed ensures we don't need console logs to see what went wrong.

From:
before

To:
after


stopfstedt (approved):
LGTM

PR #7045, opened 2023-02-10
css minification is borked, turn it off.

stopfstedt:
refs https://github.com/ilios/ilios/issues/4622

see https://cli.emberjs.com/release/advanced-use/asset-compilation/#minifying


stopfstedt:
nevermind, chasing the wrong rabbit here.

PR #7046, opened 2023-02-10
Disable Embroider and CSS Minification

jrjohnson:
In order to fix issues with styles out of common I've disabled embroider by default placing it behind an environmental variable and then disabled css minification as well to fix a problem with the way colors were showing in some places. We'll need to investigate both of these, but for now this unlocks us.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -52,6 +52,9 @@ module.exports = function (defaults) {
     'ember-simple-auth': {
       useSessionSetupMethod: true, //can be removed in ESA v5.x
     },
+    minifyCSS: {
+      enabled: false,
+    },
   });
 
   if (process.env.BUILD_WITH_EMBROIDER) {
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -54,42 +54,47 @@ module.exports = function (defaults) {
     },
   });
 
-  return require('@embroider/compat').compatBuild(app, Webpack, {
-    staticAddonTestSupportTrees: true,
-    staticAddonTrees: true,
-    staticHelpers: true,
-    staticComponents: true,
-    // splitAtRoutes: [ //temporarily disabled route splitting for https://github.com/ilios/ilios/issues/4508
-    //   /admin[a-z-]*/,
-    //   'assign-students',
-    //   /course[a-z-]*/,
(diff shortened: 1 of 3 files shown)


jrjohnson:
@dartajax please click around and see if anything looks weird compared to production. I've switched out some CSS options here that could have strange consequences.

dartajax (commented):
This is on prod ...

image

image

dartajax:
You may be able to disregard my comment / review above since it has nothing to do with this exact PR - our Netlify current build also has the fainter borders for entry and search fields

dartajax:
Screen shot from PR frontend ...

image

Screen shot from prod ...

image

I'm not sure if this matters or not but since I am testing, I am reporting my findings.

dartajax:
I was trying to delete my comment above about the "borders" in the fields but I can't because it was part of my review. Sorry about that - you can disregard that comment. If it was just a comment, I would have deleted it. Just as an ...

dartajax:
I noticed with this PR that the hours of the day on the left of the calendar now match in format when you switch between day and week view - in prod they currently do not match - I think this is actually an improvement - so far even though I have been commenting up a storm, I have found no issues that would put this in `DO NOT MERGE` territory.

dartajax (approved):
I did not notice anything awry when testing against production - only format improvements.

ember-cli-build.js
```diff
@@ -52,44 +52,52 @@ module.exports = function (defaults) {
     'ember-simple-auth': {
       useSessionSetupMethod: true, //can be removed in ESA v5.x
     },
+    minifyCSS: {
```
stopfstedt, line 55: i don't think this is necessary anymore. i took this out and ran it with the `--environment=production` flag on `ember build/serve`, and it worked just fine. yanking embroider might be enough to set things right again.

stopfstedt (changes requested):

ember-cli-build.js
jrjohnson: Without that we lose border color on the calendar and other random places where color just doesn't get applied. It actually matches production then as @dartajax noted above, but production is currently broken.

stopfstedt: i stand corrected :+1:

stopfstedt (approved):

PR #7054, opened 2023-02-14
suppress flashing "none" output on subject reports reload.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4625


dartajax (approved):
looks good to me

PR #7055, opened 2023-02-15
makes subject report title editable.

stopfstedt:
adds in-place editing capabilities to subject reports.

image

image

fixes https://github.com/ilios/ilios/issues/4609

**Updated screenshot** - now with icon on edit-in-place title, and report description. 

image


app/components/reports/subject.hbs
```diff
@@ -31,6 +54,13 @@
     </div>
   </div>
   <div class="content-container">
+    <div class="description" data-test-report-description>
+      {{#if this.reportDescriptionLoaded}}
+        {{this.reportDescription}}
+      {{else}}
+        <LoadingSpinner />
```
jrjohnson, line 61: I'm not sure about the loading spinners here. Mostly they don't show up because the data for them is already loaded, but when they do it's so fast as to appear janky. Maybe call both `this.reporting.buildReportDescription(selectedReport)` and `this.reporting.buildReportTitle(selectedReport)` in the routes afterModel hook to ensure this data is loaded and then remove the loading indicators?

app/services/reporting.js
```diff
@@ -211,4 +240,112 @@ export default class ReportingService extends Service {
...
+      });
+    }
+
+    return this.intl.t('general.reportDisplayTitleWithoutObject', {
+      subject: subjectTranslation,
+      school: schoolTitle,
+    });
+  }
+
+  async buildReportDescription(report) {
```
jrjohnson, line 292: I think this is nearly identical to `buildReportTitle`, can the loading be refactored into a third method and then these two methods just do the translation?

app/styles/components/reports/subject.scss
```diff
@@ -1,28 +1,47 @@
-@use "../../ilios-common/mixins" as cm;
-@use "../../mixins" as m;
+@use "../../ilios-common/mixins" as m;
```
jrjohnson, line 1: I aliased all the common mixins as `cm` to make it a bit easier to see, while in the sass where something was coming from without having to scroll up. I don't see any way to code lint this, so it's just a convention at this point, but I think it's a good one.

jrjohnson (changes requested):

app/styles/components/reports/subject.scss
stopfstedt: :heavy_check_mark:

app/components/reports/subject.hbs
stopfstedt: :heavy_check_mark:

app/services/reporting.js
stopfstedt: :heavy_check_mark:

jrjohnson (approved):

PR #7056, opened 2023-02-16
Refactor Styles

jrjohnson:
Sass modules are the preferred way to link our mixins and variables together in new versions of sass. This is the least amount of change I could make work while still making an attempt to align with this new paradigm.

Requires ilios/common#3239

### DIFF
--- a/app/components/global-search-box.hbs
+++ b/app/components/global-search-box.hbs
@@ -26,7 +26,7 @@
     <FaIcon @icon="magnifying-glass"/>
   </button>
   {{#if this.hasResults}}
-    <ClickOutside @onClickOutside={{set this.autocompleteResults null}}>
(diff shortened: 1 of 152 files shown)


jrjohnson:
@dartajax this needs a click around style review for anything that looks weird, it's hopefully all working.

dartajax:
It looks even weirder in non-expanded mode - that's where I noticed it first ...

image

dartajax:
same is true with objectives ...

image

jrjohnson:
Style issues are fixed in https://github.com/ilios/common/pull/3267 I think we can probably move ahead with merging this and get those out with the next common update, which can be done pre-release of this change.

stopfstedt (approved):

PR #7111, opened 2023-04-04
Convert automerge to be a pull request action

stopfstedt:
We were using the special pull_request_target event here to be able to merge, but now that we're using the github CLI we may be able to do this with a normal pull request event and a checkout.

see https://github.com/ilios/common/pull/3314

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,8 +1,5 @@
 name: Dependabot auto-merge
 on: pull_request_target
-permissions:
-  pull-requests: write
-  contents: write
 jobs:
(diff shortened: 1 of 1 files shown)


jrjohnson:
I'd like to see this run successfully once on common, but I don't think we've had any matching updates there since making the change.

stopfstedt:
> I'd like to see this run successfully once on common, but I don't think we've had any matching updates there since making the change.

we have a pending caniuse-db PR here in frontend that could be our test case, see #7109. if it doesn't hold up then we can take it back to the workbench. don't think we'll have to wait for something to happen in common, necessarily.

PR #7119, opened 2023-04-06
rm await from instructor-group component templates.

stopfstedt:
replaced with Resources if/as applicable.

### DIFF
--- a/app/components/instructor-group/courses.js
+++ b/app/components/instructor-group/courses.js
@@ -1,0 +1,7 @@
+import Component from '@glimmer/component';
+import { use } from 'ember-could-get-used-to-this';
+import ResolveAsyncValue from 'ilios-common/classes/resolve-async-value';
+
+export default class InstructorGroupCoursesComponent extends Component {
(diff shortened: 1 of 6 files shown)


stopfstedt:
@dependabot rebase

app/components/instructor-group/courses.js
```diff
@@ -0,0 +1,7 @@
+import Component from '@glimmer/component';
+import { use } from 'ember-could-get-used-to-this';
+import ResolveAsyncValue from 'ilios-common/classes/resolve-async-value';
+
+export default class InstructorGroupCoursesComponent extends Component {
+  @use courses = new ResolveAsyncValue(() => [this.args.instructorGroup.courses, []]);
```
jrjohnson, line 6: This file seems unneeded as `instructorGroup.courses` isn't a promise, it's a getter on the model that already returns an empty array before it is resolved.

jrjohnson (changes requested):

app/components/instructor-group/courses.js
stopfstedt: :heavy_check_mark:

jrjohnson (approved):

PR #7152, opened 2023-05-03
Replace resolve async value resource

jrjohnson:
Codemoded removing these resources and replacing them with the TrackedAsyncData resource from ember-async-data. This is a rough auto pass, we'll need to touch these up.

wip:
- [ ] see what tests tell us
- [ ] clean up some of these that are offensively verbose / ugly


jrjohnson:
Going to re-open.

PR #7155, opened 2023-05-05
Enable Percy Snapshots for CI

jrjohnson:
Adding percy visual diffs to many of our tests in order to catch regressions that make ilios ugly. I could not find a way to protect the percy write only token and still allow tests to run from forks so I've added the token directly to the yaml in the same way we do on our simple charts runs.


stopfstedt (approved):
:fire:

dartajax:
Can't merge this due to the conflicts. Would love to but can't ... at this point in time.

PR #7175, opened 2023-05-22
Replace resolve async value resource

jrjohnson:
Remove these resources and replace them with the TrackedAsyncData resource from ember-async data. This performs better than our version and is on the paved path for ember. The declarations are move verbose requiring a double getter, but this is an OK compromise for now. With time we may be able to refactor some of these getters out and replace them with the load template helper from this addon.

### DIFF
--- a/app/components/learner-group/header.js
+++ b/app/components/learner-group/header.js
@@ -2,22 +2,48 @@ import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { dropTask } from 'ember-concurrency';
-import { use } from 'ember-could-get-used-to-this';
 import { validatable, Length, NotBlank } from 'ilios-common/decorators/validation';
-import ResolveAsyncValue from 'ilios-common/classes/resolve-async-value';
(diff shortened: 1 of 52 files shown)


app/components/learner-group/header.js
```diff
@@ -2,24 +2,50 @@ import Component from '@glimmer/component';
...
+    return this.upstreamRelationships?.school;
+  }
 
   get usersOnlyAtThisLevel() {
     return this.args.learnerGroup.usersOnlyAtThisLevel
       ? this.args.learnerGroup.usersOnlyAtThisLevel
       : [];
   }
 
+  async resolveUpstreamRelationships(learnerGroup) {
```
stopfstedt, line 40: :fire:

jrjohnson: I like this one too! Especially in some of our models where we're loading a big chain, I think this might be useful. It's a potential foot gun though because if any of those connections change this won't re-render. Safe here because we don't ever move a learner group into another cohort or a cohort into another program etc...

jrjohnson:
😢, good catch! Fixed.

stopfstedt (approved):

PR #7177, opened 2023-05-22
adds duplicate username prevention.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4737

image

image

### DIFF
--- a/app/components/new-user.js
+++ b/app/components/new-user.js
@@ -4,11 +4,12 @@ import { inject as service } from '@ember/service';
 import { all, filter } from 'rsvp';
 import { dropTask, restartableTask } from 'ember-concurrency';
 import moment from 'moment';
-import { validatable, IsEmail, Length, NotBlank } from 'ilios-common/decorators/validation';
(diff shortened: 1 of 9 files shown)


dartajax (approved):
seems to work fine for me - I'm wondering how it would be in a directory-based environment rather than form authentication - I am assuming it would work there as well - looks good to me

app/components/new-user.js
```diff
@@ -122,6 +127,17 @@ export default class NewUserComponent extends Component {
     });
   }
 
+  async validateUsernameCallback() {
+    const auths = await this.store.query('authentication', {
```
jrjohnson, line 131: Let's pre-load all authentication records at the top and then check the list. There aren't so many users that will break anything and the way it is built currently there is a new network request for every keystroke. Better to load the list when it is needed the first time and then use that list.

jrjohnson (changes requested):
Awesome. One network / performance change. UI is perfect and works great.

PR #7187, opened 2023-05-26
revamps user profile permissions toggles for a11y

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4783
fixes https://github.com/ilios/ilios/issues/4784

- toggles are now actual buttons, with aria- attributes etc.
- sub-sections (programs, courses, etc) without any relevant roles are collapsed by default and won't be expandable.


dartajax (approved):
looks good to me

PR #7190, opened 2023-05-30
Refactor New User to Remove Async Loading

jrjohnson:
In an effort to remove reliance on deprecated model async getters I ended up removing all the async loading from this component. By ensuring a pre-load before the component is loaded we can use the peek method of the store to get the data we need without needing much async in this component.


stopfstedt (approved):
LGTM

PR #7195, opened 2023-05-31
adds caching layer to username validation.

stopfstedt:
only issue a callback to the API endpoint if the username input actually changed.

fixes https://github.com/ilios/ilios/issues/4792

### DIFF
--- a/app/components/new-user.hbs
+++ b/app/components/new-user.hbs
@@ -115,9 +115,15 @@
           type="text"
           value={{this.username}}
           {{on "input" (pick "target.value" (set this.username))}}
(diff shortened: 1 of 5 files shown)


jrjohnson (approved):

stopfstedt:
@dartajax @jrjohnson  - i ended up rewiring the username validation. the check now happens when the user form is submitted, not on username input.

slightly less idea from a UX perspective, but at least it's stable.

anyway, please re-review. thanks.

dartajax (approved):
seems fine to me - added a user smoothly - everything else seems fine too

jrjohnson (approved):

PR #7232, opened 2023-06-12
rm await template helper with getter/Resource-backed equivalent.

stopfstedt:

### DIFF
--- a/app/components/school-competencies-list.js
+++ b/app/components/school-competencies-list.js
@@ -1,0 +1,26 @@
+import Component from '@glimmer/component';
+import { cached } from '@glimmer/tracking';
+import { TrackedAsyncData } from 'ember-async-data';
+import { map } from 'rsvp';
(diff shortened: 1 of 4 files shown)


app/components/school-competencies-list.js
```diff
@@ -0,0 +1,26 @@
+import Component from '@glimmer/component';
+import { cached } from '@glimmer/tracking';
+import { TrackedAsyncData } from 'ember-async-data';
+import { map } from 'rsvp';
+import { sortBy } from 'ilios-common/utils/array-helpers';
+
+export default class SchoolCompetenciesListComponent extends Component {
+  @cached
+  get data() {
+    return new TrackedAsyncData(this.loadData(this.args.domains));
```
jrjohnson, line 10: This is a little bit under-tracked because the `children` of a `domain` changing won't trigger this to re-compute, however where this is used we re-render the page when saving from manage competencies so it doesn't matter. I just wanted to mention it as a possible gotcha in the future.

PR #7254, opened 2023-06-21
replace await template helper from courses list

stopfstedt:

### DIFF
--- a/app/components/courses/list-item.hbs
+++ b/app/components/courses/list-item.hbs
@@ -1,0 +1,69 @@
+<tr
+  class="courses-list-item {{if (includes @course.id @coursesForRemovalConfirmation) "confirm-removal"}}"
+  data-test-courses-list-item
+>
+  <td class="text-left" colspan="8" data-test-course-title>
+    <LinkTo @route="course" @model={{@course}}>
+      {{@course.title}}
+      {{#if @course.externalId}}
+        ({{@course.externalId}})
+      {{/if}}
+    </LinkTo>
+  </td>
+  <td class="text-center hide-from-small-screen" colspan="1" data-test-level>
+    {{@course.level}}
+  </td>
+  <td class="text-center hide-from-small-screen" colspan="2" data-test-start-date>
+    {{moment-format @course.startDate "L"}}
+  </td>
+  <td class="text-center hide-from-small-screen" colspan="2" data-test-end-date>
+    {{moment-format @course.endDate "L"}}
+  </td>
+  <td class="text-right" colspan="2" data-test-status>
+    {{#if (includes @course.id @savingCourseIds)}}
+      <LoadingSpinner />
+    {{else}}
+      <PublicationStatus @item={{@course}} @showText={{media "isLaptopAndUp"}} @showIcon={{not (media "isLaptopAndUp")}} />
+      {{#if @course.locked}}
+        {{#if this.canUnlock}}
+          <button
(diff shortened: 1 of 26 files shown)


app/components/courses/list-item.hbs
```diff
@@ -0,0 +1,69 @@
...
+          </button>
+        {{else}}
+          <FaIcon @icon="lock" class="disabled" />
+        {{/if}}
+      {{else if this.canLock}}
+        <button
+          type="button"
+          class="link-button"
+          title={{t "general.lockCourse"}}
+          {{on "click" (perform this.lockCourse)}}
```
jrjohnson, line 46: Could maybe be `{{on "click" (fn @lockCourse @course)}}`?

app/components/courses/list-item.hbs
```diff
@@ -0,0 +1,69 @@
...
+      <LoadingSpinner />
+    {{else}}
+      <PublicationStatus @item={{@course}} @showText={{media "isLaptopAndUp"}} @showIcon={{not (media "isLaptopAndUp")}} />
+      {{#if @course.locked}}
+        {{#if this.canUnlock}}
+          <button
+            type="button"
+            class="link-button"
+            title={{t "general.unlockCourse"}}
+            {{on "click" (perform this.unlockCourse)}}
```
jrjohnson, line 33: Could maybe be `{{on "click" (fn @unlockCourse @course)}}`?

app/components/courses/list.js
```diff
@@ -39,21 +38,22 @@ export default class IliosCourseListComponent extends Component {
     return this.intl.t(translation);
   }
 
-  @task
-  *unlockCourse(course) {
-    const permission = yield this.permissionChecker.canUnlockCourse(course);
+  @action
+  async unlockCourse(course) {
```
jrjohnson, line 42: This should be a task, otherwise if the component gets destroyed while one of the `await`s is going it will cause a complete failure of the app since executing the next line will be on something that doesn't exist anymore.

app/styles/components.scss
```diff
@@ -150,3 +150,4 @@
 @import 'components/school/visualizer-session-type-vocabularies';
 @import 'components/school/visualizer-session-type-vocabulary';
 
+@import 'components/courses/list-item';
```
jrjohnson, line 153: it feels like the courses-list scss should also move here? `import 'components/courses/list';`?

jrjohnson (changes requested):
Excellent refactor! Couple of suggestions

app/styles/components.scss
stopfstedt: i didn't touch that on purpose, since it pertains to the controller template, and not the actual component. it's a mess.

app/components/courses/list.js
stopfstedt: :heavy_check_mark:

app/components/courses/list-item.hbs
stopfstedt: :heavy_check_mark:

app/components/courses/list-item.hbs
stopfstedt: :heavy_check_mark:

app/styles/components.scss
stopfstedt: :heavy_check_mark:

jrjohnson (approved):
Love it!

PR #7257, opened 2023-06-22
corrects import of filesaver library.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4803

### DIFF
--- a/app/components/program-year/objective-list.js
+++ b/app/components/program-year/objective-list.js
@@ -8,6 +8,7 @@ import { TrackedAsyncData } from 'ember-async-data';
 import { cached } from '@glimmer/tracking';
 import AsyncProcess from 'ilios-common/classes/async-process';
 import { mapBy, uniqueValues } from 'ilios-common/utils/array-helpers';
+import { saveAs } from 'file-saver';
 
 export default class ProgramYearObjectiveListComponent extends Component {
   @service iliosConfig;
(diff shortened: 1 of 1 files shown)


app/components/program-year/objective-list.js
```diff
@@ -73,8 +74,6 @@ export default class ProgramYearObjectiveListComponent extends Component {
     const resourcePath = `/programyears/${this.args.programYear.id}/downloadobjectivesmapping`;
     const host = this.iliosConfig.apiHost ?? `${window.location.protocol}//${window.location.host}`;
     const url = host + apiPath + resourcePath;
-    const { saveAs } = yield import('file-saver');
```
jrjohnson, line 76: The issue is that import filesaver returns an `FileSaver.default.saveAs` and not `FileSaver.saveAs`. Check my syntax on the destrucutring, but I think this should be `const { default: saveAs } = yield import('file-saver');`.

jrjohnson (changes requested):
Nice find!

Moving this to the top level works, but then the file-saver library gets imported into the main application bundle and downloaded for everyone all the time. It's not a huge amount of javascript (4k compressed), but every bit counts. Doing this as a dynamic import ensures that only users of this button have to download that file.

app/components/program-year/objective-list.js
stopfstedt: updated :heavy_check_mark:

jrjohnson (approved):

PR #7260, opened 2023-06-23
Update Ember CLI 4.12

jrjohnson:
Updated ember-cli and all blueprints to 4.12 and then fixed some linting errors caused by the updated libraries.


stopfstedt (approved):
LGTM

PR #7266, opened 2023-06-27
remove async template helper.

stopfstedt:
this is the least invasive approach i could come up with. :shrug: 


jrjohnson (approved):
LGTM

PR #7267, opened 2023-06-28
Enable embroider

jrjohnson:
Tests are passing and everything appears to be working, let's turn this on and see how it works.

Issues:
- [x] https://github.com/ilios/ilios/issues/4508

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,36 +74,6 @@ jobs:
       - name: test build
         run: npm run build
 
-  test-with-embroider:
-    name: Test With Embroider
-    runs-on: ubuntu-latest
-    timeout-minutes: 15
-    env:
-      BUILD_WITH_EMBROIDER: true
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 18.17
-          cache: npm
-      - run: npm ci
-      - run: npm run test:ember
-
-  build-with-embroider:
-    name: Build With Embroider
-    runs-on: ubuntu-latest
-    timeout-minutes: 10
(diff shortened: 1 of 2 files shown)


jrjohnson:
I'm seeing a speedup of 1-4 seconds loading time on 4G simulated network and several megabytes less javascript. Even compressed it's 500K less JS shipped. Yay. @dartajax please test this netlify build for any weirdness. Initial load and speed for students should be significantly improved.

I'd like to test on a variety of devices and with several users so let's not merge this yet, but it's ready for that testing in my opinion.

dartajax:
This is not a review but merely a comment. This PR acts weirdly if the session gets stale like when you go eat lunch and come back, which is what I did. I even raised a sentry upon rage-clicking after eating my lunch. @jrjohnson may want us to continue testing on this.

dartajax:
Except for my previous comment, my testing has gone fine on this PR - added `DO NOT MERGE` just so we don't accidentally force this through - more review / testing is still necessary.

jrjohnson:
@dartajax ready for another review, still in the do not merge status, but if it checks out we may be ready to merge this week.

dartajax:
image

image

dartajax:
I was testing against demo with my screen shots above. It turns out I should have been testing against our current netlify build. The current netlify build has the same issue as the screen shots above; therefore, this issue is not related to this PR but is another issue we will need to fix once these changes have been pushed up to demo, which they have not been pushed to ... yet.

dartajax (approved):
so far so good - I plan on testing this more still but it seems fine to me

jrjohnson:
@dartajax when you feel like this has been well tested it's ready to merge! 🎉

PR #7281, opened 2023-06-30
align the years filter in the subject reports left instead of right.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4835

image


dartajax (approved):
Looks good to me.

PR #7297, opened 2023-07-07
Get routes from common

jrjohnson:
We export these so they can be consumed, we should go ahead and use that very nice feature.


stopfstedt (approved):
LGTM

PR #7306, opened 2023-07-13
realigns capabilities checks for single user group removal/assignment…

stopfstedt:
… with bulk removal/assignment checks.

fixes https://github.com/ilios/ilios/issues/4751

### DIFF
--- a/app/components/learner-group/cohort-user-manager.hbs
+++ b/app/components/learner-group/cohort-user-manager.hbs
@@ -114,27 +114,25 @@
                   {{/if}}
                 </td>
                 <td class="text-left" colspan="1">
-                  {{#if user.enabled}}
-                    {{#if (includes user this.usersBeingMoved)}}
-                      <LoadingSpinner />
-                    {{else if (and @canUpdate (eq this.selectedUsers.length 0))}}
-                      <button
-                        type="button"
-                        class="inline-button"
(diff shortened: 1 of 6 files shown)


jrjohnson (approved):
I'm good with this change, but does the current user need to be root to do this action? I think maybe anyone who is allowed to edit should be able to do this because the core issue is that the learner group isn't disabled, just one of the members who now needs to be removed.

stopfstedt:
> I'm good with this change, but does the current user need to be root to do this action? I think maybe anyone who is allowed to edit should be able to do this because the core issue is that the learner group isn't disabled, just one of the members who now needs to be removed.

i'll have a look at the extraneous is-root-user check, perhaps this can be removed.

stopfstedt:
@jrjohnson - i removed the checks that were prohibiting _non-root users with learner-group edit permissions_  from adding/removing inactive users to/from learner groups.

@dartajax please take note on this.

dartajax:
I think this is a great change/update. @dartajax = noted. Do you want this merged?

stopfstedt:
> I think this is a great change/update. @dartajax = noted. Do you want this merged?

thanks @dartajax. let's run it by @saschaben  in the meeting today, I'd want him to sign off on this.

stopfstedt:
this issue was discussed in today's Ilios team meeting - Sascha signed off on this change.

PR #7318, opened 2023-07-24
autofocus when program-year objective editor is opened.

stopfstedt:
companion ticket to https://github.com/ilios/common/pull/3465
(won't work until the PR above lands and we get a new common release into frontend, but can be front-loaded)

refs https://github.com/ilios/ilios/issues/4360


jrjohnson (approved):

dartajax (approved):
I looked through the slight changes percy detected - I like the changes minor as they are - looks good to me

PR #7343, opened 2023-08-09
replaces all elements that have the button role with actual buttons. #3489 

stopfstedt:
companion PR to https://github.com/ilios/common/pull/3489

refs https://github.com/ilios/ilios/issues/4782


app/components/curriculum-inventory/verification-preview.hbs
```diff
@@ -12,145 +12,149 @@
       {{did-insert this.setScrollTarget "toc"}}
     >
       <li>
-        <span class="link clickable" role="button" {{on "click" (fn this.scrollTo "table1")}}>
+        <button class="link link-button" type="button" {{on "click" (fn this.scrollTo "table1")}}>
```
jrjohnson, line 15: Not sure these need both `link` and `link-button`, I think maybe just `link-button` would do. Otherwise it's kind of confusing, but I'm also ok with both if it makes this changes smaller.

tests/integration/components/school-session-attributes-test.js
```diff
@@ -92,6 +92,7 @@ module('Integration | Component | school session attributes', function (hooks) {
       @expand={{(noop)}}
     />`);
 
+    // await pauseTest();
```
jrjohnson, line 95: `//`

tests/integration/components/school-session-types-expanded-test.js
```diff
@@ -170,7 +170,7 @@ module('Integration | Component | school session types expanded', function (hook
       @setSchoolManagedSessionType={{(noop)}}
       @setSchoolNewSessionType={{(noop)}}
     />`);
-
+    //await pauseTest();
```
jrjohnson, line 173: `//`

jrjohnson (changes requested):
I can't even really describe how much removing these removes a source of long standing background anxiety for me about our app and its accessibility and usability. I love this.

app/components/curriculum-inventory/verification-preview.hbs
stopfstedt: i see your point.
but we actually need both styles here to make that button fully appear like a link - `link-button` gets us there mostly, but not completely. the `link` class adds the rest of the styles to make it look like a link. 
what it boils down is that we need to rename `link-button` to something like `inline-button` and replace it everywhere, and then redefine `link-button` as something that actually lives up to its name.
that's another push that needs to come out of common, we're too far along here in frontend to change that now.

tests/integration/components/school-session-attributes-test.js
stopfstedt: :heavy_check_mark:

tests/integration/components/school-session-types-expanded-test.js
stopfstedt: :heavy_check_mark:

jrjohnson (approved):

PR #7375, opened 2023-08-30
Update Ember to 5.2

jrjohnson:
The 5.x series of ember is stable and we've cleared our deprecation and fixed everything upstream that needed fixing. I *think* we're good to go here, will see what tests and some manual clicking say.


dartajax (approved):
pretty heavy side by side with current demo click review - no issues - looks good to me

stopfstedt (approved):

PR #7394, opened 2023-09-08
hides user profile calendar behind toggle-button by default.

stopfstedt:
this also moves the calendar up top, in order to be closer to the toggle controls.

fixes https://github.com/ilios/ilios/issues/4947


jrjohnson (approved):
Good code, great tests. I also like moving the toggle up top.

PR #7402, opened 2023-09-11
replaces calls to obsolete route transition method with supported equivalent

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4952

### DIFF
--- a/app/components/ilios-users.hbs
+++ b/app/components/ilios-users.hbs
@@ -65,14 +65,15 @@
       {{/if}}
     </section>
     {{#if this.searchForUsers.lastSuccessful}}
-      <PagedlistControls
-        @offset={{@offset}}
-        @limit={{@limit}}
-        @limitless={{true}}
-        @setOffset={{@setOffset}}
-        @setLimit={{@setLimit}}
-        data-test-top-paged-list-controls
-      />
+      <div data-test-top-paged-list-controls>
+        <PagedlistControls
+          @offset={{@offset}}
+          @limit={{@limit}}
+          @limitless={{true}}
+          @setOffset={{@setOffset}}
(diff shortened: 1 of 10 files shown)


app/components/ilios-users.hbs
```diff
@@ -65,14 +65,15 @@
...
     {{#if this.searchForUsers.lastSuccessful}}
-      <PagedlistControls
-        @offset={{@offset}}
-        @limit={{@limit}}
-        @limitless={{true}}
-        @setOffset={{@setOffset}}
-        @setLimit={{@setLimit}}
-        data-test-top-paged-list-controls
-      />
+      <div data-test-top-paged-list-controls>
```
jrjohnson, line 68: I don't know why, but I'm sort of bothered by doing a div wrap just for testing. I don't see an easy way to target the element even if we pass the `data-test-` into the component though so let's go with it, mainly it's just late and I'm ornery about all code today.

app/components/new-user.js
```diff
@@ -49,6 +49,7 @@ export default class NewUserComponent extends Component {
   get schoolsWithCreatePermissions() {
     return new TrackedAsyncData(
       filter(this.allSchools, async (school) => {
+        console.log(school);
```
jrjohnson, line 52: `//be cool console log school`

jrjohnson (changes requested):

app/components/ilios-users.hbs
stopfstedt: i know, i was annoyed by that too.

app/components/new-user.js
stopfstedt:
```suggestion
```

jrjohnson (approved):

PR #7404, opened 2023-09-12
prevent console function calls by default from passing linting.

stopfstedt:
refs https://github.com/ilios/common/pull/3544


jrjohnson (approved):
💯 X 💯 === THIS CHANGE ❤️

PR #7406, opened 2023-09-13
Temporarily add school config serializer

jrjohnson:
This [belongs in common](https://github.com/ilios/common/pull/3545), however I'm adding it here temporarily while we work on a release plan. This file should be removed with the next common update. This was we can do a release to fix ilios/ilios#4953 without needing to do any gymnastics to avoid the API bump in common.


stopfstedt (approved):
turbo-approved.

PR #7410, opened 2023-09-15
Save learner group changes in two batches

jrjohnson:
We were optimizing our network changes here by combining the remove and the add step for groups, however this doesn't always result in consistent data and sometimes results in errors in the API. Instead we should first remove and then add users to groups. This is less efficient, but prevents data errors so it's worth doing here.

Refs ilios/common#3548
Refs ilios/ilios#4953


stopfstedt (approved):
click tested and reviewed, LGTM.

dartajax (approved):
I did a bunch of click testing - works fine for me.

PR #7412, opened 2023-09-15
Replace deprecated outputs in release workflow

jrjohnson:
This set-output way of moving data between steps has been deprecated by github, this replaces it with the recommended environment variable alternative.


stopfstedt (approved):
clean - i like this very much.

PR #7413, opened 2023-09-15
corrects UI glitch with cohort selection in user profile management

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4965


dartajax (approved):
works just like it should - my approval is granted

jrjohnson (approved):

PR #7417, opened 2023-09-18
fixes state management in school session type manager

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4973

### DIFF
--- a/app/components/school-session-type-form.js
+++ b/app/components/school-session-type-form.js
@@ -75,6 +75,7 @@ export default class SchoolSessionTypeFormComponent extends Component {
     if (!isValid) {
       return false;
     }
+    this.clearErrorDisplay();
     yield this.args.save(
       this.title,
       this.calendarColor,
@@ -83,7 +84,6 @@ export default class SchoolSessionTypeFormComponent extends Component {
       this.selectedAamcMethod,
       this.isActive,
     );
-    this.clearErrorDisplay();
   }
 
   @dropTask
--- a/app/components/school-session-type-manager.hbs
+++ b/app/components/school-session-type-manager.hbs
@@ -1,7 +1,6 @@
 <div
   class="school-session-type-manager"
   {{did-insert (perform this.load)}}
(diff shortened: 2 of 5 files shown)


dartajax (dismissed):
I found that if you click the Assessment toggle, whatever value was in the AAMC method drop-down gets cleared out. Here it is before clicking the toggle ...

image

Here is the screen after clicking the toggle ...

image

Since this bug that I found involves the same control, I figured I'd put it here. If you end up wanting a new ticket for this, and want to close this since it does seems to work a lot better - except if you tinker with the Assessment toggle.

stopfstedt:
> I found that if you click the Assessment toggle, whatever value was in the AAMC method drop-down gets cleared out. Here it is before clicking the toggle ...
> 
> image
> 
> Here is the screen after clicking the toggle ...
> 
> image
> 
> Since this bug that I found involves the same control, I figured I'd put it here. If you end up wanting a new ticket for this, and want to close this since it does seems to work a lot better - except if you tinker with the Assessment toggle.

i checked the code and this behaviour appears to be intentional. it's not a regression in any case - this is how it currently works.

i believe the original reasoning for this was that if the session type is not an assessment, then it shouldn't be mapped to an AAMC Method.
paging @saschaben  for clarification.

dartajax:
If you decide to toggle it back before saving, I would hope or expect that whatever was set before would be restored and not lost.

stopfstedt:
> If you decide to toggle it back before saving, I would hope or expect that whatever was set before would be restored and not lost.

i disagree. you switch it off, it's gone. if it's a mistake, you can always cancel out of that form and do it again.

dartajax (approved):
There are two entirely distinct sets of data for the drop-down menu depending on whether or not the Assessment toggle has been selected. A new selection would be required if toggled - it's a bit of an outlier to have to save the previous selection in the interface especially since hitting "Cancel" will result in the previous value being saved. I approve this PR.

PR #7438, opened 2023-09-29
Temporarily pin node to v18.7

jrjohnson:
This is a workaround for a linux-only issue with node 18.8 that breaks webpack and prevents our app from building at all. While we await a backport and fix we can pin this version.


stopfstedt (approved):
:+1:

PR #7528, opened 2023-12-06
use component method to set duration on selection change.

stopfstedt:
this is necessary to perform the type conversion to integer before assigning a new value to the tracked variable.

fixes https://github.com/ilios/ilios/issues/5101


jrjohnson (approved):

dartajax (approved):
looks good to me

PR #7536, opened 2023-12-15
Run reports without saving them first

jrjohnson:
Added a "Run Report" button to new subject report creation that allows a report to be run without saving it. If it works as expected it can then be saved.
WIP needs:
- [x] translations
- [x] lots of clicking around
- [x] maybe some more tests


dartajax (commented):
I'm not sure if this a change request or not but here it goes ...

Steps to reproduce ...

* Run a Report with a custom name ...

image

* Save this report 
image

To be honest, the whole "run before saving" bit is not really part of what is the crux of the issue here.

* Erase the name of the report to go with the default instead because "Zap CIty" is not very indicative of the report's intentions.

image

Either press Return on the keyboard or click the green check box button. The title appears to return to the system generated one - see below.

image

* Click "Back to Reports" 
* Voila - see below 

image

stopfstedt (approved):

dartajax:
image

dartajax:
Another thing I noticed -- 

* run a report (new and unsaved) for all sessions related to Program which is BRIDGES or whatever you choose
* once the report completes and displays the data, click on the title of one of the sessions to review what's going on
* click the browser's back button to return to the report 
* notice it is not there - well it was never saved or not yet anyway but the initial run should probably remain, non?

dartajax:
I ended up getting this now that I am on this branch on my local environment. I saw my suggested change to `general.deleteReport` but haven't updated the translation files. If we decide against this, I never will and will undo the suggested change but again this is on my local environment and newly created branch.

image

dartajax (approved):

PR #7546, opened 2023-12-27
Use new service import from Ember

jrjohnson:
This is better, let's us it. Plus the old way is going to deprecated at some point.


stopfstedt (approved):
LGTM

PR #7560, opened 2024-01-08
Catch missing model 404s

jrjohnson:
I'm not sure if this broke or was never working but we're not displaying a nice error when a model is missing. This doesn't happen that often for us, but when something is deleted we need to do better than or standard error message because refreshing the page will just continue to be broken.  Instead we now display our "Rats!" message with a link back to the dashboard.

Fixes ilios/ilios#5127

### DIFF
--- a/app/controllers/error.js
+++ b/app/controllers/error.js
@@ -6,4 +6,12 @@ export default class ErrorController extends Controller {
   forceRefresh() {
     location.reload();
   }
+
+  get isA404() {
+    if (this.model?.errors.length > 0) {
+      return Number(this.model.errors[0].status) === 404;
+    }
+
+    return false;
+  }
 }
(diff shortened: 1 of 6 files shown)


stopfstedt (changes requested):
this does not work for me, hard stop on the default route (unauthenticated). i was expecting the login form, but got an exception instead. from the browser console:

```
Uncaught (in promise) TypeError: this.model.errors is undefined
    get isA404 error.js:19
...    
```

image

stopfstedt:
hmm, seems to be working fine on netlify. i'm having another go at it locally.

stopfstedt:
i figured it out - forgot that we have to provide the backend URL on startup as env variable now  since dotenv support is gone. *headdesk

all is well, this checks out.

stopfstedt (approved):
LGTM

PR #7589, opened 2024-02-01
Rename app to frontend

jrjohnson:
We called this package ilios, which is nice, but it's going to be annoying when dealing with it in a monorepo. I've changed the name everywhere (except the <title> tag) to be frontend.

Requires ilios/common#3680


stopfstedt (approved):
grepped the code-base, and couldn't find any stragglers.

LGTM

PR #7593, opened 2024-02-01
Migrate frontend into monorepo

jrjohnson:
For now the only repo within this monorepo is the frontend itself, but later we'll be moving everything else in as well. For now get it working alone.


jrjohnson:
Netlify issues will fix after merging and Percy failures are due to this being the first time Percy has been working in a while.

stopfstedt (approved):
LGTM, but i'm holding this back due to some routing issues coming out of common.

stopfstedt:
unblocked from merging again after further discussion with @jrjohnson. we'll address the mentioned issues as follow-up.

PR #7595, opened 2024-02-02
Switch off embroider again

jrjohnson:
We're having some issues with route splitting on sub course routes like print and publication_check. Will need to fix these and then re-enable embroider for our builds.

Refs ilios/ilios#5213


stopfstedt (approved):
404 route tests are currently failing with embroider, we'll fix this afterwards.

PR #7602, opened 2024-02-06
Cleanup our lint dependencies and scripts

jrjohnson:
Consolidated all of this at the top of the monorepo.


jrjohnson:
Browserstack tests are failing because of an issue with our account, but appear to be triggering the correct browser and workspace, so I think we're good to go.

stopfstedt (approved):

PR #7607, opened 2024-02-07
Move acceptance tests into the frontend app

jrjohnson:
This will make it much easier to spot issues, all of the integration tests stay where they are closer to the component.


jrjohnson:
The embroider failures are real and the reason we had to revert the embroider go live. Once this is merged we can start plugging away at them.

PR #7611, opened 2024-02-07
Remove test-app chrome

jrjohnson:
The test app isn't for manual testing anymore, all of that can now be done in place in the frontend so we can ditch the chrome.


jrjohnson:
never mind, we'll need to bring the LTI's into this mono repo first so they can test the back link functionality in themselves. Not willing to remove that test yet so most of this has to stay.

PR #7615, opened 2024-02-07
fixes right-hand alignment of cancel btn in objective sort manager.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/5229

image


dartajax (approved):
looks great - gonna merge

PR #7633, opened 2024-02-12
Pin @sentry/ember version

jrjohnson:
This update is breaking our ability to mock the router service, pinning at a previous release.


stopfstedt:
Sherlock @jrjohnson !!  how'd you sleuth this one out?

jrjohnson:
I'm pretty proud 😊!  When I saw the unrelated #7632 was also failing I ran it on `master` and go lucky to spot some sentry references in the console output. Then I remembered that we'd had an issue with the required status checks and saw that the update was merged automatically without all the checks passing.

PR #7637, opened 2024-02-15
Moment to Luxon migration - models

stopfstedt:
i'm piece-mealing this a bit, just models for now.

for reference: https://github.com/moment/luxon/blob/master/docs/moment.md

refs https://github.com/ilios/ilios/issues/5193

### DIFF
--- a/packages/ilios-common/addon/models/course.js
+++ b/packages/ilios-common/addon/models/course.js
@@ -294,7 +294,11 @@ export default class Course extends Model {
 
   setDatesBasedOnYear() {
     const today = DateTime.now();
-    const firstDayOfYear = DateTime.fromFormat(this.year + '-07-01', 'yyyy-LL-dd');
+    const firstDayOfYear = DateTime.fromObject({
+      year: this.year,
+      month: 7,
+      day: 1,
+    });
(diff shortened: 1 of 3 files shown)


packages/ilios-common/addon/models/course.js
```diff
@@ -293,12 +293,12 @@ export default class Course extends Model {
   }
 
   setDatesBasedOnYear() {
-    const today = moment();
-    const firstDayOfYear = moment(this.year + '-7-1', 'YYYY-MM-DD');
+    const today = DateTime.now();
+    const firstDayOfYear = DateTime.fromFormat(this.year + '-07-01', 'yyyy-LL-dd');
```
jrjohnson, line 297: I think the `fromObject` is a bit clearer than `fromFormat`. 

```
const firstDayOfYear = DateTime.fromObject({
  year: this.year,
  month: 7,
  day: 1,
});
```

jrjohnson (commented):
A suggestion, but I can be talked into merging as is.

packages/ilios-common/addon/models/course.js
stopfstedt: good catch.

replaced :white_check_mark:

jrjohnson (approved):

PR #7640, opened 2024-02-15
moves lti-course-manager into monorepo.

stopfstedt:


jrjohnson (changes requested):
Editor seems to have munged the `yaml` again. Otherwise Perfect.

PR #7645, opened 2024-02-16
update ember-concurrency task implementations

stopfstedt:
- [x] convert tasks from modifier, `yield` syntax to fat-arrow, `async/await` equivalent
- [ ] ~~in templates, replace `perform` helper with calls to the `perform()` method directly on the givens Tasks~~ 

 on second thought - pass on the `perform` helper replacement for now.
 
from the [v4.0.0 release notes](https://github.com/machty/ember-concurrency/releases/tag/4.0.0):

> (Not really an enhancement, but) users are encouraged to directly use a task's bound .perform method (and curry with fn helper if needed) rather than to use the classic perform helper, which doesn't really add much value anymore

looking at how pervasive our use of the `perform` helper throughout common/frontend is, this seems like a pretty big lift without much to gain from at this point. 
suggesting we wait until that helper gets deprecated in future versions (having proper deprecation output will make it easier to find and replace anyway).

### DIFF
--- a/packages/frontend/app/components/assign-students.js
+++ b/packages/frontend/app/components/assign-students.js
@@ -39,17 +39,16 @@ export default class AssignStudentsComponent extends Component {
     return this.args.students.length - this.savedUserIds.length;
   }
 
-  @restartableTask
-  *load(element, [school]) {
-    let cohorts = yield this.store.query('cohort', {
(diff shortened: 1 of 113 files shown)


jrjohnson (approved):
Skimmed it, tests pass. 🚀

PR #7646, opened 2024-02-16
Re-enable browserstack tests

jrjohnson:
We can run these periodically to determine if our account has been fixed. I've also added the LTIs and upped the max-parallel setting because our account is supposed to have 5 parallel enabled.


jrjohnson:
Marking this as ready, we're in another trial, I think this time we'll get the contract done.

stopfstedt (approved):

PR #7651, opened 2024-02-21
adds babel transforms to build config.

stopfstedt:
noticed that the login form on the test app was broken with an ember-concurrency related error. 

this fixes it.

i threw the ember-auto-import transform into the mix, might not be necessary but doesn't seem to hurt.


jrjohnson (approved):
Nice catch!

PR #7658, opened 2024-02-23
remove router mock from integration tests

stopfstedt:
> Zoey says...
If you are thinking of stubbing the router service in a unit or integration test, consider writing an acceptance test instead. Acceptance tests let you navigate through many routes in your app, and the router does not need to be stubbed in them. If you choose to stub the router, you will need to stub multiple methods.

see https://guides.emberjs.com/release/testing/testing-components/#toc_stubbing-services

refs https://github.com/ilios/frontend/pull/7656


jrjohnson (approved):
The weird situation with CourseSummaryHeader bothers me a bit, but I concur, it's only used in the pace we were checking it it was used....

PR #7668, opened 2024-02-29
Fix common specifier in LTIs

jrjohnson:
No longer using the published version, now we use what is in the workspace.


stopfstedt (approved):
today i learned.

PR #7669, opened 2024-02-29
Lint dependencies correctly

jrjohnson:
This removes an annoying build time error as well as actually linting dependencies in all of our apps.


stopfstedt (approved):
heckyea :fire:

PR #7670, opened 2024-02-29
Add Some Reporting by Academic Year

jrjohnson:
Now reports on some things can be run based on the academic year allowing easier access to some complicated things like instructors in a year or terms.

This required an API bump to add the filters we need (which will be in ilios/ilios#5264) so it isn't easily testable outside of a local build with that PR in it as well.

Fixes ilios/ilios#3594


stopfstedt (approved):
LGTM

PR #7678, opened 2024-03-04
Move linting into workspaces

jrjohnson:
In order to clean up our dependency tree I needed to move all of the linting back into the workspaces and out of the root. This will make updating the ember apps easier, but we'll have to figure out how to configure editors to read the config at each level.


stopfstedt (approved):

stopfstedt:
@jrjohnson this needs rebasing and merge-conflict resolution. should have looked here first before rolling dependency updates into master, sorry.

PR #7685, opened 2024-03-05
Build and test all apps with embroider

jrjohnson:
Added CI jobs to build and test and the infrastructure to do this and then cleaned up the relationship with ember-data in all the apps by moving it to dependencies of ilios-common and removing it from each app. This lets embroider handle making it available correctly.


.npmrc
stopfstedt, line 1: great idea. i appreciate the extra context in the commit message that comes with this as well.

packages/frontend/ember-cli-build.js
```diff
@@ -25,10 +25,7 @@ module.exports = function (defaults) {
 
     hinting: isTestBuild,
     babel: {
-      plugins: [
-        // eslint-disable-next-line
```
stopfstedt, line 30: interesting. so eslint was onto something here after all.

dartajax:
`Build (20)` Expected - is it stalled out?

PR #7687, opened 2024-03-05
correctly parse given target date input when transitioning from week/…

stopfstedt:
…month view to day view for multi-instance events.

we're getting an ISO formatted string, not a YYYY-MM-DD formatted day string. this fixes it.


dartajax (approved):
passed click test of exact thing that failed yesterday - looks good to me

jrjohnson (approved):
I feel like we're bouncing back and forth a bit on date interchange and maybe we need to sit down and choose an interchange format (maybe this is what typescript is for), but I accept this fix for sure.

PR #7690, opened 2024-03-06
applies vertical margins to expanded program-year objective component.

stopfstedt:
this visually aligns it better with its sibling components in py objective management.

fixes https://github.com/ilios/ilios/issues/5277


dartajax (approved):
LGTM = looks good to me

PR #7692, opened 2024-03-06
Use husky to run pre-commit lint

jrjohnson:
This automates the linting job so every commit gets a check. This restore the functionality we had with the pre-commit package before moving to a monorepo.


stopfstedt (approved):
yes. 

i should ask you for stuff more often :laughing:

PR #7696, opened 2024-03-07
Cleanup release process after workspace consolidation

jrjohnson:
I've synced up the version for the LTI dashboard and it will increment with frontend now whenever we tag a new release. I consolidated the release workflows into a single pnpm run that will release all three apps in the same run.  LTI Course Manager hasn't ever been released, so I left it's version at 0 and removed the release script and didn't set it up to be incremented with a new tag.


stopfstedt (approved):
reviewed. this looks good to me.

PR #7700, opened 2024-03-08
Revert "Bump the embroider group with 4 updates"

stopfstedt:
Reverts ilios/frontend#7699

this reverts a dependency update that breaks the frontend.


stopfstedt:
nevermind - something broke locally, it works again now.

PR #7712, opened 2024-03-13
Add code owners file

jrjohnson:
Github allows us to check in this file to establish ownership of the repository and require some reviews. We can do many more complicated things with this concept, but I've checked in the basics for now.


stopfstedt (approved):
i approve of this.

PR #7719, opened 2024-03-15
Fixes flash-of-content during saving of program- and program-year-leadership.

stopfstedt:
Going back to the modifier-based approach to loading/unloading component state is the solution here.
This realigns this leadership management component with the one used for course management.

fixes https://github.com/ilios/ilios/issues/5139

also fixes https://github.com/ilios/ilios/issues/5324

### DIFF
--- a/packages/frontend/app/components/program-year/leadership-expanded.hbs
+++ b/packages/frontend/app/components/program-year/leadership-expanded.hbs
@@ -1,33 +1,42 @@
 <div
   class="program-year-leadership-expanded"
   data-test-program-year-leadership-expanded
-  ...attributes
+  {{did-insert (perform this.load) @programYear}}
 >
+{{#if this.load.isRunning}}
+    <LoadingSpinner />
+{{else}}
   <div class="program-year-leadership-expanded-header">
-    <button
-      class="title link-button"
-      type="button"
-      aria-expanded="true"
-      data-test-title
-      {{on "click" @collapse}}
-    >
-      {{t "general.leadership"}} ({{this.count}})
-      <FaIcon @icon="caret-down" />
-    </button>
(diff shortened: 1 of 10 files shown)


dartajax (approved):
This fixes both of those issues. Looks good to me.

packages/frontend/app/components/program-year/leadership-expanded.hbs
```diff
@@ -1,33 +1,42 @@
 <div
   class="program-year-leadership-expanded"
   data-test-program-year-leadership-expanded
-  ...attributes
+  {{did-insert (perform this.load) @programYear}}
 >
+{{#if this.load.isRunning}}
```
jrjohnson, line 6: I think instead of taking a step back into modifiers we could just observe `this.programYearDirectors.isResolved` here and get the same behavior without any other change.

jrjohnson (changes requested):
I'm not clear on why we'd go back to the modifier approach here, seems like a step back. Am I not seeing it?

packages/frontend/app/components/program-year/leadership-expanded.hbs
stopfstedt: @jrjohnson i tried the approach where is checked the state of the Resource in the template as you described,  and it does not eliminate the flash of content.

to your point - going back to modifiers feels like a regression, so this is a bit of a knee-jerk solution to the problem. on the other hand - this works. 

i can take this back to the bench to see if i can make this work with Resources somehow. let me know, thanks.

jrjohnson: Going to merge as is and @stopfstedt will follow up.

PR #7728, opened 2024-03-20
Go back to a resource-based loading approach.

stopfstedt:
moving the resetting the buffers up in the saving process eliminates the flash-of-content that occurred in previous iterations of this component.

converts all leadership management components to this pattern.

fixes https://github.com/ilios/ilios/issues/5337


jrjohnson (approved):
LGTM

PR #7729, opened 2024-03-21
Handle invalid calendar date gracefully

jrjohnson:
If an invalid date is specified as a query param in the URL we fall through to today instead of dying horribly.

I elected, after some experimentation, to not show this error to the user. It's in the console for power users and troubleshooting, however I don't think this case is common enough to warrant more explicit error handling.

Fixes ilios/ilios#5342


stopfstedt (approved):
nice solution. LGTM

PR #7732, opened 2024-03-21
We're seeing errors from JS hitting this line

jrjohnson:
For some reason not all the content here has an error object. No idea why that would be true, but it's easy to protect against.

Fixes ilios/ilios#5348


stopfstedt (approved):
ok.

PR #7736, opened 2024-03-22
Roll all leadership-expanded components int one shared component

stopfstedt:
fixes ilios/ilios#5347


packages/ilios-common/addon/components/leadership-expanded.js
```diff
@@ -18,31 +18,61 @@ export default class CourseLeadershipExpandedComponent extends Component {
   }
 
   @cached
-  get courseDirectors() {
-    return new TrackedAsyncData(this.args.course.directors);
+  get modelHasDirectors() {
+    return 'directors' in this.args.model;
```
jrjohnson, line 22: This is terrifying to me and I'm not sure why. Nice though, looks good.

packages/ilios-common/addon/components/leadership-expanded.js
```diff
@@ -18,31 +18,61 @@ export default class CourseLeadershipExpandedComponent extends Component {
...
+    }
+    return null;
   }
 
   get directors() {
-    const directors = this.courseDirectors.isResolved ? this.courseDirectors.value.slice() : [];
+    if (!this.modelHasDirectors) {
+      return [];
+    }
+    const directors = this.modelDirectors.isResolved ? this.modelDirectors.value.slice() : [];
```
jrjohnson, line 63: I'm not sure these need to get sliced. IIRC in Ember data these are arrays now or at least array proxies and don't need special handling anymore.

packages/ilios-common/addon/components/leadership-expanded.js
stopfstedt: yea, it's low-level and a somewhat obscure, feels a bit odd to use. 

i checked the ember-data docs for something that could do introspection, like a 'hasAttribute()' method or the likes, but couldn't find anything.

packages/ilios-common/addon/components/leadership-expanded.js
stopfstedt: nice catch. removed. we should do that everywhere then.

stopfstedt: :heavy_check_mark:

packages/ilios-common/addon/components/leadership-expanded.js
stopfstedt: :shrug:

jrjohnson (approved):

PR #7737, opened 2024-03-25
Refactor UserProfileCalendar component

jrjohnson:
The impetus of this was to fix the component's data loading on Sunday night in en-us where the data from the previous week would get loaded instead. I replaced the loading modifier with a tracked async getter and linked this up to the localeDays service.

I also replaced the loader with the built in one that is now in the weekly calendar.

### DIFF
--- a/packages/frontend/app/components/user-profile-calendar.js
+++ b/packages/frontend/app/components/user-profile-calendar.js
@@ -1,34 +1,49 @@
 import Component from '@glimmer/component';
 import { service } from '@ember/service';
 import { DateTime } from 'luxon';
-import { tracked } from '@glimmer/tracking';
+import { tracked, cached } from '@glimmer/tracking';
 import { action } from '@ember/object';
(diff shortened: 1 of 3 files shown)


packages/frontend/app/components/user-profile-calendar.js
```diff
@@ -1,34 +1,49 @@
...
   @service fetch;
   @service iliosConfig;
   @service userEvents;
+  @service localeDays;
+
   @tracked date = new Date();
-  @tracked calendarEvents;
 
-  load = dropTask(async () => {
-    //luxon weeks always stort on Monday to we have to adjust for that
```
stopfstedt, line 17: "stort" :laughing:

jrjohnson: Didn't really need to git blame to know who that line belonged to 😁

stopfstedt (approved):
LGTM.

PR #7741, opened 2024-03-25
Use tag for release name

jrjohnson:
Sentry doesn't seem to be able to guess our release anymore and is sending it as the commit hash instead. I've subbed in the github.ref_name variable as the docs indicate that will contain the name of tag which triggered the workflow run.


stopfstedt (approved):
i accept this.

PR #7746, opened 2024-03-28
Add sweet new qunit theme

jrjohnson:
There is an RFC to put this in the default blueprint, I like it, let's use it!


stopfstedt (approved):
sweet new theme indeed!

PR #7753, opened 2024-04-03
Replace ignored lint rule with todo

jrjohnson:
Instead of disabling the no-at-ember-render-modifiers linting rule I've added todo for each existing case. This will give us a runway for fixing this issue and ensure we don't create any new cases.


stopfstedt (approved):
nice. thanks!

PR #7754, opened 2024-04-03
Replace async data loading in report subject components

jrjohnson:
Refs ilios/ilios#5371


stopfstedt (approved):
nice drop-in replacement.

PR #7758, opened 2024-04-05
Add v prefix to sentry errors

jrjohnson:
Our releases are prefixed with this in the tag and in the data we're sending to sentry describing the build. However we pull the info from the package.json in frontend where npm asks for the version without the prefix. We need to add it to the sentry errors to ensure our error reports can line up with released versions and their source maps and commit data.


stopfstedt (approved):
this is so fiddly. approved.

PR #7764, opened 2024-04-05
Use github release note generation

jrjohnson:
Can our old approach and use something I don't have to maintain. Keeping the dartajax song names though!


stopfstedt (changes requested):
please copy https://github.com/ilios/ilios/blob/master/.github/release.yml to `.github/release.yml`, we need that config.

jrjohnson:
Whoops, great catch, thanks! Handled.

stopfstedt (approved):

PR #7770, opened 2024-04-09
Replaces ember-cli-new-version dependency with home-grown equivalent

stopfstedt:
fixes ilios/ilios#5285

i stripped this down to what i think is the absolute minimum necessary.

the in-place-plugin was necessary to add the version file creation and to add some info to the application config. i couldn't suss how how to do the former without a plugin, so here it is `¯\_(ツ)_/¯`

todo: 

- [x] add/update test coverage
- [x] dial in styles

### DIFF
--- a/packages/frontend/.eslintrc.js
+++ b/packages/frontend/.eslintrc.js
@@ -21,6 +21,12 @@ module.exports = {
     'ember/no-classic-classes': 0,
     'ember/no-mixins': 0,
     'ember/no-new-mixins': 0,
+    'n/no-extraneous-require': [
+      'error',
+      {
+        allowModules: ['broccoli-file-creator'],
+      },
+    ],
   },
   overrides: [
     // node files
--- a/packages/frontend/.eslintrc.js
+++ b/packages/frontend/.eslintrc.js
@@ -21,12 +21,6 @@ module.exports = {
     'ember/no-classic-classes': 0,
     'ember/no-mixins': 0,
     'ember/no-new-mixins': 0,
-    'n/no-extraneous-require': [
-      'error',
-      {
-        allowModules: ['broccoli-file-creator'],
-      },
-    ],
   },
   overrides: [
     // node files
--- a/packages/frontend/app/services/graphql.js
+++ b/packages/frontend/app/services/graphql.js
@@ -1,5 +1,4 @@
 import Service from '@ember/service';
-import fetch from 'fetch';
 import { service } from '@ember/service';
 
 export default class GraphqlService extends Service {
--- a/packages/frontend/app/services/new-version.js
+++ b/packages/frontend/app/services/new-version.js
@@ -1,6 +1,6 @@
 import Service from '@ember/service';
 import { tracked } from '@glimmer/tracking';
-import Ember from 'ember';
+import { isTesting } from '@embroider/macros';
 import { task, timeout } from 'ember-concurrency';
 import config from 'frontend/config/environment';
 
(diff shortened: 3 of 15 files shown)


packages/frontend/app/services/graphql.js
```diff
@@ -1,5 +1,4 @@
 import Service from '@ember/service';
-import fetch from 'fetch';
```
jrjohnson, line 2: I feel like this was important, but can't remember exactly why. Making a note for discussion.

packages/frontend/app/services/new-version.js
```diff
@@ -0,0 +1,57 @@
+import Service from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import Ember from 'ember';
```
jrjohnson, line 3: Importing `Ember` is on the way to final deprecation, use `import { isTesting } from '@embroider/macros';` instead to discover this.

packages/frontend/app/services/new-version.js
```diff
@@ -0,0 +1,57 @@
...
+
+  get currentVersion() {
+    return config.newVersion.currentVersion;
+  }
+
+  get isNewVersionAvailable() {
+    return this.latestVersion && this.currentVersion !== this.latestVersion;
+  }
+
+  get url() {
```
jrjohnson, line 20: This should go into the [fetch service](https://github.com/ilios/frontend/blob/master/packages/ilios-common/addon/services/fetch.js) where most of this is done already and is a nice single place to look for networking stuff.

packages/frontend/.eslintrc.js
```diff
@@ -21,6 +21,12 @@ module.exports = {
     'ember/no-classic-classes': 0,
     'ember/no-mixins': 0,
     'ember/no-new-mixins': 0,
+    'n/no-extraneous-require': [
+      'error',
+      {
+        allowModules: ['broccoli-file-creator'],
```
jrjohnson, line 27: This feels dangerous, `embroider` tends to hate it when things are used without importing. I have no idea if it's possible, but can this dependency go in `packages/frontend/lib/new-version/package.json`?

jrjohnson (changes requested):
Good approach, looks like a solid drop in replacement.

packages/frontend/app/services/graphql.js
stopfstedt: it's the only place that i found where were using the polyfill instead of the browser-native implementation of `fetch()`.
if need be, we could add `ember-fetch` as dependency to frontend.

packages/frontend/app/services/new-version.js
stopfstedt: thanks. replaced. :heavy_check_mark:

packages/frontend/app/services/new-version.js
stopfstedt: so i stripped the implementation of this `url()` getter down a bit. please see https://github.com/ilios/frontend/pull/7770/commits/147359c9b6138d36200dbbd9b101403aaa179233

re: using our own `fetch` service for this - i had a look and it's built for backend API interactions, and not for anything else.

jrjohnson: Makes sense. ✅

jrjohnson (approved):

jrjohnson:
Code looks great, going to deploy a version and play around with it locally before passing it off for merging.

jrjohnson:
Works perfect!

PR #7784, opened 2024-04-17
Refactor a few more element modifiers away

jrjohnson:
Refs ilios/ilios#5374


packages/frontend/app/components/global-search.js
```diff
@@ -108,13 +123,4 @@ export default class GlobalSearchComponent extends Component {
     this.args.onSelectPage(1);
     this.args.setIgnoredSchoolIds(ignoredSchoolIds);
   }
-
-  search = restartableTask(async (el, [query]) => {
```
stopfstedt, line 112: this is incomplete. the component's template is supposed to renders a spinner while the search task is running.

image

it now displays `No results` instead while search is running. 

image

stopfstedt (changes requested):
apart from the missing spinner on the global search, this looks good to me.

packages/frontend/app/components/global-search.js
jrjohnson: Fixed, added a test.

stopfstedt (approved):
LGTM

PR #7787, opened 2024-04-19
fixes getter/method mismatch in CurrentUser service

stopfstedt:
refs https://github.com/ilios/ilios/issues/5410


jrjohnson (approved):
Nice find!

PR #7794, opened 2024-04-24
Respect showCampusNameOfRecord setting

jrjohnson:
When configured to hide this data by configuration we don't show the icon or tooltip. The data is still in the API response, but it is hidden from the UI.


stopfstedt (approved):
reviewed and click-tested against https://github.com/ilios/ilios/pull/5417, LGTM

PR #7799, opened 2024-04-26
changed cohort label and value to be left-justified

michaelchadwick:
Looks like this just needed a class change from `text-center` to `text-left`.

fixes https://github.com/ilios/ilios/issues/4309


stopfstedt (approved):
LGTM

PR #7801, opened 2024-04-29
Update Ember to v5.8.1

jrjohnson:
Used ember-cli update across all addons and apps and updated all blueprints as well.


michaelchadwick:
Back in my Drupal days, I would do patch version updates of core and its deps without much trepidation, but that's because I'd been using Drupal for years. I just started using Ember, and so wanted to check: is updating a minor version (5.6.0->5.8.1) two versions (rather than going 5.6->5.7) common? No issues doing that in the past? As long as tests pass, we're good?

jrjohnson:
Ember is very good about BC and breakages, though I usually try and update us as new versions come out, I missed this one so we're skipping a bit. I'm also confident our tests would show any issues (as they have in the past). One of the reasons I try and update us early is so we can report any problems we surface in tests upstream.

michaelchadwick (approved):
All this looks good to me. Let's get current!

PR #7805, opened 2024-05-01
Add course years to Instructor/LearnerGroup displays

michaelchadwick:
InstructorGroups and LearnerGroups both show Associated Courses, but only the title. To fix ilios/3422, this adds the display of the course year range, starting with `course.year` and appending ` - course.year+1` if it crosses a year boundary.

### DIFF
--- a/packages/frontend/app/components/instructor-group/courses.hbs
+++ b/packages/frontend/app/components/instructor-group/courses.hbs
@@ -11,7 +11,7 @@
           <li data-test-course>
             <LinkTo @route="course" @model={{course}}>
               <FaIcon @icon="square-up-right" />
-              {{course.title}}
+              {{course.title}} ({{course.year}} - {{year-next course.year}})
             </LinkTo>
           </li>
         {{/each}}
--- a/packages/frontend/app/components/instructor-group/courses.hbs
+++ b/packages/frontend/app/components/instructor-group/courses.hbs
@@ -10,11 +10,11 @@
         {{#each (sort-by "title" @instructorGroup.courses) as |course|}}
           <li data-test-course>
             <LinkTo @route="course" @model={{course}}>
-              <FaIcon @icon="square-up-right" />
+              <FaIcon @icon="square-up-right" /> {{course.title}}
               {{#if this.academicYearCrossesCalendarYearBoundaries}}
-                {{course.title}} ({{course.year}} - {{year-next course.year}})
+                ({{course.year}} - {{add course.year 1}})
               {{else}}
-                {{course.title}} ({{course.year}})
+                ({{course.year}})
               {{/if}}
             </LinkTo>
           </li>
--- a/packages/frontend/app/components/instructor-group/courses.hbs
+++ b/packages/frontend/app/components/instructor-group/courses.hbs
@@ -11,7 +11,11 @@
           <li data-test-course>
             <LinkTo @route="course" @model={{course}}>
               <FaIcon @icon="square-up-right" />
-              {{course.title}} ({{course.year}} - {{year-next course.year}})
+              {{#if this.academicYearCrossesCalendarYearBoundaries}}
+                {{course.title}} ({{course.year}} - {{year-next course.year}})
+              {{else}}
+                {{course.title}} ({{course.year}})
+              {{/if}}
             </LinkTo>
           </li>
         {{/each}}
--- a/packages/frontend/app/helpers/year-next.js
+++ b/packages/frontend/app/helpers/year-next.js
@@ -1,7 +1,0 @@
-import ENV from 'frontend/config/environment';
-
-export default function yearNext(year) {
-  const yearNext = (parseInt(year) + 1).toString();
-
-  return ENV.environment != 'production' ? yearNext.substring(2) : yearNext;
-}
--- a/packages/frontend/app/helpers/year-next.js
+++ b/packages/frontend/app/helpers/year-next.js
@@ -1,0 +1,7 @@
+import ENV from 'frontend/config/environment';
+
+export default function yearNext(year) {
+  const yearNext = (parseInt(year) + 1).toString();
+
+  return ENV.environment != 'production' ? yearNext.substring(-2) : yearNext;
+}
--- a/packages/frontend/app/helpers/year-next.js
+++ b/packages/frontend/app/helpers/year-next.js
@@ -3,5 +3,5 @@ import ENV from 'frontend/config/environment';
 export default function yearNext(year) {
   const yearNext = (parseInt(year) + 1).toString();
 
-  return ENV.environment != 'production' ? yearNext.substring(-2) : yearNext;
+  return ENV.environment != 'production' ? yearNext.substring(2) : yearNext;
 }
--- a/packages/frontend/tests/integration/components/instructor-group/courses-test.js
+++ b/packages/frontend/tests/integration/components/instructor-group/courses-test.js
@@ -43,11 +43,11 @@ module('Integration | Component | instructor-group/courses', function (hooks) {
     await render(hbs`<InstructorGroup::Courses @instructorGroup={{this.instructorGroup}} />`);
     assert.strictEqual(component.title, 'Associated Courses (3)');
     assert.strictEqual(component.courses.length, 3);
-    assert.strictEqual(component.courses[0].text, 'course 0 (2013 - 14)');
+    assert.strictEqual(component.courses[0].text, 'course 0 (2013)');
     assert.strictEqual(component.courses[0].url, '/courses/1');
-    assert.strictEqual(component.courses[1].text, 'course 1 (2013 - 14)');
+    assert.strictEqual(component.courses[1].text, 'course 1 (2013)');
     assert.strictEqual(component.courses[1].url, '/courses/2');
-    assert.strictEqual(component.courses[2].text, 'course 2 (2013 - 14)');
+    assert.strictEqual(component.courses[2].text, 'course 2 (2013)');
     assert.strictEqual(component.courses[2].url, '/courses/3');
     await a11yAudit(this.element);
     assert.ok(true, 'no a11y errors found!');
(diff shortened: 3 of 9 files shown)


michaelchadwick:
OK, I updated this to use the `academicYearCrossesCalendarYearBoundaries` iliosConfig service so that the end year is only added if needed.

packages/frontend/app/helpers/year-next.js
```diff
@@ -0,0 +1,7 @@
+import ENV from 'frontend/config/environment';
+
+export default function yearNext(year) {
+  const yearNext = (parseInt(year) + 1).toString();
+
+  return ENV.environment != 'production' ? yearNext.substring(2) : yearNext;
```
jrjohnson, line 6: I'm not sure why this would change based on the environment, can you add a comment explaining?

packages/frontend/app/helpers/year-next.js
```diff
@@ -0,0 +1,7 @@
+import ENV from 'frontend/config/environment';
+
+export default function yearNext(year) {
```
jrjohnson, line 3: I like the helper creation, but looking at where this might be used I don't think it's worth the overhead since we can use a math helper in the template as easily. It's possible that a class based helper could take the course and output the year string *after* looking in the config to determine which output. That would be nice to consolidate. I'm not sure helpers can access services though. Worth investigating.

packages/frontend/tests/integration/components/instructor-group/courses-test.js
```diff
@@ -43,11 +43,11 @@ module('Integration | Component | instructor-group/courses', function (hooks) {
     await render(hbs`<InstructorGroup::Courses @instructorGroup={{this.instructorGroup}} />`);
     assert.strictEqual(component.title, 'Associated Courses (3)');
     assert.strictEqual(component.courses.length, 3);
-    assert.strictEqual(component.courses[0].text, 'course 0');
+    assert.strictEqual(component.courses[0].text, 'course 0 (2013)');
     assert.strictEqual(component.courses[0].url, '/courses/1');
-    assert.strictEqual(component.courses[1].text, 'course 1');
+    assert.strictEqual(component.courses[1].text, 'course 1 (2013)');
     assert.strictEqual(component.courses[1].url, '/courses/2');
-    assert.strictEqual(component.courses[2].text, 'course 2');
```
jrjohnson, line 50: I'd like to see a second test added that uses the alternative config and outputs the xxxx - xxxx format. Searching for the config name in tests may get you something to copy and paste from as an example for how to use mirage to setup a different config option.

packages/frontend/app/components/instructor-group/courses.hbs
```diff
@@ -11,7 +11,11 @@
           <li data-test-course>
             <LinkTo @route="course" @model={{course}}>
               <FaIcon @icon="square-up-right" />
-              {{course.title}}
+              {{#if this.academicYearCrossesCalendarYearBoundaries}}
+                {{course.title}} ({{course.year}} - {{year-next course.year}})
```
jrjohnson, line 15: I think the title can fall out of this if/else block to avoid needing to be duplicated in both sections.

packages/frontend/tests/integration/components/learner-group/root-test.js
```diff
@@ -104,8 +104,8 @@ module('Integration | Component | learner-group/root', function (hooks) {
     );
     assert.notOk(component.instructorManager.assignedInstructors[1].userNameInfo.hasAdditionalInfo);
     assert.strictEqual(component.associatedCourses.courses.length, 2);
-    assert.strictEqual(component.associatedCourses.courses[0].text, 'course 0');
-    assert.strictEqual(component.associatedCourses.courses[1].text, 'course 1');
+    assert.strictEqual(component.associatedCourses.courses[0].text, 'course 0 (2013)');
```
jrjohnson, line 107: As with instructor group this should get a test for the alternative xxxx - xxxx syntax.

jrjohnson (changes requested):
Getting close. Worth taking time to explore what we might do with a helper, but if you don't care to go down that road it's ok to use the `add` math helper we use elsewhere to increment the year in the template.

packages/frontend/app/helpers/year-next.js
michaelchadwick: This was done in response to this: https://github.com/ilios/ilios/issues/3422#issuecomment-2083248287

michaelchadwick:
@jrjohnson Removed unneeded global helper, updated all tests to check both conditions

jrjohnson (approved):
🚀

dartajax:
I would to review this one a bit more and discuss it just to make sure it conforms with all of the other ways we display the academic year to which a course is associated.

dartajax:
Original ticket comment ...

```InstructorGroups and LearnerGroups both show Associated Courses, but only the title. To fix ilios/3422, this adds the display of the course year range, starting with course.year and appending  - course.year+1 if it crosses a year boundary.```

Reality from a non-technical perspective ...
For most schools, 

```InstructorGroups and LearnerGroups both show Associated Courses, but only the title. To fix ilios/3422, this adds the display of the course year, starting with course.year and appending  - course.year+1.```

For schools that have academic year configured as a single year such as we have on demo at UCSF, 

```InstructorGroups and LearnerGroups both show Associated Courses, but only the title. To fix ilios/3422, this adds the display of the course year.```

michaelchadwick:
@dartajax Will you please link an example of a school that is configured to only show a single course year on demo?

dartajax:
Demo was set up so we could test the single year configuration we customized as a request from a school that was checking Ilios out - in Australia I believe. This is a configurable setting - I do not know how this configuration was customized. I do remember testing it though. That is why we put it on demo but never went back to our UCSF, configuration and most other schools', configuration of `course.year - course.year + 1` that appear in courses and sessions and probably in program as well - so ....

This example is on demo (or possibly at other schools who use this single year formatting).

This is a course where the instruction involved takes place in 2024 but this is a 2023 course, regardless of when the instruction takes place. This is true if the instruction started in November 2023 and continued until March 2024 - still a 2023 course. Technically, it could even be in 2022 or 2025 but I throw that in there just to demonstrate the reality of the fact that we only care about the academic year (the course year) with which this course is associated - it will always be 2023 for this course.

image

Another example still on demo ...

image

These same two courses can be found on stage or prod and the course year stuff is the same -- 2023 course is a 2023 course regardless of when the instructional activity occurs - no need to calculate "spanning".

This next screen shot comes from prod and reflects the way 90-99% of the Ilios-using world would see this scenario. The label of 2023-2024 is what we want to have appended in parentheses for these courses when the concatenated label appears in Learner Group and Instructor Group assignment labels. 

image

So to summarize, if your school is configured like our prod environment is, and like most implementations are, we want to have the label mimic (in parentheses) what is displayed in the drop-down on the course list. 

Standard (course.year - course.year + 1) configuration ...

`Renal, Endocrine, GI, Nutrition (2023 - 2024)`
`Core Inquiry Curriculum (2023 - 2024)`

Alternate configuration - (like on https://demo.iliosproject.org/ and possibly other schools who have used this) ...

`Renal, Endocrine, GI, Nutrition (2023)`
`Core Inquiry Curriculum (2023)`

Keep in mind it doesn't matter if the actual instruction takes or took place in another year. Only the `(course.year - course.year + 1)` or `(course.year)`, depending on the configuration, should be appended to the course title in the badges that appear on Instructor Group / Learner Group pages.

dartajax:
You can also see this formatting in Program Years ...

https://curriculum.ucsf.edu/programs/29

See how in prod (above) we format the years like (year - year + 1) and this matches how we describe course year in courses and sessions course list (available in selector drop-down).

Below we have the same thing on demo ...

https://demo.iliosproject.org/programs/29

This is for School Medicine BRIDGES - our current program. Our labels on the Instructor Group and Learner Group management pages should match these formats.

dartajax:
If you want screen shots instead, feel free to let me know.

michaelchadwick:
Thanks for the detail! Looks like I need to do some more research to cover all the display cases.

jrjohnson:
I'm not clear on the way the current change doesn't handle this @dartajax, I think it's working as intended, but I may be missing something. Netlify feeds off of demo so the PR preview here will use the single year config. We can update demo in two seconds though and make it easier to test the other path.

dartajax:
Well. @jrjohnson I have documented as much as I will be in all likelihood- just to add detail to what's supposed to happen here. I figured we could pull a switch and have it reflect the situation on prod with 2023-2024 instead of 2023 and we will for testing purposes do exactly that. The thing I wanted to point out was that the concept of a course "spanning" into another year was not really applicable since the only thing that matters is `course.year` and how that gets appended onto the labels for display purposes either `course.year - (course.year + 1)` or simply `course.year`. Maybe this "spanning" business was something else entirely and the naming convention used in the code freaked me out. If so, I apologize for the noise but I still believe the documentation will help us test and make sure all is well.

dartajax:
Why does GitHub not have time stamps on ticket comments by the way? I can't say "see my comment from 4 hours ago" and have that mean anything. My first comment today summarizes my concern. The title of this ticket and "if it crosses a year boundary" was freaking me out.

dartajax:
Even though I way over-documented this, I am going to re-test on Monday and probably merge it then. My initial testing actually came out fine. @jrjohnson approved it and once I double-verify all is well on the netlify demo proxy, I have confidence in the reality of production and stage as well.

dartajax:
I am testing on my local environment, proxied to demo so the course shows up with the course year appended to the title without the "spanning" of the label. It looks good to me. @jrjohnson has approved the code. I am planning on merging this as it looks good to me. I don't foresee any problems with merging this, pushing it to stage, testing there, and then going to prod when we do a new a new release. Great stuff @michaelchadwick.

dartajax (approved):
looks good to me - I click tested quite a bit

PR #7811, opened 2024-05-07
Move ICS Feed button from Ilios Calendar to Main Navigation

michaelchadwick:
Fixes ilios/ilios#4116.

The old functionality is to click on Calendar in the main navigation, click the ICS Feed button, and then click the "Copy My ICS Link" button in the flyout to copy the URL to the user's clipboard.

ilios-ics-pre

The new functionality is to click the ICS Feed button to copy the URL to the user's clipboard, directly from the main navigation, always available, and without the flyout.

ilios-ics-post

Tests for this are still being figured out, so this is a draft for now.

### DIFF
--- a/packages/ilios-common/addon/components/dashboard/navigation.hbs
+++ b/packages/ilios-common/addon/components/dashboard/navigation.hbs
@@ -35,5 +35,11 @@
         {{t "general.calendar"}}
       </LinkTo>
     </li>
+    <li>
+      <IcsFeed
+        @url={{this.icsFeedUrl}}
+        @instructions={{this.icsInstructions}}
+      />
+    </li>
   </ul>
 </nav>
--- a/packages/ilios-common/addon/components/dashboard/navigation.js
+++ b/packages/ilios-common/addon/components/dashboard/navigation.js
@@ -1,0 +1,27 @@
+import Component from '@glimmer/component';
+import { service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import { dropTask } from 'ember-concurrency';
+
+export default class NavigationComponent extends Component {
+  @service currentUser;
+  @service iliosConfig;
+
+  @tracked icsFeedUrl;
+  @tracked icsInstructions;
+
+  constructor() {
+    super(...arguments);
+    this.setup.perform();
+  }
+
+  setup = dropTask(async () => {
+    const user = await this.currentUser.getModel();
+    const icsFeedKey = user.icsFeedKey;
+    const apiHost = this.iliosConfig.apiHost;
+    const loc = window.location.protocol + '//' + window.location.hostname;
+    const server = apiHost ? apiHost : loc;
+    this.icsFeedUrl = server + '/ics/' + icsFeedKey;
+    this.icsInstructions = 'Copy My ICS Link';
+  });
+}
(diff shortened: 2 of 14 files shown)


michaelchadwick:
I wasn't able to figure out how to test the new tooltip notification upon clicking the ICS Feed button, but I removed some testing code that was no longer relevant.

jrjohnson (commented):
I think putting this here was my idea, but now that it's here I don't love it. It feels like too prominent a location for this and doesn't balance particularly well with the other buttons. Some discussion needed for sure. Maybe in the profile menu?

The tooltip copy works, but I think we have some better options. Take a look at:
1. The copy button on a virtual learning URL in an offering
2. The copy button on the API key generator at the bottom of the user profile page

I didn't look too deeply at the code as is, think we need to work out some UX first.

michaelchadwick:
The API key copy notification is nice, yes, and could be used instead. As for the placement of the icon itself, I don't have a strong feeling one way or another. It looks like the user profile menu is pretty bare and could easily incorporate more functionality. However, would users know to look there for this?

michaelchadwick:
All righty: I changed the button to have an IliosTooltip on mouse hover, and the click success triggers a flash message instead.

jrjohnson (changes requested):
Can you have the tooltip pop out to the right instead of underneath when there is space? If this ends up being more than an hour's work never mind. I'd also like to see the button switch to a check after it is clicked along with the confirmation.

michaelchadwick:
Moved the tooltip so it displays to the right. If there's no room, then it falls back to bottom.

As for the icon itself turning into a checkmark, you mean you want the copy button to go from RSS Feed (pre-click)->Checkmark (clicked)->RSS Feed (post-click)?

jrjohnson (approved):

jrjohnson:
Code looks good, assigning to @dartajax for another visual pass and merge.

dartajax:
percy won't let me merge this one either but quick assumption - it looks like the .ics feed button is snuggling up to the Calendar button - they are related but is this visual representation intentional? I am cool with it just wondering

dartajax:
we need to be aware - other support people as well as me - that when a user clicks Calendar, the screen shifts and the top menu is not viewable meaning that they may not be able to find the .ics feed button since it has been moved to the top of the screen - just pointing this out

michaelchadwick:
> ...it looks like the .ics feed button is snuggling up to the Calendar button - they are related but is this visual representation intentional? I am cool with it just wondering

@dartajax The visual representation is intentional per https://github.com/ilios/ilios/issues/4116#issuecomment-2025798087.

michaelchadwick:
> we need to be aware...that when a user clicks Calendar, the screen shifts and the top menu is not viewable meaning that they may not be able to find the .ics feed button since it has been moved to the top of the screen

@dartajax This is a good point. I'm not sure how often people use this icon link, so not sure if it matters, though. @jrjohnson and @stopfstedt, thoughts?

dartajax:
> > we need to be aware...that when a user clicks Calendar, the screen shifts and the top menu is not viewable meaning that they may not be able to find the .ics feed button since it has been moved to the top of the screen
> 
> @dartajax This is a good point. I'm not sure how often people use this icon link, so not sure if it matters, though. @jrjohnson and @stopfstedt, thoughts?

My $.02 is that most students grab their feed from here so a lot of people use it. But we are in summer right now and new students can have a new interface and updated instructions on grabbing their feeds.

jrjohnson:
I think the scrolling display is a different issue, we don't need to consider it here. Percy has now run, just needs someone to check and make sure all the changes are valid.

PR #7816, opened 2024-05-09
Fix the global search test

jrjohnson:
This test wasn't actually working, it only appeared to pass because we didn't wait for the re-render.

Found this while working on #7784, but didn't want to mess with the tests and the app code in the same place.


michaelchadwick (approved):
Pulled the change, test still passes. Checked if I changed `@page=1` to `@page=2`, and that broke the test as I assumed it might, so I think this is good to go.

PR #7819, opened 2024-05-09
Fixes course search when searching for title AND externalId

michaelchadwick:
Fixes ilios/ilios#5435 by adding a computed field that concatenates both title and externalId so that either or both will match.


jrjohnson (approved):
I had initial reservations thinking maybe we should just strip the parentheses from the input, but that wouldn't actually solve the issue. I now approve this fully!

PR #7821, opened 2024-05-10
Add icon to header to link to User Guide

michaelchadwick:
Fixes ilios/ilios#3759.

Created new `UserGuideLink` component that displays an icon to click that opens the Ilios User Guide in a new window. It has translations for the `<svg><title>` element, and a simple component test.

### DIFF
--- a/packages/frontend/app/styles/components/user-guide-link.scss
+++ b/packages/frontend/app/styles/components/user-guide-link.scss
@@ -1,7 +1,10 @@
 @use "../ilios-common/mixins" as cm;
 @use "../ilios-common/colors" as c;
+@use "../mixins" as m;
 
 .user-guide-link {
+  @include m.header-menu;
+  margin-left: 0;
   @include cm.for-tablet-and-up {
     margin-right: 0.5rem;
   }
@@ -9,6 +12,8 @@
   a {
     svg {
       color: c.$white;
+      height: 1.5em;
+      width: 1.5em;
     }
   }
 }
--- a/packages/frontend/app/styles/components/user-guide-link.scss
+++ b/packages/frontend/app/styles/components/user-guide-link.scss
@@ -1,18 +1,12 @@
 @use "../ilios-common/mixins" as cm;
 @use "../ilios-common/colors" as c;
-@use "../mixins" as m;
 
 .user-guide-link {
-  @include m.header-menu;
+  margin-left: 0.5rem;
 
   @include cm.for-tablet-and-up {
     margin-left: 0.25rem;
     margin-right: 0.5rem;
-  }
-
-  @include cm.for-phone-only {
-    /* stylelint-disable property-disallowed-list */
-    font-size: 3.5vw;
   }
 
   a {
@@ -20,8 +14,17 @@
       align-items: center;
       color: c.$white;
       display: flex;
-      height: 1.85em;
-      width: 1.85em;
+      height: 30px;
+      width: 30px;
+    }
+  }
+
+  @include cm.for-phone-only {
+    a {
+      svg {
+        height: 24px;
+        width: 24px;
+      }
     }
   }
 }
--- a/packages/frontend/app/styles/components/user-guide-link.scss
+++ b/packages/frontend/app/styles/components/user-guide-link.scss
@@ -6,7 +6,7 @@
   @include m.header-menu;
 
   @include cm.for-tablet-and-up {
-    margin-left: 0;
+    margin-left: 0.25rem;
     margin-right: 0.5rem;
   }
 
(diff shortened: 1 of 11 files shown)


dartajax:
neat - any chance of making the icon a little bit bigger? 20% or so - that's my $.02, which could get vetoed - looks good though all in all

michaelchadwick:
@dartajax No problem on the size change. I also noticed I need to adjust it for smaller screens, so I'll fix that, too.

packages/frontend/tests/integration/components/user-guide-link-test.js
```diff
@@ -0,0 +1,30 @@
...
+  setupIntl(hooks, 'en-us');
+
+  test('it renders', async function (assert) {
+    await render(hbs`<UserGuideLink />`);
+
+    assert.dom('.user-guide-link').exists();
+    assert.dom('.user-guide-link svg.fa-circle-question').exists();
+    assert.dom('.user-guide-link svg.fa-circle-question title').hasText('Ilios User Guide');
+
+    this.owner.lookup('service:intl').setLocale('es');
```
jrjohnson, line 18: You don't need to test the translations, but if you want to follow the pattern in https://ember-intl.github.io/ember-intl/docs/guide/testing to set the locale or test the untranslated output.

jrjohnson (commented):
Long time coming this one. I'm not in love with the icon and think this probably belongs in the footer instead of the header but will follow @dartajax's lead here. Otherwise just needs a small adjustment to the tests.

packages/frontend/tests/integration/components/user-guide-link-test.js
michaelchadwick: Updated tests to be more in line with convention.

michaelchadwick:
Some alternatives for icons:
* Book: https://fontawesome.com/icons/book?f=classic&s=regular
* Info: https://fontawesome.com/icons/circle-info?f=classic&s=regular
* Life Ring: https://fontawesome.com/icons/life-ring?f=classic&s=regular
* Clipboard Question: https://fontawesome.com/icons/clipboard-question?f=classic&s=regular

@dartajax What say you?

jrjohnson:
Might be the "regular" I don't love. We have a license for everything.

michaelchadwick:
Oddly, though, the `<FaIcon>` component had no hook to use `fa-regular`, only `fa-solid` and `fa-brands`, so that is why I made the change to the component in this PR.

Regardless, I ended up using the `fa-solid` version, anyway, but just changed the color of the svg to be white.

packages/frontend/app/styles/components/user-guide-link.scss
```diff
@@ -0,0 +1,27 @@
...
+  @include m.header-menu;
+
+  @include cm.for-tablet-and-up {
+    margin-left: 0.25rem;
+    margin-right: 0.5rem;
+  }
+
+  @include cm.for-phone-only {
+    /* stylelint-disable property-disallowed-list */
+    font-size: 3.5vw;
```
jrjohnson, line 15: I don't love using this disallowed property, Could this be done by controlling width and height of the icon?

packages/frontend/app/styles/components/user-guide-link.scss
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  a {
+    svg {
+      align-items: center;
+      color: c.$white;
+      display: flex;
+      height: 1.85em;
```
jrjohnson, line 23: I'm wondering if this can be done as a percentage of the container. Not sure, just feels a little arbitrary and might break if we accidentally slightly change the header height.

jrjohnson (changes requested):
Pedantic CSS changes requested. Sorry.

packages/frontend/app/styles/components/user-guide-link.scss
```diff
@@ -0,0 +1,27 @@
+@use "../ilios-common/mixins" as cm;
+@use "../ilios-common/colors" as c;
+@use "../mixins" as m;
+
+.user-guide-link {
+  @include m.header-menu;
```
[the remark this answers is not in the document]
michaelchadwick: I removed the mixin and used a simpler style to get the same effect.

packages/frontend/app/styles/components/user-guide-link.scss
michaelchadwick: I removed this disallowed property and used height/width instead.

packages/frontend/app/styles/components/user-guide-link.scss
michaelchadwick: I changed this to use pixels instead. Using percentage of the container did not seem to work.

jrjohnson (approved):

jrjohnson:
Code looks good, assigning to @dartajax for visual review and merging.

dartajax (approved):
looks good to me - percy isn't happy though

jrjohnson:
Percy just needs approval of the changes @dartajax, because this header is on every page it's a lot of changes, but still worth checking through I think.

PR #7826, opened 2024-05-14
Use display name in search results

jrjohnson:
When searching the directory for users we get display name as part of the results. We should use that whenever we have it, otherwise fallback to the existing first last pattern.

Fixes ilios/ilios#5453


michaelchadwick (approved):
LGTM!

PR #7838, opened 2024-05-20
Add missing test-support.css

michaelchadwick:
Running tests at `/tests` always throws a 404 error for `test-support.css`, which is annoying as it pollutes the dev console. This file seems to be missing even from a new EmberJS application, but the `tests/index.html` file links to it.

It _could_ be useful as a last-resort place to override things, but at the very least it kills the dumb error.

### DIFF
--- a/packages/frontend/public/assets/test-support.css
+++ b/packages/frontend/public/assets/test-support.css
@@ -1,0 +1,1 @@
+/* add styles here specifically for testing */


jrjohnson (changes requested):
There is an open issue at `https://github.com/ember-cli/ember-cli/issues/10428` about this. I'm not keen to add a blank, would rather fix this in `ember-cli` or just remove the call to this file from `packages/*/tests/index.html`.

michaelchadwick:
@jrjohnson I can see your point. Until it's fixed at the source, I've opened a new PR that removes the references instead: https://github.com/ilios/frontend/pull/7839

PR #7839, opened 2024-05-21
Fix test-support.css bug

michaelchadwick:
Undoes what https://github.com/ilios/frontend/pull/7838 was planning, and instead removes the reference to `test-support.css` in our testing `index.html` files.

### DIFF
--- a/packages/frontend/public/assets/test-support.css
+++ b/packages/frontend/public/assets/test-support.css
@@ -1,0 +1,1 @@
+/* add styles here specifically for testing */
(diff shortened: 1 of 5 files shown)


jrjohnson (changes requested):
Looks like your editor did an autocorrect number on this. I'd be fine with these changes, but ideally we'd like to stick as close to [the default blueprint](https://github.com/ember-cli/ember-new-output/blob/e0b5c6f055f42b6b6b49293987b0459cab9718f1/tests/index.html) provided by ember-cli as possible as it makes upgrading to new ember-cli versions much easier so in this case revert and just remove the one line.

michaelchadwick:
Oh yikes. I'm assuming this is a Prettier issue, where it's trying to be _helpful_ -_-

jrjohnson (approved):
Let's give it a go.

PR #7856, opened 2024-06-03
Prune non-significant years from user profile available cohorts

michaelchadwick:
Fixes ilios/ilios#5463.


packages/frontend/app/components/user-profile-cohorts-manager.js
```diff
@@ -99,8 +99,22 @@ export default class UserProfileCohortsManagerComponent extends Component {
     const programYears = (await Promise.all(programs.map((p) => p.programYears))).flat();
     const cohorts = await Promise.all(programYears.map((py) => py.cohort));
     const sortedCohorts = await sortCohorts(cohorts);
+
     return sortedCohorts.filter((cohort) => {
-      return selectedPrimaryCohort !== cohort && !selectedSecondaryCohorts.includes(cohort);
+      // filter cohorts through relevant range
+      const programYearEndYear = parseInt(cohort.programYear.content.startYear);
```
jrjohnson, line 105: It's not easily lintable (that I can find) but I'd prefer you use:
```suggestion
      const programYearEndYear = Number(cohort.programYear.content.startYear);
```
when doing an explicit type conversion to a number. `parseInt` has a gotcha around the `radix` and can turn some non-numbers like `100%` into numbers where you'd probably prefer an error.

jrjohnson: I'm confused by this variable name shouldn't it be `programYearStartYear`?

jrjohnson (changes requested):
Semantic changes needed, otherwise ✅

packages/frontend/app/components/user-profile-cohorts-manager.js
michaelchadwick: I wasn't aware of the `parseInt()` vs `Number()` difference, so I learned something!

Also, that variable name has been corrected, because you are right about its use.

jrjohnson (approved):

PR #7857, opened 2024-06-04
Learner Groups now use same loading style as Instructor Groups

michaelchadwick:
Fixes ilios/ilios#5426

Learner Groups were using the deprecated(?) `<PulseLoader>` loading component, but now they are using the same style as Instructor Groups (i.e. obfuscated shimmer method).

### DIFF
--- a/packages/frontend/app/components/learner-groups/root.hbs
+++ b/packages/frontend/app/components/learner-groups/root.hbs
@@ -122,7 +122,7 @@
           @setSortBy={{@setSortBy}}
         />
       {{else}}
-        <LearnerGroups::Loading />
+        <LearnerGroups::Loading @count={{this.countForSelectedCohort}} />
       {{/if}}
     </div>
   </div>
--- a/packages/frontend/app/components/learner-groups/root.hbs
+++ b/packages/frontend/app/components/learner-groups/root.hbs
@@ -122,7 +122,7 @@
           @setSortBy={{@setSortBy}}
         />
       {{else}}
-        <LearnerGroups::Loading @count={{this.countForSelectedCohort}} />
+        <LearnerGroups::Loading @count="2" />
       {{/if}}
     </div>
   </div>
--- a/packages/frontend/app/components/learner-groups/root.hbs
+++ b/packages/frontend/app/components/learner-groups/root.hbs
@@ -122,7 +122,7 @@
           @setSortBy={{@setSortBy}}
         />
       {{else}}
-        <LearnerGroups::Loading @count="2" />
+        <LearnerGroups::Loading @count={{2}} />
       {{/if}}
     </div>
   </div>
--- a/packages/frontend/app/components/learner-groups/root.hbs
+++ b/packages/frontend/app/components/learner-groups/root.hbs
@@ -122,7 +122,7 @@
           @setSortBy={{@setSortBy}}
         />
       {{else}}
-        <PulseLoader />
+        <LearnerGroups::Loading />
       {{/if}}
     </div>
   </div>
--- a/packages/frontend/tests/integration/components/learner-groups/loading-test.js
+++ b/packages/frontend/tests/integration/components/learner-groups/loading-test.js
@@ -10,8 +10,8 @@ module('Integration | Component | learner-groups/loading', function (hooks) {
   setupIntl(hooks, 'en-us');
 
   test('it renders', async function (assert) {
-    await render(hbs`<LearnerGroups::Loading />`);
-    assert.dom('tbody tr').exists({ count: 1 });
+    await render(hbs`<LearnerGroups::Loading @count={{2}} />`);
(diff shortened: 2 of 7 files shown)


packages/frontend/app/components/learner-groups/root.hbs
```diff
@@ -122,7 +122,7 @@
           @setSortBy={{@setSortBy}}
         />
       {{else}}
-        <PulseLoader />
+        <LearnerGroups::Loading />
```
jrjohnson, line 125: This accepts a `count` of all the learner groups that will eventually show up on the page and has a row for each one. I think you'll need to massage the data loading a bit to get the same info you need here.

jrjohnson (changes requested):

packages/frontend/app/components/learner-groups/root.hbs
michaelchadwick: I've tried a couple times to figure out how to load the count for Learner Groups, but it does it differently than Instructor Groups, as the model relationships are different. Can I get a hint? :P

michaelchadwick:
@jrjohnson Using graphql to get the LearnerGroup count works great...*except* that you need to get the `id` of the `cohort` that `belongsTo` to the `selectedProgramYear`. Class in question: https://github.com/ilios/frontend/blob/master/packages/frontend/app/components/learner-groups/root.js

My work-in-progress attempt:
```
root.hbs: <LearnerGroups::Loading @count={{this.countForSelectedCohort}} />

root.js:
learnerGroupCountData = new TrackedAsyncData(this.findLearnerGroupCount());

...

get countForSelectedCohort() {
  return this.learnerGroupCountData.isResolved ? this.learnerGroupCountData.value : null;
}

...

async findLearnerGroupCount() {
    // cohortId requires:
    //  this.programYears requires:
    //    this.programYearsData requires:
    //      this.selectedProgram?.programYears requires:
    //        this.programs requires:
    //          this.programsData requires:
    //            this.selectedSchool.programs
    //
    // // the below assignment does not finish before the LearnerGroups::Loading
    // // component loads, so no value comes through   
    // const cohortId = findById(this.programYears.slice(), this.args.programYearId);
    const cohortId = null;

    if (cohortId) {
      const filters = [`cohorts: ${cohortId}`, 'parent: null'];
      const result = await this.graphql.find('learnerGroups', filters, 'id, title');

      return result.data.learnerGroups.length;
    } else {
      return 1;
    }
  }
```

michaelchadwick:
This method has the `id` I'm looking for, but not sure how to properly extract it.

```
get cohortLoadingPromise() {
    const id = this.selectedProgramYear?.belongsTo('cohort').id();
    if (id) {
      return this.dataLoader.loadCohortLearnerGroups(id);
    }

    return undefined;
  }
```

jrjohnson:
I'd just re-use that expressions: ```
const cohortId = this.selectedProgramYear?.belongsTo('cohort').id();
```

It might be null sometimes so you'd then need to check it. It's entirely possible this will still be too slow to show a loader before the data is already done loading. I might try and create a summary of all counts for all cohorts in the app at the route level and pass that down. That may be a bad idea though, would have to try it and find out.

**Comment by michaelchadwick** [comment]
That line still depends on other stuff loading, and currently leads to an error that breaks the page. I'm not quite sure how to go forward so taking a break on this one for a bit.

**Comment by jrjohnson** [comment]
Ok. I think this one isn't solvable. I'm going to withdraw my initial objection to just having two rows by default. Let's go with that, get it tested and merged, and then maybe consider another path forward some other time. Sorry for leading you into a box canyon.

**Comment by michaelchadwick** [comment]
@jrjohnson Not sure what's up with this failing: https://github.com/ilios/frontend/actions/runs/10407150198/job/28821847906?pr=7857

File: packages/frontend/tests/integration/components/learner-groups/loading-test.js [hunk 2]
```diff
@@ -0,0 +1,17 @@
...
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import a11yAudit from 'ember-a11y-testing/test-support/audit';
+
+module('Integration | Component | learner-groups/loading', function (hooks) {
+  setupRenderingTest(hooks);
+  setupIntl(hooks, 'en-us');
+
+  test('it renders', async function (assert) {
+    await render(hbs`<LearnerGroups::Loading />`);
```
jrjohnson, line 13: Should test the `count` even though we're not using it quite yet. Hopefully we'll get it working at some point and then we should have a test for it.

jrjohnson (changes requested):

packages/frontend/tests/integration/components/learner-groups/loading-test.js
michaelchadwick: There is a test for this now.

jrjohnson (approved):

PR #7862, opened 2024-06-05
Use ember test setup to setup intl

jrjohnson:
This is the currently recommended way to setup things we need in all tests. I've configured it for rendering tests (acceptance and integration) and left the few calls that are in unit tests alone. Many of our older tests didn't call their setup methods from the right place so I fixed those imports as well.

Refs #7860


michaelchadwick (approved):
I just cross-referenced this with my tutorial app and this appears to be the modern way to do it, so yuuuuuuup.

PR #7867, opened 2024-06-07
removed 'deprecate-non-strict-id' rule

michaelchadwick:
Removed this rule, ran both frontend and test-app test suite, with all passing. I don't know if anything else needs to be checked.


stopfstedt (approved):
nice find.

PR #7871, opened 2024-06-10
updated config/environment to allow for cli arguments

michaelchadwick:
There are useful logging flags that can be used when needed for local development. This change allows them to be overridden by command line arguments so the `config/environment.js` file doesn't need to be changed (beyond this change).

### DIFF
--- a/packages/frontend/config/environment.js
+++ b/packages/frontend/config/environment.js
@@ -79,11 +79,11 @@ module.exports = function (environment) {
   };
 
   if (environment === 'development') {
-    ENV.APP.LOG_RESOLVER = process.env.LOG_RESOLVER || false;
-    ENV.APP.LOG_ACTIVE_GENERATION = process.env.LOG_ACTIVE_GENERATION || false;
-    ENV.APP.LOG_TRANSITIONS = process.env.LOG_TRANSITIONS || false;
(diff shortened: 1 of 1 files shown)


michaelchadwick:
Codeclimate says to "Consider simplifying this complex logical expression", but I'm not sure how I would.

stopfstedt (approved):
LGTM

packages/frontend/config/environment.js
```diff
@@ -79,11 +79,11 @@ module.exports = function (environment) {
   };
 
   if (environment === 'development') {
-    // ENV.APP.LOG_RESOLVER = true;
-    // ENV.APP.LOG_ACTIVE_GENERATION = true;
-    // ENV.APP.LOG_TRANSITIONS = true;
-    // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
-    // ENV.APP.LOG_VIEW_LOOKUPS = true;
+    ENV.APP.LOG_RESOLVER = process.env.LOG_RESOLVER || false;
```
stopfstedt, line 82: since code climate is being a jerk about it, could you try the following? should be functionally equivalent.

```suggestion
    ENV.APP.LOG_RESOLVER = !!process.env.LOG_RESOLVER;
```

stopfstedt (changes requested):
pls see comment above, and convert the fallback to FALSE into double negation. :crossed_fingers: this will appease CodeClimate.

PR #7886, opened 2024-06-14
use setLocale() test helper instead of interacting with the intl service directly.

stopfstedt:
see https://ember-intl.github.io/ember-intl/docs/test-helpers/set-locale

### DIFF
--- a/packages/test-app/tests/integration/components/session-offerings-time-block-offerings-test.js
+++ b/packages/test-app/tests/integration/components/session-offerings-time-block-offerings-test.js
@@ -10,7 +10,6 @@ module('Integration | Component | session-offerings-time-block-offerings', funct
 
   test('it renders', async function (assert) {
     const store = this.owner.lookup('service:store');
-    this.owner.lookup('service:intl').setLocale('en-us');
     const key = '2023005134520240120430'; // value doesn't really matter, although it needs to be parseable
     const offeringTimeBlock = new OfferingTimeBlock(key);
     const course = store.createRecord('course');
--- a/packages/test-app/tests/unit/utils/calendar-event-tooltip-test.js
+++ b/packages/test-app/tests/unit/utils/calendar-event-tooltip-test.js
@@ -2,13 +2,14 @@ import calendarEventTooltip from 'test-app/utils/calendar-event-tooltip';
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 import { DateTime } from 'luxon';
+import { setupIntl } from 'ember-intl/test-support';
 
 module('Unit | Utility | calendar-event-tooltip', function (hooks) {
   setupTest(hooks);
+  setupIntl(hooks, 'en-us');
 
   hooks.beforeEach(function () {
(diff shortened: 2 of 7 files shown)


michaelchadwick:
@stopfstedt How does this unit test work _without_ the `await` in front of `setLocale()`? https://github.com/ilios/frontend/blob/f786213d61ed597c1e80e1c29b572458dc6cee58/packages/test-app/tests/unit/services/locale-days-test.js

stopfstedt:
> @stopfstedt How does this unit test work _without_ the `await` in front of `setLocale()`? https://github.com/ilios/frontend/blob/f786213d61ed597c1e80e1c29b572458dc6cee58/packages/test-app/tests/unit/services/locale-days-test.js

that shouldn't be, the docs for [`setLocale()`](https://ember-intl.github.io/ember-intl/docs/test-helpers/set-locale) clearly show `await` in the example.

will make another pass over the codebase, and add `async/await` where currently missing from `setLocale()`.

packages/test-app/tests/integration/components/session-offerings-time-block-offerings-test.js
michaelchadwick, line 1: Was this just an errant call? Like, there was no need to set the locale?

packages/test-app/tests/unit/utils/calendar-event-tooltip-test.js
```diff
@@ -2,13 +2,14 @@ import calendarEventTooltip from 'test-app/utils/calendar-event-tooltip';
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 import { DateTime } from 'luxon';
+import { setupIntl } from 'ember-intl/test-support';
 
 module('Unit | Utility | calendar-event-tooltip', function (hooks) {
   setupTest(hooks);
+  setupIntl(hooks, 'en-us');
```
michaelchadwick, line 9: Isn't this unneeded? I thought the calls in https://github.com/ilios/frontend/blob/d392857d2cd5e16d075f85a95ba9b47a520ad101/packages/test-app/tests/helpers/index.js were enough?

packages/test-app/tests/integration/components/session-offerings-time-block-offerings-test.js
stopfstedt: no need to set the locale here explicitly to en-us - that service gets initialized with that locale further upstream. might have been necessary some time ago, but now it's just kruft.

packages/test-app/tests/unit/utils/calendar-event-tooltip-test.js
stopfstedt: we're only setting up intl for acceptance and integration tests, not for unit tests. since we only need it there in rare circumstances, like this one.

michaelchadwick: OK, yes. I see that we only run it for two of the three `setup` functions. Thanks!

michaelchadwick (approved):
A-OK

PR #7901, opened 2024-06-20
Update ember data to 5.3.6

jrjohnson:
We shouldn't have ignored this with dependabot, updating now and removing the ignore instruction so we get future updates as well.


stopfstedt:
superseded by https://github.com/ilios/frontend/pull/7901

PR #7905, opened 2024-06-24
Fix datepicker month/year vertical alignment

michaelchadwick:
Fixed ilios/ilios#5541


stopfstedt (approved):
code reviewed, and visually reviewed in current FF, Chromium, and Edge on Linux.

LGTM across the board now.

PR #7911, opened 2024-06-26
Declare pnpm in package

jrjohnson:
We need to do this in order to work around an issue with dependabot that is using an old pnpm version. The downside is we'll need to keep it up to date. My IDE is telling me this format isn't valid and it needs to be x.y.z format, but let's try it with just the major version and see if that solves our issues as we won't need to keep it up to date this way.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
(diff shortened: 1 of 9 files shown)


stopfstedt (approved):

jrjohnson:
Hold off on merge, may be fixed at dependabot

jrjohnson:
Our dependabot issues seem to have been fixed upstream. Closing in favor of that.

PR #7914, opened 2024-06-28
Replace Ember Data Packages

jrjohnson:
The future of Ember Data is in smaller packages, currently this is done by providing the `ember-data` meta package which includes the others as well as configuring a store. I've removed the middle man and relied directly on the individual packages as well as configuring our store to match the default one. Doing things this way resolves issues we've been having with peer dependencies, ember data, mirage, and ember-file upload.

Because we're jumping from the meta package forward to this point I also had to clear deprecations for providing transforms and inflector rules as well as configure the build in a new way.

I wasn't able to get around `pnpm` issues by installing these packages only in common, I tried all the combinations of dev/dep/peer I could think of, still got errors, instead I needed to install the data packages in each app. I think this is fine, we should get all the updates as a group from dependabot anyway.


stopfstedt (changes requested):

stopfstedt (approved):
LGTM

dartajax:
I'd love to merge this but the conflicting file - `pnpm-lock.yaml` is preventing me from doing that.

stopfstedt:
> I'd love to merge this but the conflicting file - `pnpm-lock.yaml` is preventing me from doing that.

@dartajax - merge conflict resolved. pls merge once tests have passed again.

stopfstedt:
supersedes https://github.com/ilios/frontend/pull/7901

PR #7915, opened 2024-06-29
fix visual layer display bug when editing session offering

michaelchadwick:
Fixes ilios/ilios#5560

Added a tracked value with an action that gets sent down the component pipe from `<CoursesSessions>` to `<Session-Grid-Offering>`. Toggling the pencil icon now uses that action to set the value and update `position: sticky` or not appropriately.


michaelchadwick:
I realize this is a sizable code change, and I'm not sure the variable names are as conventional as they could be, but the functionality is there.

stopfstedt (approved):

PR #7918, opened 2024-07-02
rm obsolete ResolveAllValues resource.

stopfstedt:


michaelchadwick:
What did this deleted file do?

stopfstedt:
> @stopfstedt What did this deleted file do?

@michaelchadwick  this file provided a helper for resolving a list of Promises, using something called a `Resource`. we're in the process of phasing out all remaining custom Resources that we created in favor of `TrackedAsyncData` since Resources turned out to be an evolutionary dead end.

This one can be removed without further refactoring since it is currently not in use anywhere.


For more on Resources, please see https://github.com/tracked-tools/ember-could-get-used-to-this?tab=readme-ov-file#resources

michaelchadwick (approved):
A-OK

PR #7919, opened 2024-07-02
created "on-insert" custom modifier; updated LocaleChooser

michaelchadwick:
Refs ilios/ilios#5374

In order to fix the deprecated `{{did-insert}}` modifier, I created an `{{on-insert}}` custom modifier that just takes a callback, runs if valid, or performs a `console.error` if not.

I tested this on the `<LocaleChooser>` component, and it seems to work as expected. Can probably reuse this in other spots where we are doing a simple "did this thing load? if so, do this other thing" check.


michaelchadwick:
Not sure if this was the intended "fix" for this problem, but I learned a bit, regardless.

stopfstedt:
superseded by https://github.com/ilios/frontend/pull/7922

PR #7922, opened 2024-07-02
replace {{did-insert}} modifiers in LocaleChooser component

stopfstedt:
refs ilios/ilios#5374

an alternative solution to https://github.com/ilios/frontend/pull/7919


michaelchadwick:
I nominate your alternative solution after our discussion.

jrjohnson:
Oh this is nice, I was originally thinking we'd have to go with @michaelchadwick's solution in a bunch of places where we use `did-insert` because it's actually hooked up to the DOM so it makes sense. This works very well though, especially the focus on `index` like that a lot.

packages/frontend/.lint-todo
```diff
@@ -98,3 +98,5 @@ add|ember-template-lint|no-at-ember-render-modifiers|10|6|10|6|d919d2af254f782c0
 add|ember-template-lint|no-at-ember-render-modifiers|11|6|11|6|940005188b476a060b0e5d3f05baea24ba178878|1719360000000|1734915600000|1750464000000|app/components/reports/subject/new/session-type.hbs
 add|ember-template-lint|no-at-ember-render-modifiers|10|6|10|6|d919d2af254f782c01fe2ba15416673e52e91124|1719360000000|1734915600000|1750464000000|app/components/reports/subject/new/term.hbs
 add|ember-template-lint|no-at-ember-render-modifiers|11|6|11|6|940005188b476a060b0e5d3f05baea24ba178878|1719360000000|1734915600000|1750464000000|app/components/reports/subject/new/term.hbs
+remove|ember-template-lint|no-at-ember-render-modifiers|16|4|16|4|46d995f60904f6b6f4b1b6b24bb65b9c97441bc1|1719360000000|1734915600000|1750464000000|app/components/locale-chooser.hbs
```
jrjohnson, line 101: If you run `pnpm run lint:hbs --compact-todo` it will clean this up. Otherwise I think it creates a double entry instead of removing the original.

packages/frontend/tests/integration/components/locale-chooser-test.js
```diff
@@ -57,8 +57,19 @@ module('Integration | Component | locale-chooser', function (hooks) {
     await render(hbs`<LocaleChooser />`);
 
     await component.toggle.click();
+    assert.notOk(component.toggle.hasFocus);
     await component.locales[1].click();
     assert.strictEqual(component.locales.length, 0);
     assert.strictEqual(component.text, 'Español (es)');
+    assert.ok(component.toggle.hasFocus);
```
jrjohnson, line 64: 💯

jrjohnson (changes requested):

packages/frontend/.lint-todo
stopfstedt: :heavy_check_mark:

jrjohnson (approved):

PR #7925, opened 2024-07-03
fixed ilios-error/style.css file path

michaelchadwick:
Refs ilios/ilios#5566

The stylesheet that gets added in `packages/*/lib/ilios-error/index.js` had a incorrect path, causing a console error. I fixed it so it will now load.


stopfstedt (approved):
LGTM

PR #7927, opened 2024-07-03
improve default selectedYear choice for Course Rollover

michaelchadwick:
Fixes ilios/ilios#4330

* Change first year of select options to be `currentYear` instead of `lastYear`
* Try to find next valid year (in range `currentYear` + 5 (e.g. 2024 - 2029)) by checking if each year is not found in `unavailableYears()`
* Default to `currentYear` if valid year not found


stopfstedt (approved):
LGTM

PR #7934, opened 2024-07-05
Remove render modifiers in school components

jrjohnson:
I did these all at once as I was working on some data loading at the same time, just going to push this PR first though as it's big enough already. Each component in its own commit, but still a big change. Sorry.

### DIFF
--- a/packages/frontend/app/components/school-competencies-expanded.js
+++ b/packages/frontend/app/components/school-competencies-expanded.js
@@ -1,27 +1,27 @@
 import Component from '@glimmer/component';
 import { service } from '@ember/service';
 import { all } from 'rsvp';
-import { tracked } from '@glimmer/tracking';
+import { cached, tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
-import { dropTask, restartableTask } from 'ember-concurrency';
+import { dropTask } from 'ember-concurrency';
 import { filterBy, uniqueValues } from 'ilios-common/utils/array-helpers';
+import { TrackedAsyncData } from 'ember-async-data';
 
 export default class SchoolCompetenciesExpandedComponent extends Component {
   @service store;
   @tracked competenciesToAdd = [];
   @tracked competenciesToRemove = [];
-  @tracked schoolCompetencies;
 
-  load = restartableTask(async () => {
-    this.cleanup();
-    this.schoolCompetencies = await this.args.school.competencies;
-  });
+  @cached
+  get schoolCompetenciesData() {
+    return new TrackedAsyncData(this.args.school.competencies);
+  }
 
   get competencies() {
-    if (!this.schoolCompetencies) {
+    if (!this.schoolCompetenciesData.isResolved) {
       return [];
     }
-    const arr = [...this.schoolCompetencies.slice(), ...this.competenciesToAdd];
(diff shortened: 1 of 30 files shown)


packages/frontend/app/components/school-competencies-expanded.js
```diff
@@ -1,27 +1,27 @@
...
-import { dropTask, restartableTask } from 'ember-concurrency';
+import { dropTask } from 'ember-concurrency';
 import { filterBy, uniqueValues } from 'ilios-common/utils/array-helpers';
+import { TrackedAsyncData } from 'ember-async-data';
 
 export default class SchoolCompetenciesExpandedComponent extends Component {
   @service store;
   @tracked competenciesToAdd = [];
   @tracked competenciesToRemove = [];
-  @tracked schoolCompetencies;
```
stopfstedt, line 13: needs to be replaced with a getter that wraps around `get schoolCompetenciesData()`.

stopfstedt (commented):
line:  `this.schoolCompetencies = await this.args.school.competencies;` in `SchoolCompetenciesExpandedComponent::save()` should be removed.

packages/frontend/app/components/school-vocabularies-expanded.js
```diff
@@ -1,14 +1,13 @@
...
 import { service } from '@ember/service';
-import { dropTask, restartableTask } from 'ember-concurrency';
-import { map } from 'rsvp';
+import { dropTask } from 'ember-concurrency';
 import { findById } from 'ilios-common/utils/array-helpers';
+import { TrackedAsyncData } from 'ember-async-data';
 
 export default class SchoolVocabulariesExpandedComponent extends Component {
   @service store;
-  @tracked schoolVocabularies = [];
```
stopfstedt, line 11: something broke in this component refactor. when adding and saving a new vocabulary, i get this error coming out of this component in the browser console `Uncaught (in promise) Error: Assertion Failed: Expected id to be a string or number, received null`

packages/frontend/app/components/school-session-type-form.hbs
```diff
@@ -2,13 +2,10 @@
   <div
     class="school-session-type-form"
     data-test-school-session-type-form
-    {{did-insert (perform this.load)}}
```
stopfstedt, line 5: something broken with this refactor. clicking the "save" button on this form does nothing now now - regardless of updating or adding an item.

stopfstedt (changes requested):
several components turned out to be broken during click-testing. might be a good opportunity to add acceptance tests while at, we are very light on it right now.

jrjohnson:
Thanks for saving me from myself! Clearing was overconfident in our test coverage. I'll add a bunch of tests.

packages/frontend/app/components/school-vocabularies-expanded.js
jrjohnson: ✅

packages/frontend/app/components/school-manager.js
```diff
@@ -11,8 +11,8 @@ export default class SchoolManagerComponent extends Component {
   @service flashMessages;
   @tracked @NotBlank() @Length(1, 60) title;
 
-  @action
-  load() {
+  constructor() {
```
stopfstedt, line 14: :D

PR #7935, opened 2024-07-05
add query param for chosen school on programs route

michaelchadwick:
Fixes ilios/ilios#3457

### DIFF
--- a/packages/frontend/app/components/programs/root.js
+++ b/packages/frontend/app/components/programs/root.js
@@ -4,7 +4,6 @@ import { cached, tracked } from '@glimmer/tracking';
 import { TrackedAsyncData } from 'ember-async-data';
 import { findById } from 'ilios-common/utils/array-helpers';
 import { dropTask } from 'ember-concurrency';
-import { action } from '@ember/object';
 
 export default class ProgramRootComponent extends Component {
   @service currentUser;
@@ -55,9 +54,4 @@ export default class ProgramRootComponent extends Component {
     this.newProgram = await newProgram.save();
     this.showNewProgramForm = false;
   });
-
-  @action
-  setSchoolId(schoolId) {
-    this.args.setSchoolId(schoolId);
-  }
 }
--- a/packages/frontend/app/components/programs/root.js
+++ b/packages/frontend/app/components/programs/root.js
@@ -4,11 +4,11 @@ import { cached, tracked } from '@glimmer/tracking';
 import { TrackedAsyncData } from 'ember-async-data';
 import { findById } from 'ilios-common/utils/array-helpers';
 import { dropTask } from 'ember-concurrency';
+import { action } from '@ember/object';
 
 export default class ProgramRootComponent extends Component {
   @service currentUser;
   @service permissionChecker;
-  @tracked selectedSchoolId;
   @tracked showNewProgramForm = false;
   @tracked newProgram;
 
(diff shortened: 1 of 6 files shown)


packages/frontend/app/components/programs/root.js
```diff
@@ -41,8 +40,8 @@ export default class ProgramRootComponent extends Component {
   }
 
   get bestSelectedSchool() {
-    if (this.selectedSchoolId) {
-      return findById(this.args.schools.slice(), this.selectedSchoolId);
```
jrjohnson, line 45: This slice used to be needed as ember-data returned a thing that wasn't quite an array, however those days are gone and this can go away too and should do the same thing. Unless I'm wrong. Which isn't impossible.

jrjohnson (changes requested):
Needs an acceptance test for the original bug where you do exactly the steps in the report. I'd add that into `packages/frontend/tests/acceptance/programs-test.js`.

michaelchadwick:
> Needs an acceptance test for the original bug where you do exactly the steps in the report. I'd add that into `packages/frontend/tests/acceptance/programs-test.js`.

This seems like a great idea. Just as an aside, I don't see any kind of test for this kind of thing in the main [Courses](https://github.com/ilios/frontend/blob/3d6adfbf1428c20c383222bf5b3cea541def8c51/packages/frontend/tests/acceptance/courses-test.js) test file, but maybe it's located somewhere else.

packages/frontend/tests/acceptance/programs-test.js
```diff
@@ -88,6 +87,35 @@ module('Acceptance | Programs', function (hooks) {
...
+      assert.strictEqual(currentURL(), '/programs');
+      assert.strictEqual(page.root.schoolFilter.schools.length, 2);
+
+      assert.strictEqual(page.root.schoolFilter.selectedSchool, '1');
+
+      await page.root.schoolFilter.select(2);
+      assert.strictEqual(page.root.schoolFilter.selectedSchool, '2');
+      assert.strictEqual(currentURL(), '/programs?school=2');
+
+      await click('.list tbody tr:nth-of-type(1) td:nth-of-type(1) a');
```
jrjohnson, line 111: This should also get done in the page object, but you'll have to add a new one. It looks like it belongs in `packages/frontend/tests/pages/components/programs/list-item.js` probably something like:
`open: clickable('[data-test-link]'),`

and then you can use it here with `await page.root.list.listItem.open()` (I free handed this, copy with caution!)

packages/frontend/tests/acceptance/programs-test.js
```diff
@@ -88,6 +87,35 @@ module('Acceptance | Programs', function (hooks) {
...
+      assert.strictEqual(currentURL(), '/programs/2');
+      await click('.backtolink a');
```
jrjohnson, line 113: This is is going to be a page object too, you'll have to import the one for the program and probably add this link there in the tree.

jrjohnson (changes requested):

michaelchadwick:
@jrjohnson Changes requested exist in the same file outside of the PR, so will follow up with a PR after this that fixes that for all of them.

packages/frontend/tests/acceptance/programs-test.js
michaelchadwick: Fixed in new PR https://github.com/ilios/frontend/pull/7942

packages/frontend/tests/acceptance/programs-test.js
michaelchadwick: Fixed in new PR https://github.com/ilios/frontend/pull/7942

PR #7937, opened 2024-07-05
move course-* components into course directory and ditch the prefix.

stopfstedt:
does the same with course visualisation components.


packages/ilios-common/app/styles/ilios-common/components/course/visualizations.scss
```diff
@@ -0,0 +1,16 @@
+@use "../../mixins" as m;
+
+.course-visualizations {
```
jrjohnson, line 3: Is this a new consolidation? I couldn't find the matching file in the list, but make it's buried in deletes from other places?

packages/test-app/tests/integration/components/course/header-test.js
```diff
@@ -3,10 +3,10 @@ import { setupRenderingTest } from 'test-app/tests/helpers';
 import { render } from '@ember/test-helpers';
 import { hbs } from 'ember-cli-htmlbars';
 import { a11yAudit } from 'ember-a11y-testing/test-support';
-import { component } from 'ilios-common/page-objects/components/course-header';
+import { component } from 'ilios-common/page-objects/components/course/header';
 import { setupMirage } from 'ember-cli-mirage/test-support';
 
-module('Integration | Component | course-header', function (hooks) {
+module('Integration | Component | course/header', function (hooks) {
```
jrjohnson, line 9: Now you're just showing off! (would never have occurred to me to fix this too 😆)

jrjohnson (approved):

packages/ilios-common/app/styles/ilios-common/components/course/visualizations.scss
stopfstedt: yes. was `packages/ilios-common/app/styles/ilios-common/components/course-visualizations.scss` before.

packages/ilios-common/addon/components/course/details.js
```diff
@@ -3,7 +3,7 @@ import { action } from '@ember/object';
 import scrollIntoView from 'scroll-into-view';
 import { service } from '@ember/service';
 
-export default class IliosCourseDetailsComponent extends Component {
```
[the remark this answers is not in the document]
stopfstedt: right!? we have less than a  handful `Ilios*` components left. can't remember either what compelled us to use this prefix in the first place.

PR #7940, opened 2024-07-08
adds test coverage for all routes where missing

stopfstedt:
follow-up ticket to #7926


jrjohnson (approved):
We'd mostly been omitting these, but I guess that was a mistake, thanks for re-adding.

PR #7942, opened 2024-07-09
programs acceptance test now only uses page objects

michaelchadwick:
Refs ilios/frontend#7935

The acceptance test for Programs still used direct selector access (e.g. `.list tbody`) in some places instead of page objects. I've updated this.


stopfstedt (approved):
LGTM

PR #7943, opened 2024-07-09
Move percy back into CI

jrjohnson:
We've been playing some tricks to lower our screenshot count and not expose the percy token, however it's not working as we rarely remember to add this label. I'm tossing all the previous ideas and going back to running percy in CI.

I did add a check so that it won't run on draft pull requests, but will run everywhere else.


jrjohnson:
Tests didn't re-run when I took this out of draft mode, gonna see if there is a way to trigger that.

stopfstedt (approved):

PR #7947, opened 2024-07-10
Add acceptance tests for schools

jrjohnson:
This area of the app was undercovered by acceptance tests which was discovered in #7934. I've added the tests here first so we can validate them before using them in that PR. I had to `todo` one test which will be fixed in that PR so I didn't dig too deep into figuring out why it doesn't work currently.


stopfstedt (approved):
niice.

PR #7955, opened 2024-07-10
fixes moveTarget() and clearFocus() actions.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/5578


packages/frontend/app/components/locale-chooser.js
```diff
@@ -67,8 +66,8 @@ export default class LocaleChooserComponent extends Component {
     }
   }
   @action
-  clearFocus() {
-    const buttons = this.menuElement.querySelectorAll('button');
+  clearFocus(event) {
+    const buttons = event.target.parentElement.children;
```
michaelchadwick, line 70: This still gives me an error:
Screenshot 2024-07-10 at 2 36 17 PM

michaelchadwick (changes requested):
`clearFocus()` change still gives an error:
Screenshot 2024-07-10 at 2 36 17 PM

stopfstedt:
> `clearFocus()` change still gives an error

yes. turns out that `HtmlCollection` does not have a `forEach()` method, since it's not an array.  So I had to use a plain old  `for` loop instead, boring.

anyway - _now_ it's ready for review.

PR #7958, opened 2024-07-12
Remove Launcher from Browserstack Job

jrjohnson:
Looking at the dashboard I think each run may be testing in every browser, going to try switching this up and see what gets reported.


jrjohnson:
No, just one at a time.

PR #7960, opened 2024-07-12
Import Controller for Tests

jrjohnson:
This is needed in our embroider build when these tests run before an application has been booted anywhere else.


michaelchadwick:
I'm assuming these changes need to be made because of the `packages/*` monorepo-ness of Ilios? Or is because of the LTI stuff?

My tutorial app [controller tests](https://github.com/michaelchadwick/ember-remember-stuff/blob/8e320377819281bd34abcd737e9f7e7d44e4fe76/tests/unit/controllers/application-test.js) don't need to import the controller to work, even when built with Embroider.

jrjohnson:
It's due to building with the not-quite-stable-yet embroider build with `splitAtRoutes` enabled. That configuration bundles the parts of the app into individual JS files which are then loaded when needed. This is great for users who don't have to download the entire app just to see their calendar, but has some challenges as well when something (like these controllers) is automatically loaded into the application on boot.

stopfstedt (approved):
LGTM

PR #7962, opened 2024-07-12
Dasherize Model Names

jrjohnson:
This clears the deprecate-non-strict-types ember-data deprecation. My find a replace got a little bit carried away and I ended up updating the mirage declarations as well, but I don't mind the consistency so I left them in.


stopfstedt (approved):
LGTM

PR #7963, opened 2024-07-12
Enable Testing with Ember Exam

jrjohnson:
Splits the tests into chunks and launches a bunch of browsers to run
them. I've defaulted this to 8 for running tests on a local machine, but
only 3 for github where there are less cores available.


packages/test-app/tests/integration/components/learnergroup-selection-cohort-manager-test.js
```diff
@@ -74,7 +74,7 @@ module('Integration | Component | learnergroup-selection-cohort-manager', functi
     assert.expect(2);
     this.set('filter', '');
     this.set('remove', (learnerGroup) => {
-      assert.ok(this.secondLevelLearnerGroup1, learnerGroup);
+      assert.deepEqual(this.secondLevelLearnerGroup1, learnerGroup);
```
stopfstedt, line 77: nice.

stopfstedt (approved):

PR #7969, opened 2024-07-15
Replace Mirage Implementation

jrjohnson:
Replace ember-cli-mirage with ember-mirage. This new, much more embroider compatible, addon is the way forward for mirage support.

Last commit is the annoying one as it changes all the import paths.

*[EDIT: ST 2024/07/16]* for reference: https://github.com/bgantzler/ember-mirage/blob/main/docs/migration.md


stopfstedt (approved):
LGTM

PR #7973, opened 2024-07-17
Improve Data Loading Within the School Manager

jrjohnson:
By pre-loading more of this data in the route we can do less loading and spinning in the components and significantly reduce the number of network calls required to manage a school.

Followup to #7934


stopfstedt (approved):
LGTM

PR #7984, opened 2024-07-18
Split UI Tests

jrjohnson:
Creating a new UI only workflow that has the browser tests in it. This will allow us to get faster results from CI runs in general as well as limit the number of times we hit the expensive percy and automate browserstack services.

Now our UI tests will only run once a PR if our of draft status and has at least one approved review.


zorgbort (dismissed):

zorgbort (dismissed):

stopfstedt (approved):
LGTM. let's give it a try.

PR #7995, opened 2024-07-23
Remove Matrix Info From Test Names

jrjohnson:
These are nice sometimes, but they don't play well with github's UI and branch protection setup. Let's use plain names and see how that feels.


zorgbort (approved):

stopfstedt (approved):
Zorgbort approved already, I concur.

PR #7996, opened 2024-07-24
Fix Browserstack Test Name

jrjohnson:
Sometimes this job runs from a context that ember-cli-browserstack isn't
ready to understand. We can build this id ourselves and save some pain
when tests refuse to run.


zorgbort (approved):
Not really approved, just testing

PR #7997, opened 2024-07-24
add competency domain info to course objective parent objective selector component

stopfstedt:
image

fixes https://github.com/ilios/ilios/issues/5608

### DIFF
--- a/packages/ilios-common/addon/components/course/manage-objective-parents.hbs
+++ b/packages/ilios-common/addon/components/course/manage-objective-parents.hbs
@@ -34,8 +34,14 @@
           >
             <h4
               class="competency-title"
+              data-test-competency-title
             >
               {{competency.title}}
+              {{#if competency.parent}}
+                <span class="domain-title">
+                ({{competency.parent.title}})
+                </span>
+              {{/if}}
             </h4>
             <ul>
               {{#each (sort-by "title" competency.objectives) as |objective|}}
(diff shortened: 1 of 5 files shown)


packages/ilios-common/addon/components/course/manage-objective-parents.hbs
```diff
@@ -34,8 +34,14 @@
           >
             <h4
               class="competency-title"
+              data-test-competency-title
             >
               {{competency.title}}
+              {{#if competency.parent}}
```
jrjohnson, line 40: Competencies can be arbitrarily deep, is this the parent or the domain? If the domain that will need to be fetched with `getDomain` in case it isn't the direct parent.

stopfstedt: domain/competencies only go one level deep, per business rules. 
the database schema technically supports arbitrary nesting levels, but the application code won't support it.
if there's a parent, then it must be a top level domain.

jrjohnson: ✅

jrjohnson (approved):

stopfstedt:
^ @dartajax this component might need a glamour shot for the user guide.

dartajax:
I approve of this PR - it would be my honor to merge it.

PR #7998, opened 2024-07-24
improve User Guide icon

michaelchadwick:
Fixes ilios/ilios#5609

My attempt to address all the issues:
* Color scheme made consistent
* Icon shape made consistent
* Question mark uses the same base font as all other Fontawesome icons, so...maybe not fixable?
* Large screen icon alignment fixed

To do this, I had to go from using an `<a>` (which I chose since the icon takes you somewhere instead of just doing something) to using a `<button>` (which all the other elements use, giving it a lot of the same existing CSS properties) and adding a component class to model the "click this to open a new window/tab" action

### DIFF
--- a/packages/frontend/app/styles/components/ilios-header.scss
+++ b/packages/frontend/app/styles/components/ilios-header.scss
@@ -37,11 +37,11 @@
       grid-area: user;
     }
     .user-guide-link {
+      display: flex;
       grid-area: guide;
-      height: 72%;
 
       @include m.for-phone-only {
-        height: 100%;
+        vertical-align: -0.25em;
       }
     }
 
--- a/packages/frontend/app/styles/components/ilios-header.scss
+++ b/packages/frontend/app/styles/components/ilios-header.scss
@@ -38,10 +38,10 @@
     }
     .user-guide-link {
       grid-area: guide;
-      height: 66%;
+      height: 72%;
 
       @include m.for-phone-only {
-        height: 90%;
+        height: 100%;
       }
     }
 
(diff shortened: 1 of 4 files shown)


jrjohnson (commented):
I agree this should be a link. I'd rather keep the semantically correct HTML, but if that isn't possible I'll accept it. Take a look at the `ilios-link-button` mixin though, it may help.

michaelchadwick:
I'll take another crack at this to keep the semantic `<a>`-ness of it, but I wanted to get it out there for others to look at for visual and functional reasons.

michaelchadwick:
Changed back from `<button>` to `<a>` so semantic again. Had to fudge some CSS here and there, but I think it is both a bit smaller and a more distant from the other icons.

jrjohnson (approved):
Excellent. Looks great to me.

saschaben:
please if we are able: align center vertically, rather than align-bottom viz the other icons; add 2px padding on right, since when vertical scroll appears, it overlays the right edge of the icon.

otherwise, all A-OK

michaelchadwick:
@saschaben Thanks for the feedback. Will hammer it out some more soon.

michaelchadwick:
@saschaben Made some adjustments. Let me know if this works better!

packages/frontend/app/styles/components/ilios-header.scss
```diff
@@ -38,7 +38,11 @@
     }
     .user-guide-link {
       grid-area: guide;
-      height: 100%;
+      height: 72%;
```
jrjohnson, line 41: Is this a magic number that looks good, or a calculation based on something else? I'm ok with either, but just wanted to understand where it was coming from. Might need a comment to that effect.

michaelchadwick: I switched this out for a flexbox implementation instead. Thanks for the nudge ;P

jrjohnson (approved):
Excellent

PR #7999, opened 2024-07-24
actually save the change.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/5611

as an aside - i seem to be unable to provide proper test coverage for this. tried to handle this in an acceptance test - make the change, "reload" the page by visiting it again, then check if the made changes are sticking. seems straightforward, but i can't get the page object in the acceptance to "reset" on consecutive calls to `page.visit()` - it's seems to be holding on to its current state, so it's actually not reloading/resetting.  any ideas on how to get this to work?

### DIFF
--- a/packages/frontend/app/components/competency-title-editor.js
+++ b/packages/frontend/app/components/competency-title-editor.js
@@ -25,6 +25,7 @@ export default class CompetencyTitleEditorComponent extends Component {
       return false;
     }
     this.args.competency.set('title', this.title);
+    await this.args.competency.save();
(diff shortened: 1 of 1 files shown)


stopfstedt:
upon further inspection, this is the right solution to the wrong problem. turns out that changes are not supposed to be persisted when the in-place editor is closed, but when the competency manager is closed via its "done" button.

this works as intended.

dartajax:
I think I fell for this the other day too. 2 step shuffle.

PR #8000, opened 2024-07-24
Fix Dashboard Scrolling When the Calendar is Opened

jrjohnson:
This should resolve the jump of the entire page when the calendar is opened. Instead of scrolling the calendar element into view I've used the scrolling container element itself to scroll to the position of the first event. This happens anytime the events list is updated.

The offsetTop property on an element uses a feature called offsetParent which discovers the nearest ancestor element with a relative or absolute position. I've added that positioning property to the scrolling container so it is correctly identified as the offsetParent.

Refs: ilios/ilios#4838 ilios/ilios#4867 which have both been closed in other ways, but should be re-tested to ensure I didn't break them more.


stopfstedt (approved):
LGTM

stopfstedt:
@dartajax  please have a look at this as well.

dartajax (approved):
that's quite nice indeed

dartajax:
In regression testing, testing the learner group calendar, I noticed that on demo currently the user is routed down to the calendar when the "Show Calendar" button is pressed. This is the preferred behavior. In this netlify PR version, this does not happen. It should - this constitutes a change request from me. @jrjohnson?

dartajax (changes requested):
In regression testing, testing the learner group calendar, I noticed that on demo currently the user is routed down to the calendar when the "Show Calendar" button is pressed. This is the preferred behavior. In this netlify PR version, this does not happen. It should - this constitutes a change request from me. @jrjohnson?

dartajax (approved):
gonna put in a new ticket for my change request which is outside the scope of this one - and I approve of this one

PR #8002, opened 2024-07-25
Add Timeouts for Deploys

jrjohnson:
These are defaulting to 6 hours, when they hang we don't get a failure message. Adding a step timeout should fail so we get a message, and then adding a job timeout just to save the world a bit if we run long.


stopfstedt (approved):

stopfstedt:
saving the world - one PR at a time.

PR #8008, opened 2024-07-25
prevent the application of user context filters on school events.

stopfstedt:
https://github.com/ilios/ilios/issues/5615


jrjohnson (approved):
Nice fix!

PR #8009, opened 2024-07-25
Improvements to Deploy

jrjohnson:
I have:
- Added a deploy test, not sure how useful it is compared to our build test, but we'll see
- Split deploys up into a matrix instead of using parallel `pnpm` tasks
- Added notification if PRs fail to deploy

This doesn't fix our current deploy issue, but it will make it easier to figure out what is wrong (and next time as well).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,6 +156,32 @@ jobs:
       - name: Build
         run: pnpm --filter ${{matrix.workspace}} run build
 
+  test-deploy-build:
+    name: Test Deploy Build
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -156,6 +156,32 @@ jobs:
...
 
+  test-deploy-build:
+    name: Test Deploy Build
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [20]
+        workspace:
```
stopfstedt, line 167: i like this a lot!

stopfstedt (approved):
nice one.

PR #8012, opened 2024-07-25
Relocate the Learner Group Calendar

jrjohnson:
Moving this up by it's toggle so it is instantly visible in groups with many members.

Fixes ilios/ilios#5616

### DIFF
--- a/packages/frontend/app/components/learner-group/root.hbs
+++ b/packages/frontend/app/components/learner-group/root.hbs
@@ -193,6 +193,9 @@
             />
           </div>
         {{else}}
+          {{#if this.showLearnerGroupCalendar}}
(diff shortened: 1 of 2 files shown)


stopfstedt (dismissed):
LGTM

dartajax (changes requested):
I hate to be that guy but a tidbit more vertical spacing would be nice just to move the calendar down a couple rows so the border line is not intermeshed with the button barriers - "Upload Group Assignments" "Manage" buttons ...

image

jrjohnson:
Good catch @dartajax, this is ready for re-review now.

dartajax (approved):
I am fine with this. Thanks @jrjohnson for doing that adjustment - it looks cleaner.

PR #8015, opened 2024-07-26
set phone-only cutoff at 399 so it covers the 380->400 pixel gap

michaelchadwick:
This should fix navigation buttons in non-English languages that cause the layout to careen ever so too much to the side on small screens

Fixes ilios/ilios#5607


stopfstedt (approved):
LGTM, but could use a second (or third) pair of eyes on it.

jrjohnson:
I see no issues, looks good on my real phone and on a simulated iPhone 12 mini.

PR #8025, opened 2024-07-30
Pass the Pull Request Number to Percy

jrjohnson:
Without this percy doesn't know how to connect tests with the right PR and report status.


stopfstedt (approved):
LGTM

for reference: https://www.browserstack.com/docs/percy/get-started/set-env-var#supported-parameters

PR #8027, opened 2024-07-30
normalized line-height for Course Descriptions and Parent Objectives

michaelchadwick:
Fixes ilios/ilios#3802

### DIFF
--- a/packages/ilios-common/app/styles/ilios-common/mixins/objectives.scss
+++ b/packages/ilios-common/app/styles/ilios-common/mixins/objectives.scss
@@ -1,4 +1,5 @@
 @use "../colors" as c;
+@use "font-size";
 @use "ilios-button";
 @use "ilios-form";
 @use "ilios-heading";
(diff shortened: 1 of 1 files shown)


jrjohnson (approved):

dartajax (dismissed):
image

michaelchadwick:
@dartajax Is the problem that there is space around that first description? Or that the first Parent Objectives column is too wide? Regardless, demo does that, too, and is a new bug to fix: https://demo.iliosproject.org/courses/2224?courseObjectiveDetails=true&details=true

dartajax:
> @dartajax Is the problem that there is space around that first description? Or that the first Parent Objectives column is too wide? Regardless, demo does that, too, and is a new bug to fix: https://demo.iliosproject.org/courses/2224?courseObjectiveDetails=true&details=true

I am fine with filing the new issue and will do so soon. I think it has to do with the fact that that one course objective was using the numbered list inside of the froala text box. Maybe when calculating the width of the content, the indentation caused by the use of the list is causing issues - something like that.

michaelchadwick:
@dartajax All tests now passing

PR #8031, opened 2024-08-01
all top-level learner groups are now bolded, regardless of subgroups

michaelchadwick:
Fixes ilios/ilios#5001


jrjohnson:
Apparently we have a test for the original way. Must have really liked it like that at some point 😆, I'll wait for tests to pass before doing a review.

michaelchadwick:
@jrjohnson The failing test is checking for `.strong` vs `.em` classes being applied, but now there's no difference, as both "branches" (subs) and "leaves" (no subs) have the same class applied. Is this test even useful?

https://github.com/ilios/frontend/blob/master/packages/test-app/tests/integration/components/learnergroup-tree-test.js#L100-L115

If so, maybe the classes should be changed like `.strong -> .branch` and `.em -> .leaf` and then if we ever change how `.branch` or `.leaf` are visually displayed we can update the CSS without worrying about breaking the test. Using class names that equate to HTML tags seems a little odd to me, as well.

jrjohnson:
I think you can probably just remove the test, or use some `data-test-*` attributes if it's testing something interesting now that they're all the same style.

michaelchadwick:
@jrjohnson I went ahead with my "branch/leaf" class idea, as it wasn't affecting anything else in the codebase, and it makes more sense (to me, at least) than using HTML tag names as CSS classes. Tests should pass now!

jrjohnson (approved):
I like it. Way better names.

dartajax (approved):
I like this and will be honored to merge it if / when tests pass.

PR #8034, opened 2024-08-01
add better text truncation method via FadeText

michaelchadwick:
References ilios/ilios#3482

My first crack at changing our current `<TruncateText>` system to a `<FadeText>` system: instead of removing HTML and squishing text together, leave it all as-is but put a fade-to-bg effect on it. Only working on Session Objectives->Description for now, but could be expanded.

**Old, Tired, Non-Ideal:**
Screenshot 2024-08-01 at 4 24 32 PM
**New, Wired, Superior(?):**
Screenshot 2024-08-01 at 4 28 07 PM

The vertical spacing it takes up can be adjusted. Here's another one with more faded away:
Screenshot 2024-08-01 at 4 31 20 PM


jrjohnson:
I like it on sessions, needs some attention on course objectives (and probably generically shouldn't specify it's own background, but should fade into whatever is behind it if possible)
Screenshot 2024-08-06 at 3 16 19 PM

michaelchadwick:
@jrjohnson Good catch. I'm also realizing `<TruncateText>` is used in a few more places that I need to account for, so back to the drawing board.

michaelchadwick:
Places where `<TruncateText>` is currently used and `<FadeText>` could replace:
* CourseObjectiveList
  * ObjectiveSortManager
* SessionObjectiveList
  * OfferingManager (this seems like it shouldn't change; only location, so usually short)
* SessionsGridOffering (this seems like it shouldn't change: only location, so usually short)
* WeekAtAGlance, as seen on an event's description (this seems like it shouldn't change: not in a grid)
  * LearningMaterialListItem, as seen on WaaG in an LM's "Public Notes" (this could use some discussion, as it's essentially the same as the parent list item, except that it displays slightly differently when expanded)
* MaterialListItem (this seems it shouldn't change: instructor name(s))

michaelchadwick:
@jrjohnson I believe I fixed the background color issue, but had to be specific instead of general (thankfully, there are only three places to be specific about right now).

packages/ilios-common/addon/components/editable-field.hbs
```diff
@@ -70,18 +74,20 @@
...
-                  aria-label={{t "general.expand"}}
-                  title={{t "general.expand"}}
-                  data-test-expand
-                  {{on "click" expand}}
-                >
-                  <FaIcon @icon="angles-down" />
-                </button>
+              {{#if isFaded}}
+                <div class="fade-text-control">
+                  <EllipsisIcon />
```
jrjohnson, line 79: I think we should ditch the ellipsis, it doesn't make as much sense in a faded world.

packages/ilios-common/app/styles/ilios-common/components/fade-text.scss
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.fade-text-control {
+  bottom: 140px;
+  height: 130px;
+  padding-top: 140px;
+  position: relative;
+}
+
+span {
```
jrjohnson, line 18: The components in here aren't automatically wrapped in any way so this `span` is targeting every span in the application. Take a look at some of the other component sass and see how they use a top level class to contain all the styles and ensure they don't leak out into the rest of the app. 

Everything here should go under `.fade-text {`.

packages/ilios-common/app/styles/ilios-common/components/fade-text.scss
```diff
@@ -0,0 +1,48 @@
...
+      background-image: linear-gradient(
+        to bottom,
+        transparent,
+        c.$culturedGrey
+      );
+    }
+  }
+}
+
+.course-objective-list {
```
jrjohnson, line 30: I'm not comfortable with this component reaching so far into the styling of other component to get its job done. There may not be a better way, but I'd like us to discuss a bit before taking this step.

jrjohnson (changes requested):
Needs team discussion (and a couple tweaks).

packages/ilios-common/addon/components/objective-sort-manager.hbs
```diff
@@ -42,7 +42,7 @@
           {{/unless}}
           <span class="draggable-object-content">
             <span>
-              <TruncateText @text={{item.title}} />
+              <FadeText @text="{{item.title}}" />
```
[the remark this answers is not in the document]
michaelchadwick: Oops, good catch

packages/ilios-common/addon/components/editable-field.hbs
michaelchadwick: Sounds like a good discussion item for next meeting.

packages/ilios-common/addon/components/fade-text.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+    if (typeOf(this.args.text) !== 'string') {
+      return this.args.text.toString();
+    }
+
+    return this.args.text;
+  }
+  get cleanText() {
+    // return this.text.replace(/(<([^>]+)>)/gi, '');
+    return this.text;
```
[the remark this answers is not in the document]
michaelchadwick: Sounds correct to me.

packages/ilios-common/addon/components/editable-field.hbs
michaelchadwick: Actually, now that I think about it more, ditching this makes perfect sense, since we're not actually chopping off any text, which is usually what the ellipses represents. Will ditch.

packages/ilios-common/app/styles/ilios-common/components/fade-text.scss
michaelchadwick: We should talk about this at our next meeting. Perhaps there's a better way to handle it, but it was the only way I could see at the time.

michaelchadwick:
Note to self: attempt to create a mixin that can be used to facilitate the gradient from transparent->text background color.

michaelchadwick:
@jrjohnson I moved the component-specific styles in `fade-text.scss` into their respective component style files, which makes more sense. Two of them had to be in addition to the mixin they included because of different background colors, whereas the other one I was able to add to the mixin itself.

jrjohnson (commented):
I really love how this looks, I found a couple of new behavior / visual issues in testing:

1. Clicking on the unexpanded text does nothing. In the current implementation if you click on a collapsed objective the editor opens. I think that is good, but I'd also accept that clicking anywhere in the text expanded the entire block.
2. If the text is indented (like in a list) the control floats up into the middle instead of being at the bottom, this isn't an issue with our current control because the indentation never happens due to the HTML stripping: 
Screenshot 2024-08-21 at 12 30 53 PM
3. Because we're using text length to collapse, but the actual behavior is height based it's a little confusing why only some things collapse (this may only be in tests and not in real life, and may not be fixable): 
Screenshot 2024-08-21 at 12 37 17 PM

Not sure if we should go ahead with these or try and find fixes. Let's discuss.

michaelchadwick:
@jrjohnson All good points. Will look into these.

michaelchadwick:
@jrjohnson I fixed all the stuff mentioned in point 1, but not sure how to approach point 2 without re-architecting the whole truncation system.

michaelchadwick:
@jrjohnson This has one failing testapp test, because it's testing for truncation length, and not the new visual fading thing. Not sure how to rewrite test, hmmm.

https://github.com/ilios/frontend/blob/master/packages/test-app/tests/integration/components/editable-field-test.js#L143-L154

michaelchadwick:
Having some issues with getting a JS triggered click to target the correct item (thus, test won't pass). Manually clicking with a mouse/touch works fine, though. Frustrating -_-

dartajax:
image

image

michaelchadwick:
@dartajax Excellent catch. The fade overlay is too tall and interfering. Back to the drawing board.

michaelchadwick:
@stopfstedt @jrjohnson This needs a review again, and also there are two `skip`ped tests in the `fade-text-test.js` file because I keep getting random results from the tests: sometimes it loads the text faded, sometimes it doesn't, almost like a race condition. Need another pair of eyes/brain cells on them, if possible.

packages/test-app/tests/integration/components/fade-text-test.js
```diff
@@ -0,0 +1,155 @@
...
+
+      await component.expand.click();
+
+      assert.dom('.display-text-wrapper', this.element).doesNotHaveClass(fadedClass);
+      await component.collapse.click();
+      assert.dom('.display-text-wrapper', this.element).hasClass(fadedClass);
+
+      await render(hbs`
+        <FadeText @text={{this.longHtml}} as |displayText expand collapse getTextDims updateTextDims isFaded expanded|>
+          <div class="display-text-wrapper{{if isFaded ' is-faded'}}">
```
jrjohnson, line 69: Big fan of testing the block invocation here, but the actual content of the test probably shouldn't mirror the component without the block. Instead just enough HTML should go here to ensure that the passed properties are as you would expect. I think the amount of content needed in the block invocation may be an issue with the existing truncation component as well. I think it's ok right now, but might be worth another pass in the future. I wonder if we could split out different parts of this into [named blocks](https://guides.emberjs.com/release/components/block-content/) so they don't have to be overridden.

packages/ilios-common/addon/components/editable-field.hbs
```diff
@@ -70,33 +77,41 @@
...
-                  aria-label={{t "general.expand"}}
-                  title={{t "general.expand"}}
-                  data-test-expand
-                  {{on "click" expand}}
+              {{#if isFaded}}
+                <div
+                  class="fade-text-control"
+                  data-test-fade-text-control
+                  {{! template-lint-disable no-invalid-interactive}}
+                  {{on "click" (fn this.setIsEditing true)}}
```
jrjohnson, line 85: This is the part I get worried about the fairly large API needed to call this in the block form. This is a button now in the component, but has remanned a clickable div here. I'm thinking passing the button itself in as a parameter might work here? But I can't recall if I've ever done that anywhere.

packages/ilios-common/addon/components/editable-field.hbs
```diff
@@ -33,7 +33,7 @@
         </span>
       </span>
     {{else}}
-      <span>
+      <span class="viewer">
```
jrjohnson, line 36: I don't see anywhere this is used. Probably old.

jrjohnson (commented):
Comments and probably useless advice added but I tried. I did discover a maybe bug, it seems like expanding doesn't work, it only gets bigger for this objective? I clicked and the icon changed, but the text is still faded.
Screenshot 2024-09-23 at 10 12 33 PM

michaelchadwick:
> Comments and probably useless advice added but I tried. I did discover a maybe bug, it seems like expanding doesn't work, it only gets bigger for this objective? I clicked and the icon changed, but the text is still faded. Screenshot 2024-09-23 at 10 12 33 PM

Excellent bug catch. I made an uneducated change and introduced this bug, but figuring out why the bug happens now informs me to why not to make that uneducated change again.

packages/ilios-common/addon/components/editable-field.hbs
michaelchadwick: I got this template from `<TruncateText>`, which does a similar thing, so figured it was kosher for now. Is this change required for merging?

michaelchadwick:
This PR (will) fix(es) ilios/ilios#3482, btw.

packages/ilios-common/addon/components/editable-field.hbs
jrjohnson: I hate to move on, but this thing is already pretty big. I accept this, let's keep it for now.

jrjohnson (dismissed):
I'm not in love with the API for this component, but I am in love with how much better this looks. Let's test it, merge it, and enjoy it.

michaelchadwick:
@dartajax I finally got every test to pass; just have Percy to finish. Can you take a look, and bring this one home? Thanks a bunch.

dartajax (approved):
looks good to me

PR #8037, opened 2024-08-02
Disabled nextPage button on Admin-Users list if no more results to show

michaelchadwick:
Fixes ilios/ilios#5544

My approach required adding another Ember Data query to be run to get the total users (as the main query limits the user count for the UI), but we get the benefit of disabling the nextPage button if appropriate AND a user count.

I understand if this is not ideal, so if it needs to be rejected (or maybe just modified?) I understand.

### DIFF
--- a/packages/frontend/app/components/ilios-users.hbs
+++ b/packages/frontend/app/components/ilios-users.hbs
@@ -19,6 +19,9 @@
     <div class="header" data-test-header>
       <span class="title" data-test-title>
         {{t "general.users"}}
+        {{#if this.getTotalUsers.lastSuccessful.value.length}}
+          ({{this.getTotalUsers.lastSuccessful.value.length}})
+        {{/if}}
       </span>
       <div class="actions">
         {{#if (or @showNewUserForm @showBulkNewUserForm)}}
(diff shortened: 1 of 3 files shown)


michaelchadwick:
Still needs some test, too.

stopfstedt (changes requested):
unfortunately, this isn't going to cut it - the load for totals lags way behind, so there's a significant disconnect between the headline updates with the total number and the nav unlocking. 
it also causes my browser to lock up (getting the "your browser has become unresponsive" message in FF), so the double-load might be a bit much.

suggesting to drop the offset/limit params on the API request, and then paginate on the entire result set each time. this might allow us to drop the `@limitless` argument from the `<PagedlistControls>` component.

michaelchadwick:
I was confused by your comment initially, because I had only been testing viewing users with an input filter, and it never took very long. I just went to the page with no input filter and see what you mean -_-

GraphQL attempt coming in next...

michaelchadwick:
After realizing that GraphQL wasn't any faster to grab all users (20k+ results), I went with @stopfstedt's idea overfetching by 1 on the limit and comparing. That seems to work and doesn't require a second API call. It loses the `Users ([userCount])` niceness, but that was an added benefit.

Also, as an added change, I made it so that the grid of users shows a spinner when it's searching, because there's no visual feedback when you change the input filter or change the offset. Maybe this isn't the slickest implementation, but it's a start.

stopfstedt (approved):
LGTM.

PR #8039, opened 2024-08-05
Replace Some Modifier Loaders

jrjohnson:
Refs ilios/ilios#5374


michaelchadwick (approved):
Death to ember-render-modifiers

PR #8046, opened 2024-08-06
Combine Leaning Material Tests

jrjohnson:
Having these in separate modules is making this test flaky, it seems like a lot of this is just old ideas about test data. By passing some stuff around I've created the same tests, but without the double setup and, hopefully, without the flakiness.


zorgbort (approved):
Approving with Zorgbort to see if this helps with this flaky test.

stopfstedt (approved):

PR #8057, opened 2024-08-08
Reduce our Dependabot Runs

jrjohnson:
Tamping down our dependabot runs to Sunday after our lock file update job runs. This will reduce the number of PRs we need to test and ensure we don't have to spend time on ones that would get auto updated with a lock file update anyway. Those that do need attention will be easier to handle once a week.


stopfstedt (approved):
weekly it is.

PR #8059, opened 2024-08-08
Silence the Mixed Declarations SASS Deprecation

jrjohnson:
This is a very noisy deprecation that it looks like we can ignore. Everywhere this is triggered we appear to be expecting the new behavior, we will probably need to wait for the update that changes things and then fix any styles that have an issue.


stopfstedt (approved):
:heart: this!

PR #8061, opened 2024-08-09
reduce padding on LG members/subgroups when none have been added

michaelchadwick:
Fixes ilios/ilios#5633

### DIFF
--- a/packages/frontend/app/styles/components/programs/list.scss
+++ b/packages/frontend/app/styles/components/programs/list.scss
@@ -2,4 +2,8 @@
 
 .list {
   @include m.main-list-box-table;
+
+  &.empty {
+    padding: 0;
+  }
 }
--- a/packages/frontend/app/styles/components/programs/list.scss
+++ b/packages/frontend/app/styles/components/programs/list.scss
@@ -1,9 +1,7 @@
 @use "../../mixins" as m;
 
-.list {
-  @include m.main-list-box-table;
-
-  &.empty {
-    padding: 0;
+.programs-root {
+  .list {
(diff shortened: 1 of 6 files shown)


packages/frontend/app/styles/components/programs/list.scss
```diff
@@ -2,4 +2,8 @@
 
 .list {
   @include m.main-list-box-table;
+
+  &.empty {
```
jrjohnson, line 6: You put this in the right place to work, but this style is coming from the wrong file.

It's outside the scope of this PR but this `programs/list.scss` style should be scoped down to only target the program list  and not every `.list` in the application (everything should be under a `.programs-list{` class). Can you make that change here or circle back and make a PR just to do that and then re-do the style in this one? It will also require creating a component style for the learner group list (and possibly other lists which are also accidentally replying on this style).

jrjohnson (changes requested):
I like it.  The change exposes an issue in our styles that I messed up 3 years ago which needs to get fixed. Sorry.

packages/frontend/app/styles/components/programs/list.scss
michaelchadwick: Let's chat about this at our next meeting.

michaelchadwick:
@jrjohnson Hoo boy. Trying fix the change requested led me down a rabbit hole of components and styles and mixins. I _think_ I was able to put on the brakes and just fix the things requested, but I may have threw in something that was needed, even if a bit out of scope for this task. The real fix would take much longer and will have to wait.

PR #8067, opened 2024-08-13
Fix Issues with Report Downloads

jrjohnson:
Both the term and instructor reports had issues with their downloads that made them unusable. 

Fixes ilios/ilios#5642


saschaben:
basic user testing seems good. not seeing any issues.

FYI Let's create a follow-on ticket to expand data in and improve display of downloaded data in particular our terms reports.

stopfstedt (approved):

PR #8072, opened 2024-08-15
Use Ember Glatpickr Addon

jrjohnson:
This addon seems to be well maintained and works well. Using it instead of our custom component means we can drop some of the modifiers and possibly help with some of our flaky tests.

WIP:
- [ ] replace the rollover picker as well
- [ ] Feedback from testing
- [ ] Remove flatpickr dependency
- [ ] Possibly replace our test helpers with those provided from the addon


jrjohnson:
Superseded by #8312

PR #8074, opened 2024-08-16
Optimize the Instructors for Course Report

jrjohnson:
Because of the joins required this report performed very poorly using the API as the number of sessions and offerings in a course grew. Instead we can use our GraphQL API to get the data we want from the course down and significantly improve the performance.

I didn't DRY this up, we probably need to replace the reporting service for download and re-use the data fetching in the subject components, but that is for another day.

### DIFF
--- a/packages/frontend/app/services/reporting.js
+++ b/packages/frontend/app/services/reporting.js
@@ -162,7 +162,7 @@ export default class ReportingService extends Service {
 
   async getInstructorsArrayResultsForCourse(courseId) {
     const userInfo = '{ id firstName middleName lastName displayName }';
-    const block = `instructorGroups {  users ${userInfo}}instructors ${userInfo}`;
+    const block = `instructorGroups {  users ${userInfo}} instructors ${userInfo}`;
     const results = await this.graphql.find(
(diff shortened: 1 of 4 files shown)


packages/frontend/app/services/reporting.js
```diff
@@ -160,27 +160,68 @@ export default class ReportingService extends Service {
     ].concat(mappedResults);
   }
 
+  async getInstructorsArrayResultsForCourse(courseId) {
+    const userInfo = '{ id firstName middleName lastName displayName }';
+    const block = `instructorGroups {  users ${userInfo}}instructors ${userInfo}`;
```
stopfstedt, line 165: shouldn't there be a space in there?

```suggestion
    const block = `instructorGroups {  users ${userInfo}} instructors ${userInfo}`;
```

jrjohnson: Apparently it's not needed, but I agree it should be there. Nice catch!

saschaben:
on netlify 'All Instructors for Session Type XX' still fail to complete. no error thrown in the inspector panel.

PR #8076, opened 2024-08-16
fixed Ember.js test framework deprecation on ember serve

michaelchadwick:
moved generateTests property to correct place in module.exports


stopfstedt (approved):
LGTM

PR #8081, opened 2024-08-16
replaces Resources with Tracked Async Data.

stopfstedt:

### DIFF
--- a/packages/ilios-common/addon/components/learnergroup-tree.js
+++ b/packages/ilios-common/addon/components/learnergroup-tree.js
@@ -1,11 +1,8 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
-import { filter } from 'rsvp';
 import { service } from '@ember/service';
-import { use } from 'ember-could-get-used-to-this';
 import { TrackedAsyncData } from 'ember-async-data';
 import { cached } from '@glimmer/tracking';
-import AsyncProcess from 'ilios-common/classes/async-process';
 
 export default class LearnergroupTree extends Component {
   @service intl;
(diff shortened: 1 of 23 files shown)


packages/ilios-common/addon/components/learnergroup-tree.js
```diff
@@ -46,21 +37,6 @@ export default class LearnergroupTree extends Component {
     return !this.filterMatch;
   }
 
-  /**
-   * Recursively search a group tree to see if there are any children which have not been selected.
-   **/
-  async getHasUnSelectedChildren(children, selectedGroups) {
```
jrjohnson, line 52: I'm assuming this is the dead code? Nice find!

jrjohnson (approved):
Never did get used to it, like to see it go.

jrjohnson:
@michaelchadwick I tossed this over to you for a look, but I mainly just want to make sure you take a look at [this post](https://www.pzuraq.com/blog/introducing-use) and as a reminder for us to talk about the evolution of async in ember and how it's sprinkled into Ilios next time we have some in person time.

michaelchadwick:
@jrjohnson I want to be up front and say that I spent a non-trivial time just clicking the light/dark mode toggle on that blog because the transition is so smooth and the little house's details change in a cool way :D

Still getting through the blog post, but will report back on my thoughts.

michaelchadwick:
@jrjohnson Got through the blog post. The whole paradigm shift it seemed to introduce seemed a lot like VueJS (JS + HTML in same file). Also, I see Ilios uses `ember-could-get-used-to-this` and the `@use` decorator quite a lot, but this change seems to be moving Ilios _away_ from that to `TrackedAsyncData`. I feel like I don't have enough background to say whether this is better or worse, so I'll trust the team..._for now_. ;P

PR #8082, opened 2024-08-17
Refactor Cohort Loading 

jrjohnson:
Replacing the buffers that get set when the component is loaded with a
change tracking system. This will ensure we're responsive to changes in
the user.

Fixes ilios/ilios#5653


stopfstedt (approved):
LGTM, especially with the new and improved tests.

PR #8085, opened 2024-08-20
remove unneeded .slice() calls on data in frontend root components

michaelchadwick:
Refs ilios/ilios#5660

This change removes all the unneeded calls to `.slice()` in all the root `frontend` root components (not located in a `components/**` subdir.


stopfstedt (dismissed):
LGTM

stopfstedt:
i'll give this another look-over tomorrow. after discovering that `sort()` does not work on these ember-data relationships, i suspect that other in-place operations like `reverse()` don't work either.

stopfstedt (dismissed):
LGTM

dartajax (approved):

PR #8088, opened 2024-08-21
rm unneccessary calls to Array.slice() from ember-data models.

stopfstedt:
refs ilios/ilios#5660

more to come in the next PR.

### DIFF
--- a/packages/ilios-common/addon/models/vocabulary.js
+++ b/packages/ilios-common/addon/models/vocabulary.js
@@ -15,8 +15,7 @@ export default class Vocabulary extends Model {
   terms;
 
   async getTopLevelTerms() {
-    const terms = await this.terms;
-    return filter(terms.slice(), async (term) => {
+    return filter(this.terms, async (term) => {
       return !(await term.parent);
     });
   }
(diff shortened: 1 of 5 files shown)


packages/ilios-common/addon/models/vocabulary.js
```diff
@@ -15,8 +15,7 @@ export default class Vocabulary extends Model {
   terms;
 
   async getTopLevelTerms() {
-    const terms = await this.terms;
-    return filter(terms.slice(), async (term) => {
+    return filter(this.terms, async (term) => {
```
jrjohnson, line 18: I'm not sure about this one. Does `RSVP.filter` resolve the first argument `this.terms` when it is a promise? Documentation at https://github.com/tildeio/rsvp.js/ doesn't even cover `filter` so I'm not sure.

jrjohnson: Nevermind, I found where it [looks like that happens](https://github.com/tildeio/rsvp.js/blob/master/lib/rsvp/filter.js#L138)

jrjohnson (approved):

PR #8090, opened 2024-08-22
fix incorrect offset on new Admin->Users queries

michaelchadwick:
Fixes ilios/ilios#5665

Start tracking current query and compare to previous queries so we can reset offset if it's a different search term.


jrjohnson (approved):
Solid solution. ✅

PR #8092, opened 2024-08-22
Drop Support for Safari 15.x

jrjohnson:
Changing our support to only cover the last two major safari versions on OSX as our stats show that this will cover our users and we can move more quickly to some modern Javascript and improve performance for everyone else by not including dead code.

We wern't able to drop the crypto polyfill because ios Safari doesn't
support it in "insecure" contexts at this version. Hoping that will be
fixed in the future.


stopfstedt (dismissed):
SUPERAPPROVED

stopfstedt (approved):
LGTM

PR #8095, opened 2024-08-23
removed unneeded slice() calls from frontend components in subdirectories

michaelchadwick:
refs ilios/ilios#5660

Except for a few instances where the data was going to be `sort()`-ed or `reverse()`-ed or similar, and we now know that does not allow for modification yet, I removed all the `slice()` calls I could find.


packages/frontend/app/components/learner-group/calendar.js
```diff
@@ -47,7 +47,7 @@ export default class LearnerGroupCalendarComponent extends Component {
     }
     const offerings = await all(mapBy(learnerGroups, 'offerings'));
     const flat = offerings.reduce((flattened, obj) => {
-      return [...flattened, ...obj.slice()];
```
jrjohnson, line 50: Well this is a bad variable name!

packages/frontend/tests/integration/components/program-year/list-test.js
```diff
@@ -96,29 +96,29 @@ module('Integration | Component | program-year/list', function (hooks) {
     await component.newProgramYear.years.select(thisYear);
     await component.newProgramYear.done.click();
     const programYears = await this.owner.lookup('service:store').findAll('program-year');
-    const sortedProgramYears = sortBy(programYears.slice(), 'id');
+    const sortedProgramYears = sortBy(programYears, 'id');
```
jrjohnson, line 99: I'm a little bit worried about updating the tests and the application code in the same PR here. Are the changes linked? Or could the tests updates be in a second PR?

michaelchadwick: Should be linked to the application code here: https://github.com/ilios/frontend/pull/8095/files#diff-12e203d3f8d49a7214eafaf3aa33d10d74239acd9d6ac6acc1bae17df2fc5256

Also, I changed tests alongside app code with my last de-slice()-ing, so... :D

jrjohnson: I accept this 🤠

jrjohnson (approved):

PR #8103, opened 2024-08-26
timed releases start/end now intelligently sync with each other

michaelchadwick:
If you add an `endDate`, it will make it the same as the `startDate`. And if you later update one or the other, it will make sure that the `endDate` is not before the `startDate`.

### DIFF
--- a/packages/ilios-common/addon/components/learningmaterial-manager.js
+++ b/packages/ilios-common/addon/components/learningmaterial-manager.js
@@ -49,6 +49,12 @@ export default class LearningMaterialManagerComponent extends Component {
     return this.type === 'citation';
   }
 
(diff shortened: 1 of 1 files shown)


michaelchadwick:
OK, fixed test issues by fixing the actual logic, which was not 100% correct.

stopfstedt:
fixed https://github.com/ilios/ilios/issues/5661

stopfstedt (approved):
LGTM

stopfstedt (approved):
LGTM, nice cleanup along the way.
code climate should be ignored here, this level of complexity is acceptable to me.

dartajax (approved):
works for me - I tested this out

PR #8107, opened 2024-08-30
rm obsolete ember-metrics package and the GA integration with it.

stopfstedt:
we don't track users via Google Analytics (anymore) - this gets rid of the code that supports it.


michaelchadwick:
re: `the_jimi_metrics_experience`

@stopfstedt I didn't realize we could use _cool_ branch names. This changes everything

jrjohnson (approved):
🤩 😍 ❗

PR #8128, opened 2024-09-10
Clean up FaIcon template's class attribute

michaelchadwick:
Currently, the `<FaIcon>` has an overstuffed `class` attribute that often has conditionals that are not met. This moves all the extra optional classes to a class function and only adds them in (and the spaces between them) if they are passed in.

### DIFF
--- a/packages/ilios-common/addon/components/fa-icon.hbs
+++ b/packages/ilios-common/addon/components/fa-icon.hbs
@@ -1,6 +1,6 @@
 {{! template-lint-disable eol-last }}
 <svg
-  class="awesome-icon fa-{{@icon}} {{if @spin "spin"}} {{this.flip}} {{if @listItem "list-item"}} {{if @fixedWidth "fixed-width"}}"
+  class="awesome-icon fa-{{@icon}}{{this.extraClasses}}"
   data-icon={{@icon}}
   aria-hidden={{this.ariaHidden}}
   focusable={{this.focusable}}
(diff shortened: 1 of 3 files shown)


jrjohnson (commented):
I think, but haven't confirmed, that you have to do some extra work to mark this string as safe before ember will let you use it like this. That might be old school though. But worth checking for sure and I wanted to add a note.

stopfstedt (approved):
LGTM

stopfstedt:
> I think, but haven't confirmed, that you have to do some extra work to mark this string as safe before ember will let you use it like this. That might be old school though. But worth checking for sure and I wanted to add a note.

i don't think that's necessary anymore. check out the `flip` getter as it is current implemented https://github.com/ilios/frontend/blob/master/packages/ilios-common/addon/components/fa-icon.js#L25 - no safe-string marking there either, and it works.

packages/ilios-common/addon/components/fa-icon.hbs
```diff
@@ -1,6 +1,6 @@
 {{! template-lint-disable eol-last }}
 <svg
-  class="awesome-icon fa-{{@icon}} {{if @spin "spin"}} {{this.flip}} {{if @listItem "list-item"}} {{if @fixedWidth "fixed-width"}}"
+  class="awesome-icon fa-{{@icon}}{{this.extraClasses}}"
```
stopfstedt, line 3: needs a space here

```suggestion
  class="awesome-icon fa-{{@icon}} {{this.extraClasses}}"
```

stopfstedt (changes requested):
needs a whitespace as separator in the class attribute, otherwise the icon flipping won't work.

for example, this can be seen in the user admin pagination

image

stopfstedt (approved):
LGTM

dartajax:
> can't merge this one or restart (re-start) the test

that's because one of the test is still running - never mind

PR #8130, opened 2024-09-11
Wait for Loaded Cohorts

jrjohnson:
Changing async patterns in this component left a gap when it is visited for the first time without the preloading that normally happens on the dashboard. I've updated the check to ensure we actually have data loaded as expected.

Fixes ilios/ilios#5699


stopfstedt (approved):
LGTM

dartajax (approved):
I followed the detailed instructions provided by @saschaben in the source ticket - this seems to fix that situation - click test = passed so from my perspective, approved.

PR #8131, opened 2024-09-11
Only Build Good Sourcemaps in Production

jrjohnson:
[devtool](https://webpack.js.org/configuration/devtool/) is what webpack uses to generate source maps and it can be configured to work in different ways for our production builds where we need high quality source maps to connect to sentry we want only the best, but for our other builds we need something with a much faster rebuild time.

Refs #8093


stopfstedt (approved):
thanks.

PR #8134, opened 2024-09-11
Remove test-app Percy Tests

jrjohnson:
Turns out we're not capturing any snapshots in the test-app so we don't need to run these tests. This solves a percy issue where the build from the test-app was reported last to percy and percy displayed a pipeline error because there aren't actually any screenshots in taken.

Refs #8108


stopfstedt (approved):
supernice.

PR #8135, opened 2024-09-11
improve Calendar filters display on mobile

michaelchadwick:
Fixes ilios/ilios#5701

* Reduced font-size of `<h5>`
* Changed `<h5>` height across breakpoints
* Removed `small-filter-list` class from filters so it takes up available width
* Added spaces to 'Program/Cohort' term in `en-us` and `es` (commit message was typo, oops) so that it wraps better


stopfstedt (approved):
LGTM

PR #8136, opened 2024-09-12
fix missing session timed release information

michaelchadwick:
Fixes ilios/ilios#5697

Looks like `.fromJSDate()` is looking for a JS `Date` object, not a string, so I converted it. There are several `DateTime` functions that could conceivably do the job, too, but this way kept tests passing, so it seemed the most prudent.

### DIFF
--- a/packages/ilios-common/addon/components/timed-release-schedule.js
+++ b/packages/ilios-common/addon/components/timed-release-schedule.js
@@ -17,6 +17,6 @@ export default class TimedReleaseSchedule extends Component {
     if (!this.args.startDate) {
       return false;
     }
-    return DateTime.fromISO(this.args.startDate) > DateTime.now();
+    return DateTime.fromJSDate(new Date(this.args.startDate)) > DateTime.now();
   }
 }
--- a/packages/ilios-common/addon/components/timed-release-schedule.js
+++ b/packages/ilios-common/addon/components/timed-release-schedule.js
@@ -17,6 +17,6 @@ export default class TimedReleaseSchedule extends Component {
     if (!this.args.startDate) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


packages/ilios-common/addon/components/timed-release-schedule.js
```diff
@@ -17,6 +17,6 @@ export default class TimedReleaseSchedule extends Component {
     if (!this.args.startDate) {
       return false;
     }
-    return DateTime.fromJSDate(this.args.startDate) > DateTime.now();
+    return DateTime.fromJSDate(new Date(this.args.startDate)) > DateTime.now();
```
jrjohnson, line 20: Probably don't need the interim `Date`, if this is a string coming in and it is correctly formatted you can use [fromISO](https://moment.github.io/luxon/#/parsing?id=iso-8601) directly.

```suggestion
    return DateTime.fromISO(this.args.startDate) > DateTime.now();
```

jrjohnson (changes requested):

packages/ilios-common/addon/components/timed-release-schedule.js
michaelchadwick: I tried that one, but tests failed. I will change it and fix the tests instead.

jrjohnson: Ah right, because tests were assuming at `Date` object. Yes, update the tests to match reality and then I guess we'll go full typescript. I see no reasonable alternative to check values like this and get a good error 😖

jrjohnson: Offline conversation, we're going to move forward with this fix and circle back for cleanup.

jrjohnson (dismissed):

dartajax (approved):
functionally this works for me

PR #8137, opened 2024-09-13
fix some routes that have horizontal scrolling on small screen

michaelchadwick:
Fixes ilios/ilios#5703

Fixed some horizontal scrolling I found while browsing the site on a mobile device in portrait (iPhone 12, viewport width: 390px)


jrjohnson (approved):
Code looks good, I'll let Percy and Dave have the final say on style.

PR #8138, opened 2024-09-13
Update ember-cli to v5.11.0

jrjohnson:
Using ember-cli-update with all blueprints.


michaelchadwick:
@jrjohnson What happened to 5.10.0?

jrjohnson:
No one got around to running the update, but this is inclusive of everything that was in 5.10.0 as well.

michaelchadwick (approved):
Checked a similar update on my test app, also from 5.9.0->5.11.0, and it went swimmingly, so I'm OK with these changes.

PR #8140, opened 2024-09-17
no need to track the component element, we can just traverse the DOM upwards instead.

stopfstedt:


stopfstedt:
for demo-ing purposes only.

PR #8141, opened 2024-09-17
dashboard->calendar scrolly fixxy

michaelchadwick:
Somewhat fixes ilios/ilios#5716

This PR does two things: 1) it removes an unnecessary `{did-insert}`, and 2) it moves the Dashboard->Calendar window scroll up a bit so the top of the viewport sits just above the filter headers, instead of hiding them.


jrjohnson (approved):
I'm ok with this as an interim solution. I'd rather see no scroll at all when clicking, but at least I can see the headers for the filters now.

dartajax (approved):
It's an improvement I'll give this PR that - enough to approve it for sure.

PR #8142, opened 2024-09-18
fix: memory-leak in connection-status

BobrImperator:
Something I've noticed while browsing the code which was a quick fix.

- Event listeners were incorrectly torn down
`window.addEventListener` doesn't return anything and the actual function reference wasn't provided to `removeEventListener`.

This is probably fine in production as there's only one instance of this component that's also only in the application.hbs, but it did cause memory-leaks between tests. Which might slow down tests and in the worst case scenario crash when they're all ran in a single tab e.g. "localhost:4200/tests" route or `ember test --server`.

This is also somewhat mitigated by the fact that the project is using `ember exam` for testing where tests are executed in chunks.

## Before
image

## After
image


jrjohnson (approved):
Awesome find, thanks!

PR #8143, opened 2024-09-19
Disable Array Prototypes

jrjohnson:
We don't need this anymore, the issue we were originally hitting was caused by upstream ember issues as well as some of the helpers from ember-composable-helpers (which we have since added directly here). We can go ahead and disable these and silence the deprecation warnings being emitted since updating ember.


stopfstedt (approved):
awesome.

PR #8144, opened 2024-09-19
Rearrange Report Results

jrjohnson:
Ripped apart the way we display results in order to move the download button inside the results. This improves the download functionality:
-  to re-use the data that is already loaded
- to build the data in the same place for display and download
- to allow reports that are run without saving to be downloaded

Fixes ilios/ilios#5652

Todo from first demo:
- [x] Move the academic year selection to the left column (download last in mobile)
- [x] Fix download button size in various places (particularly mobile where it stretches)
- [x] Singualirze course title in download
- [x] Reset the year when running another report
- [x] Ensure the best year is selected when building a report
- [x] Hide the academic year filter when the year is used to create the report


jrjohnson:
@dartajax I need to clean up a few things and move them around in the UI, but this should be functionally complete. Would you mind taking a first pass at downloads and report results and see if you can break anything or if anything pops out at you as needing work before this is merged. I'll keep working on my stuff in parallel.

dartajax:
I assigned myself as a tester not to merge - `DO NOT MERGE` in place.

dartajax:
I feel I have the luxury of testing this for a while since it won't go anywhere for a bit. `READY TO MERGE` needs to be added especially since I added "DO NOT MERGE` so it's doubled up on not merging. That having been said @jrjohnson could you make the heading when a Course report is downloaded to be "Course" rather than "Courses" - plural not needed as data table header.

dartajax (changes requested):
When running a new report, "All Academic Years" should be the value in the drop-down of the output, even if the user was tinkering with other years on the previous report generated - my $.02.

image

dartajax (changes requested):
I hate to be that person but when I run a report for all sessions in academic year, the default selecting comes up 2009. Certainly if we were going to pick a year, 2024 would be better since "All Academic Years" is not available until ... you run the report. Here is what a report based on 2024 looks like on output for reference. First if we need a year make it the current year. If we don't need a year, make "All Available Years" available on the drop-down like it is in the output as will be shown below. 

image

image

As one would surmise, there are no 2016 return values in a 2024 dataset.

jrjohnson:
Good catch @dartajax, I've hidden that filter when the year is used to build the report and also fixed the academic year report association to select the current year instead of defaulting to the oldest year.

jrjohnson:
If it ran the fetch again that's a bug. I'm on it. Thanks!

dartajax (approved):

michaelchadwick:
> If it ran the fetch again that's a bug. I'm on it. Thanks!

First fetch when running, 1.2 min and finished successfully. Second fetch when downloading, 4 min and timed out.

jrjohnson:
Actually I forgot about this one. The sessions download has a lot more data than the on screen report. So it's split up into two parts and the fetch has to re-run. All sessions in Medicine is always going to be a huge amount of data. I'd like to leave this as is for now and possibly circle back if we want to slow down the first on-screen report to load more data or accept the current second load network request.

PR #8145, opened 2024-09-20
More Improvements to Calendar Courses Filter

jrjohnson:
By marking the filter parent as relative positioned we can lookup the distance of each year from the top of that filter and scroll the correct one into view without opening anything else.

I discovered while testing this that if the current academic year doesn't have any course we were opening the oldest year instead of the latest year with courses. That seems wrong to me so I modified that behavior as well.

Refs ilios/ilios#5716

### DIFF
--- a/packages/ilios-common/addon/components/dashboard/courses-calendar-filter.js
+++ b/packages/ilios-common/addon/components/dashboard/courses-calendar-filter.js
@@ -5,7 +5,6 @@ import { action } from '@ember/object';
 import { service } from '@ember/service';
 import { DateTime } from 'luxon';
 import { findBy, sortBy } from 'ilios-common/utils/array-helpers';
-import scrollIntoView from 'scroll-into-view';
 
 export default class DashboardCoursesCalendarFilterComponent extends Component {
   @service dataLoader;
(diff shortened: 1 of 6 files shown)


michaelchadwick:
This one is a bit in-depth, so I'm working on properly reviewing it!

packages/ilios-common/addon/components/dashboard/courses-calendar-filter.js
```diff
@@ -5,7 +5,6 @@ import { action } from '@ember/object';
 import { service } from '@ember/service';
 import { DateTime } from 'luxon';
 import { findBy, sortBy } from 'ilios-common/utils/array-helpers';
-import scrollIntoView from 'scroll-into-view';
```
michaelchadwick, line 8: So, no more scrolling at all?

packages/ilios-common/addon/components/dashboard/courses-calendar-filter.js
```diff
@@ -95,9 +95,11 @@ export default class DashboardCoursesCalendarFilterComponent extends Component {
...
   @action
   scrollToLastYear(element, [year]) {
-    if (year === this.academicYear - 1) {
-      scrollIntoView(element.parentElement.parentElement, {
-        align: { top: 0 },
+    if (year === this.defaultExpandedYear) {
+      const { offsetTop } = element;
+      element.parentElement.scrollTo({
+        top: offsetTop,
+        behavior: 'instant',
```
michaelchadwick, line 102: Is the screen meant to actually "jump-scroll"? Because it doesn't do any scrolling at all now.

michaelchadwick:
@jrjohnson I tried setting my system time back to 2023 to test how the expand/collapse thing might change, but `demo.iliosproject.org` throws a cert error. Any other way to properly test that?

packages/ilios-common/addon/components/dashboard/courses-calendar-filter.js
```diff
@@ -95,9 +95,11 @@ export default class DashboardCoursesCalendarFilterComponent extends Component {
 
   @action
   scrollToLastYear(element, [year]) {
```
michaelchadwick, line 97: The method `scrollToLastYear` sounds like "go to 2023, if it is 2024". Should this not be `scrollToLatestYear` or `scrollToMostRecentYear` or `scrollToCurrentYear`?

packages/ilios-common/addon/components/dashboard/courses-calendar-filter.js
jrjohnson: Not using this helper anymore, just doing the scroll using the [scrollTo](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo) API on the element.

packages/ilios-common/addon/components/dashboard/courses-calendar-filter.js
jrjohnson: Correct and great question. The UI doesn't scroll, instead what happens is the most current academic year is focused at the top of the filter box. This is hard to test. What I did was create courses in several future years, then you will see that those are above the current year in the view and you have to scroll up to reach them. The UI doesn't move around though, it just starts in the middle.

michaelchadwick (approved):
For king and country

PR #8146, opened 2024-09-20
Enable Coverage for All Apps

jrjohnson:
This is now working with embroider, with come extra configuration, and we can enable it and run it in CI.


jrjohnson:
I'm gong to call this step one and ready. Coverage for a PR can now be downloaded from the action artifacts. Will want to hook it up to code climate and possible sonarqube as well, but that can be step two.

michaelchadwick (approved):
Confirmed goodly

PR #8147, opened 2024-09-20
Change the UI Test Label Process

jrjohnson:
I think the "ready to merge" label was a bit off, we don't need to add any more automation to this process. It is working well as is today. Updating the label to "run ui tests" is more accurate.

I also removed the test to ensure the label is present, it's just annoying. We require the ui tests that are run before merging so it will need to get added eventually anyway.


stopfstedt (approved):
i'm on board with this.

PR #8152, opened 2024-09-23
Replace Some Render Modifiers in Courses

jrjohnson:
Refs ilios/ilios#5374


packages/test-app/tests/integration/components/detail-competencies-test.js
michaelchadwick, line 1: Is there a reason to make a `hooks.beforeEach()` call when there is only one test? Or is this just future-proofing/good convention?

jrjohnson: I debated this. I ended up going with the hook because that tends to be our convention and where you'd look for this. Plus maybe we'll add some more tests.

michaelchadwick (approved):
Death to render modifiers: initiate.

PR #8153, opened 2024-09-23
Upload Coverage Info to Codeclimate

jrjohnson:
We're generating code coverage data in our tests, let's upload it somewhere and get some useful reports.

Status:
- [x] We need to await a new release of ember-cli-code-coverage which will have the option needed to configure `projectRoot` and hopefully prefix the Istanbul lcov output with the right information to connect the coverage to code climate (where the monorepo is all that is seen).

It's Alive! (you can see the status in the checks)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,3 +148,24 @@ jobs:
         with:
           args: >
             -Dsonar.projectKey=ilios_frontend_27a24e39-b0ec-405a-91a6-b6392ffcdd30
+
+  upload-coverage:
+    name: Upload Coverage
+    runs-on: ubuntu-latest
+    needs: [test]
+    if: always()
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/download-artifact@v4
+        with:
+          name: coverage-frontend
+          path: coverage
+      - run: ls -lh
+      - run: ls -lh coverage
+      - name: Publish code coverage
+        uses: paambaati/codeclimate-action@v9.0.0
+        env:
+          CC_TEST_REPORTER_ID: 8c510ad3aa4b1a2a3d504dfdbcc5605e7966c019dc1e9b68a815de50b946ebc6
+        with:
+          coverageLocations: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -148,3 +148,24 @@ jobs:
...
+      - uses: actions/download-artifact@v4
+        with:
+          name: coverage-frontend
+          path: coverage
+      - run: ls -lh
+      - run: ls -lh coverage
+      - name: Publish code coverage
+        uses: paambaati/codeclimate-action@v9.0.0
+        env:
+          CC_TEST_REPORTER_ID: 8c510ad3aa4b1a2a3d504dfdbcc5605e7966c019dc1e9b68a815de50b946ebc6
```
michaelchadwick, line 168: Why not use a Github secret?

```
env:
  CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
```

jrjohnson: That won't show coverage stats on pull requests made from forks because Github Secrets is super annoying and doesn't have a good way to accomplish this very common task. We could move to the `pull_request_target` runner, but it has its own limitations and annoyances.

michaelchadwick: Doesn't it just mean that anyone who forks it would need to create a Github Secret, as well? Is there a way to make it so it only runs this step _if_ a key exists, and otherwise skips it? I know the Code Climate Coverage key is not a nuke code in terms of importance, but it feels bad to commit it to a repo, regardless.

jrjohnson: It's a compromise for sure. Here it's much less of an issue because the [code climate docs](https://docs.codeclimate.com/docs/github-actions-test-coverage) recommend this approach, but there are other places where Github not having any kind of reasonable solution for this problem is a giant pain.

michaelchadwick: Fair enough.

michaelchadwick (approved):
Approval: granted.

PR #8156, opened 2024-09-24
Updated our Browser Test Matrix

jrjohnson:
With the release of Safari 18 I've bumped all of our browser test versions to our minimum supported 17.x. I also bumped the device for android as it might by a bit faster, but this doesn't change the browser it runs on which is evergreen.


michaelchadwick (changes requested):
Why bump the macOS, iPhone/iOS, and AndroidOS versions to `latest - 1`, but Google Pixel is bumped to `latest`?

jrjohnson:
Good question with an annoying answer - ChromeAndroid is evergreen so our target of `last 3 ChromeAndroid versions` all points to the same version [128](https://browsersl.ist/#q=last+3+ChromeAndroid+versions) so we don't need to test on anything but the latest (I'm not sure why this is, it's weird to me). The way browser stack works we don't declare the version of safari we want, instead we have to declare a device that has that version. So to get our target of [safari 17](https://browsersl.ist/#q=last+2+iOS+major+versions) we have to declare the last OS instead of the current one.

michaelchadwick (approved):
Merge sequence: initiate.

PR #8159, opened 2024-09-25
Hide the Version in Percy Snapshots

jrjohnson:
This area of the page changes when we do a release and can cause a lot of instability and false positives. Let's hide it entirely from percy.


michaelchadwick (approved):
Woot!

PR #8116, opened 2024-09-04
Remove unneeded ilios common slices

michaelchadwick:
References ilios/ilios#5660

Went through the `packages/ilios-common` directory and removed a bunch of `slice()`s. Left some I found because they had a `.sort()` or `[1]` or similar access thing that I know has thrown off our tests in the past.

PR #8160, opened 2024-09-25
Fix course LM search bug

michaelchadwick:
Fixes ilios/ilios#5727

Removing two `.slice()` calls in https://github.com/ilios/frontend/pull/8116 from LM search created this bug.

Also added two more integration tests for the "Show More" button's existence depending on how many results are returned.


jrjohnson (approved):
New tests are always the best tests!

PR #8164, opened 2024-09-27
fix editable headers with extra elements

michaelchadwick:
Fixes ilios/ilios#5730

The new `<FadeText>` component needs its `display-text-wrapper` to `display: inline-block` to account for other elements.


jrjohnson (approved):
Nice find! I was sure I broke this 😁

PR #8170, opened 2024-09-30
Reports->Terms->Program Year title fixed

michaelchadwick:
Fixes ilios/ilios#4934

### DIFF
--- a/packages/frontend/app/services/reporting.js
+++ b/packages/frontend/app/services/reporting.js
@@ -102,10 +102,15 @@ export default class ReportingService extends Service {
       const objectKey = objectTranslations[prepositionalObject];
       const objectTranslation = this.intl.t(objectKey);
       let object;
+
       if (model === 'user') {
         object = record.fullName;
       } else if (model === 'mesh-descriptor') {
         object = record.name;
+      } else if (model === 'program-year') {
+        const program = await record.program;
+        const title = program.title;
+        object = `${record.classOfYear} ${title}`;
       } else {
         object = record.title;
       }
--- a/packages/frontend/app/services/reporting.js
+++ b/packages/frontend/app/services/reporting.js
@@ -110,7 +110,8 @@ export default class ReportingService extends Service {
       } else if (model === 'program-year') {
         const program = await record.program;
         const title = program.title;
-        object = `${record.classOfYear} ${title}`;
+        const classOfYear = await record.getClassOfYear();
+        object = `${classOfYear} ${title}`;
       } else {
         object = record.title;
       }
--- a/packages/frontend/app/services/reporting.js
+++ b/packages/frontend/app/services/reporting.js
@@ -122,6 +122,9 @@ export default class ReportingService extends Service {
           'academicYearCrossesCalendarYearBoundaries',
         );
         year = crosses ? `(${record.year} - ${record.year + 1})` : `(${record.year})`;
+      } else if (model === 'program-year') {
+        const classOfYear = await record.getClassOfYear();
+        year = `(${classOfYear})`;
       }
 
       return {
--- a/packages/frontend/app/services/reporting.js
+++ b/packages/frontend/app/services/reporting.js
@@ -122,9 +122,6 @@ export default class ReportingService extends Service {
           'academicYearCrossesCalendarYearBoundaries',
         );
         year = crosses ? `(${record.year} - ${record.year + 1})` : `(${record.year})`;
-      } else if (model === 'program-year') {
-        const classOfYear = await record.getClassOfYear();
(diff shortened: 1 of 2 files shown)


michaelchadwick:
For some reason, the "year" part of the auto-generated report title only loads properly once the report has been accessed once.

To reproduce:
1. Go to `/reports`
2. Create a report like the following:
Screenshot 2024-10-09 at 4 21 37 PM
3. Save report
4. There should be a saved report called `All Terms for 2018 Doctor of Medicine in Medicine`
5. Reload site
6. Notice it now says `All Terms for Doctor of Medicine in Medicine`
7. Click on report
8. Click on "Back to Reports" or anywhere else on the site, really
9. Notice how it says the correct title again

As long as you don't reload/refresh the site, it says the correct report title. Is this not being saved in the database somewhere? If so, then it isn't loading something correctly.

jrjohnson (changes requested):
This needs a test.

jrjohnson:
This is a reactivity mess. Basically you just shouldn't use `classOfYear` and we should probably deprecated and remove it. What is happening is that `programYear.classOfYear` seems like a property, but it's actually doing some async behind the scenes. When you access it when the program isn't loaded yet it returns an empty string. What usually happens is that the program then loads and whatever getter accessed the value updates. That's a mouthful and hard to type. In this case there isn't a getter so Ember's auto tracking isn't happening, it's being resolved into this object as an empty string. The next time you access it if works because the program is resolved in the `classOfYear` getter.

I'll add a line comment for a solution here, but let's connect live and talk about this because it's a gotcha that needs discussion and then removal.

packages/frontend/app/services/reporting.js
```diff
@@ -102,10 +102,15 @@ export default class ReportingService extends Service {
...
       let object;
+
       if (model === 'user') {
         object = record.fullName;
       } else if (model === 'mesh-descriptor') {
         object = record.name;
+      } else if (model === 'program-year') {
+        const program = await record.program;
+        const title = program.title;
+        object = `${record.classOfYear} ${title}`;
```
jrjohnson, line 113:
```suggestion
        const classOfYear = await record.getClassOfYear();
        object = `${classOfYear} ${title}`;
```

jrjohnson (changes requested):
And for sure needs some tests to make sure it doesn't recur again.

jrjohnson (approved):

packages/frontend/app/services/reporting.js
```diff
@@ -116,6 +122,9 @@ export default class ReportingService extends Service {
           'academicYearCrossesCalendarYearBoundaries',
         );
         year = crosses ? `(${record.year} - ${record.year + 1})` : `(${record.year})`;
+      } else if (model === 'program-year') {
```
jrjohnson, line 125: I think this isn't needed. 
 Creating a report for all terms associated with a program year results in:

> This report shows all Terms associated with Program Year "2028 Doctor of Medicine - BRIDGES" **(2028)** in Medicine.

Which puts an extra year at the end, but the program year is already part of the description.

jrjohnson (changes requested):
I may be wrong, but I think the extra description is not needed.

packages/frontend/app/services/reporting.js
michaelchadwick: I agree that the duplication isn't necessary. The reason it's there is because the translation file uses a format that requires a `(year)` argument and it was adding a blank space otherwise.

https://github.com/ilios/frontend/blob/master/packages/frontend/translations/en-us.yaml#L345

If we don't want the extra `(year)` then wouldn't we need a new translation term solely for this case?

jrjohnson: For sure don't want the extra year just for that. I think we can live with the space, but I'd be pretty surprised if it wasn't possible to eliminate it in the message if the value wasn't set. Might be worth poking into the ICU message syntax docs and what is in https://ember-intl.github.io/ember-intl/ to see if that is possible.

michaelchadwick: OK, I see where this got "hacked" in another test without creating a new translation or adding duplicate info: https://github.com/ilios/frontend/blob/master/packages/frontend/tests/unit/services/reporting-test.js#L150

You may notice the double space before `in ${school.title}.` that I just did ;P

Not my favorite way to solve this, but this PR is getting long in the tooth so I'm just going to copy that.

michaelchadwick: The title that gets produced naturally reduces the double space to a single space, anyway, because HTML.

jrjohnson (approved):

PR #8174, opened 2024-10-03
tinkering with fractal page objects

stopfstedt:
throwaway code - i'm just trying this out as a potential replacement for  ember-cli-page-objects.

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -110,6 +110,7 @@
     "eslint-plugin-qunit": "^8.1.1",
     "eslint-plugin-yml": "^1.14.0",
     "file-saver": "^2.0.5",
+    "fractal-page-object": "^1.0.0",
     "ilios-common": "workspace:*",
     "loader.js": "^4.7.0",
     "luxon": "^3.5.0",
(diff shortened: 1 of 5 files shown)


jrjohnson:
I'm a big fan of this library and the simplification removes a lot of the things we don't use in our existing page object library. Long term it's much more maintainable and, IIRC, one of the drivers for it is that the author was putting a lot of effort into ember-cli-page-object that felt wasted. It's a clearly a pretty big lift to make the change though. We could possible code mod a chunk of this, but I'm guessing a pretty significant amount of manual work will be required too.

stopfstedt:
> The `fractal-page-objects` addon seems quite similar in syntax and usage to `ember-cli-page-objects`. Do the more advanced uses, like using the `text()` or `clickable()`, work the same?

it's all rather bare bones - the helpers have to come `qunit-dom` since fractal doesn't provide any.

stopfstedt:
also, something isn't right with fractal in ilios tests. 

consider this, for example:

`assert.dom(component.firstButton.element).hasNoClass('active');`

the `.element` shouldn't be part of that invocation, according to [the docs](https://github.com/bendemboski/fractal-page-object?tab=readme-ov-file#integrating-with-qunit-dom-and-embertest-helpers), this should be possible.

`assert.dom(component.firstButton).hasNoClass('active');`

however, this throws a test-breaking error.


i tried fractal out in a vanilla ember app, and `.element` was not necessary there. 

to be further explored.

jrjohnson:
The element dom descriptors RFC is fairly new. Maybe we have an old version of something?

stopfstedt:
i'm tabling this, hopefully forever.

PR #8176, opened 2024-10-03
improve New User password validation

michaelchadwick:
Fixes ilios/ilios#4303

Adds a 5-character length validation to new passwords. Potential addition to this PR would be to carry over the password strength logic from `<UserProfileBio>`. However, to reduce duplication, maybe it should be refactored to a service?


packages/frontend/app/components/password-validator.hbs
```diff
@@ -0,0 +1,59 @@
...
+      <input
+        id="password-{{templateId}}"
+        type="text"
+        value={{this.password}}
+        {{on "input" (pick "target.value" this.setPassword)}}
+        {{on "keyup" (queue (fn this.addErrorDisplayFor "password"))}}
+        {{on "keyup" this.keyboard}}
+        data-test-password-input
+      />
+      {{#if (this.hasErrorForPassword)}}
```
stopfstedt, line 32:
```suggestion
     {{#if this.hasErrorForPassword}}
```

michaelchadwick: This worked to fix the actual app, but tests still don't show the error message :-/
Screenshot 2024-11-12 at 11 09 57 AM

michaelchadwick: @stopfstedt OK, so the error message appears in the Integration test for `PasswordValidator`, but when it's used on the `NewUser` component, it does not. Seems it's being swallowed when trying to move upwards, but not sure how to fix.

michaelchadwick:
There was too much updating in order to get changes to work with the new `.gjs` model, so scrapped this. Superseded by #8565

PR #8177, opened 2024-10-04
fix New User refresh bug

michaelchadwick:
Fixes ilios/ilios#5742

Changed code to wait for user model to load before creating school dropdown


stopfstedt (approved):
LGTM

PR #8179, opened 2024-10-04
Remove UserMenu component render modifier

michaelchadwick:
Refs ilios/ilios#5374

Also fixed a bug where if you keyboard tabbed over to the user menu and pressed UpArrow, it would throw a console error: `Uncaught TypeError: Cannot read properties of undefined (reading 'previousElementSibling')`

### DIFF
--- a/packages/frontend/app/components/user-menu.js
+++ b/packages/frontend/app/components/user-menu.js
@@ -1,5 +1,4 @@
 import Component from '@glimmer/component';
-import { schedule } from '@ember/runloop';
 import { service } from '@ember/service';
 import { cached, tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
@@ -9,7 +8,6 @@ export default class UserMenuComponent extends Component {
   @service intl;
   @service currentUser;
   @tracked isOpen = false;
-  @tracked element;
 
   userModel = new TrackedAsyncData(this.currentUser.getModel());
 
@@ -49,26 +47,25 @@ export default class UserMenuComponent extends Component {
     return true;
   }
 
-  handleArrowDown(evt, item) {
-    if (evt.target.tagName.toLowerCase() === 'button') {
+  handleArrowDown(event, item) {
+    if (event.target.tagName.toLowerCase() === 'button') {
       this.isOpen = true;
     } else {
       if (item.nextElementSibling) {
         item.nextElementSibling.querySelector('a').focus();
       } else {
-        // eslint-disable-next-line ember/no-runloop
-        schedule('afterRender', () => {
-          this.element.querySelector('.menu li:nth-of-type(1) a').focus();
-        });
+        item.parentElement.firstElementChild.querySelector('a').focus();
       }
     }
   }
 
   handleArrowUp(item) {
-    if (item.previousElementSibling) {
-      item.previousElementSibling.querySelector('a').focus();
-    } else {
-      this.element.querySelector('.menu li:last-of-type a').focus();
+    if (item) {
+      if (item.previousElementSibling) {
+        item.previousElementSibling.querySelector('a').focus();
+      } else {
+        item.parentElement.lastElementChild.querySelector('a').focus();
+      }
     }
   }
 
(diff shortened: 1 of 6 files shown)


michaelchadwick:
Added keyboard navigation testing to `UserMenu` component. Needed to add a slight delay in both the component and integration test to properly set focus and test for it.

packages/frontend/app/components/user-menu.js
```diff
@@ -21,6 +19,12 @@ export default class UserMenuComponent extends Component {
   @action
   toggleMenu() {
     this.isOpen = !this.isOpen;
+
+    if (this.isOpen) {
+      setTimeout(() => {
```
jrjohnson, line 24: I'm guessing this is the same thing the `afterRender` runloop was handling as well. A timeout here is doing the same thing, moving execution out of the current thread, waiting for all things to finish (which includes the rendering) and then doing it on the next pass through. I'd prefer to either see the runloop used here, as that's what it's really for, or convert this into a `task` and then use the `await timeout(1)` syntax. That's doing the exact same thing as your timeout here, but is more idiomatic for our codebase.

packages/frontend/tests/integration/components/user-menu-test.js
```diff
@@ -11,6 +11,9 @@ module('Integration | Component | user-menu', function (hooks) {
   setupRenderingTest(hooks);
   setupMirage(hooks);
 
+  // My Profile and Logout
+  const linkCount = 2;
```
jrjohnson, line 15: Nice change for clarity!

packages/frontend/tests/integration/components/user-menu-test.js
```diff
@@ -56,8 +59,62 @@ module('Integration | Component | user-menu', function (hooks) {
     await render(hbs`<UserMenu />`);
 
     await component.toggle.down();
-    assert.strictEqual(component.links.length, 2);
+    assert.strictEqual(component.links.length, linkCount);
+    await component.toggle.click();
+    assert.strictEqual(component.links.length, 0);
+  });
+
+  test('keyboard navigation', async function (assert) {
```
jrjohnson, line 67: 🔥

packages/frontend/app/components/user-menu.js
```diff
@@ -49,30 +53,29 @@ export default class UserMenuComponent extends Component {
...
       }
     }
   }
 
   handleArrowUp(item) {
-    if (item.previousElementSibling) {
-      item.previousElementSibling.querySelector('a').focus();
-    } else {
-      this.element.querySelector('.menu li:last-of-type a').focus();
+    if (item) {
```
jrjohnson, line 73: I'm ok with this. Could also use the `item?.previousIndex...` optional chaining syntax which would do the same thing with slightly less indentation.

jrjohnson (changes requested):

packages/frontend/app/components/user-menu.js
michaelchadwick: I didn't like it when I wrote it, so thank you for reminding me of the `?` optional thing!

packages/frontend/app/components/user-menu.js
michaelchadwick: Initially, I just fixed this by going the runloop path, but we have to use an `eslint-disable` to do that, which seems bad and only makes the code worse. So, I turned it into a `task` as you suggested.

packages/frontend/app/components/user-menu.js
```diff
@@ -49,30 +57,24 @@ export default class UserMenuComponent extends Component {
...
     }
   }
 
   handleArrowUp(item) {
-    if (item.previousElementSibling) {
+    if (item?.previousElementSibling) {
       item.previousElementSibling.querySelector('a').focus();
     } else {
-      this.element.querySelector('.menu li:last-of-type a').focus();
+      item.parentElement.lastElementChild.querySelector('a').focus();
```
jrjohnson, line 77:
```suggestion
      item?.parentElement.lastElementChild.querySelector('a').focus();
```

jrjohnson (commented):
I *think* there needs to be another check on item?

jrjohnson (approved):

PR #8187, opened 2024-10-16
main-list mixin consistency

michaelchadwick:
Fixes ilios/ilios#5759

This PR got a whole lot more complicated than I thought, but it was needed to fix the issue.

The `mixins/main-list.scss` mixin file is in `frontend`, but had a mixin that needed to be called by `ilios-common`, so...I moved it! :-O This, of course, broke a lot of `@import`s, so I had to fix them. In the end, I was able to make the "you made a new thing" template chunk's class more consistent.

No idea why changing `.save-result` -> `.saved-result` (to be consistent with every other instance of this kind of thing) in Courses->Sessions breaks that route (gives me a Zoinks!), so I changed it back for now.

### DIFF
--- a/packages/frontend/app/styles/components/assign-students/manager.scss
+++ b/packages/frontend/app/styles/components/assign-students/manager.scss
@@ -1,5 +1,4 @@
-@use "../../ilios-common/mixins" as cm;
-@use "../../mixins" as m;
+@use "../../ilios-common/mixins" as m;
 
 .assign-students-manager {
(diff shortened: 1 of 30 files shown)


jrjohnson (approved):
This may be the most confusing PR I've ever reviewed! It's so hard to keep track of the `m` and the `cm`. I agree, got out of hand, but looks right to me. Will let Percy find any issues

PR #8188, opened 2024-10-17
icon spacing improvement

michaelchadwick:
Fixes ilios/ilios#5696

There may be more spacing fixing that needs to happen, but this fixes the issue and adds a little extra.

### DIFF
--- a/packages/ilios-common/app/styles/ilios-common/components/awesome-icon.scss
+++ b/packages/ilios-common/app/styles/ilios-common/components/awesome-icon.scss
@@ -14,10 +14,6 @@
   line-height: 1; /* stylelint-disable-line property-disallowed-list */
   text-rendering: auto;
 
-  &.fa-download {
-    margin-left: 0.1rem;
-  }
-
   &.spin {
     animation-name: awesome-spin;
     animation-delay: 0s;
--- a/packages/ilios-common/app/styles/ilios-common/components/awesome-icon.scss
+++ b/packages/ilios-common/app/styles/ilios-common/components/awesome-icon.scss
@@ -14,6 +14,10 @@
   line-height: 1; /* stylelint-disable-line property-disallowed-list */
   text-rendering: auto;
 
+  &.fa-download {
+    margin-left: 0.1rem;
+  }
+
   &.spin {
     animation-name: awesome-spin;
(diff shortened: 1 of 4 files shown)


packages/ilios-common/app/styles/ilios-common/components/awesome-icon.scss
```diff
@@ -14,6 +14,10 @@
   line-height: 1; /* stylelint-disable-line property-disallowed-list */
   text-rendering: auto;
 
+  &.fa-download {
```
jrjohnson, line 17: I'm hesitant about the placement of this, I think it should go in the `lm-icons` file instead of here as that is scoped to just that component, but this would change every instance of that icon. I think we only have one place right now, but it would be very confusing later if someone adds a second one.

jrjohnson (changes requested):

packages/ilios-common/app/styles/ilios-common/components/awesome-icon.scss
michaelchadwick: It can't go in the `lm-icons.scss` file because the `svg.awesome-icon.fa-download` is not a child of `lm-icons`, but instead a sister element:

```
<li class="week-glance-learning-material-list-item" data-test-learning-material="">
  ...

  <span class="lm-icons" data-test-lm-icons=""> ... </span>
  <a>
    <svg class="awesome-icon fa-download"> ... </svg>
  </a>
</li>
```

If we want it to be more component-specific, then we'd need to put a `.fa-download` selector in multiple components' SCSS files (at least `SingleEvent` and `WeeklyCalendarEvent`).

jrjohnson: Annoying, but yes please. Put it in each of those files.

jrjohnson: Or create a component with just the icon, slap a class on it, and then add rules for that class. Probably more annoying, but also a valid approach to scope the css here.

michaelchadwick: OK, I did the first thing.

jrjohnson (approved):

PR #8189, opened 2024-10-17
Remove Programs render modifiers

michaelchadwick:
Refs ilios/ilios#5374

### DIFF
--- a/packages/frontend/app/components/program-year/objectives.js
+++ b/packages/frontend/app/components/program-year/objectives.js
@@ -11,7 +11,6 @@ export default class ProgramYearObjectivesComponent extends Component {
 
   @tracked newObjectiveEditorOn = false;
   @tracked newObjectiveTitle;
-  @tracked objectiveCount;
 
   get showCollapsible() {
     return this.hasObjectives && !this.isManaging;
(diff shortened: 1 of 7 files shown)


packages/frontend/app/components/program-year/objectives.js
```diff
@@ -21,11 +21,16 @@ export default class ProgramYearObjectivesComponent extends Component {
...
 
-  @action
-  load(element, [programYear]) {
-    this.objectiveCount = programYear.hasMany('programYearObjectives').ids().length;
+  constructor() {
+    super(...arguments);
+    this.objectiveCount = this.args.programYear.hasMany('programYearObjectives').ids().length;
   }
 
+  // @action
```
stopfstedt, line 29: kruft, please remove.

packages/frontend/app/components/program-year/objectives.js
```diff
@@ -21,11 +21,16 @@ export default class ProgramYearObjectivesComponent extends Component {
     return this.objectiveCount > 0;
   }
 
-  @action
-  load(element, [programYear]) {
-    this.objectiveCount = programYear.hasMany('programYearObjectives').ids().length;
+  constructor() {
```
stopfstedt, line 24: this can be further simplified. we don't need to set `objectiveCount` as component attribute.

instead, a getter would suffice here:

```suggestion
  get objectivesCount() {
    return this.args.programYear.hasMany('programYearObjectives').ids().length;
  }
```

then, remove the tracked property by the same name further up in the class, and we're good to go.

stopfstedt (changes requested):

packages/frontend/app/components/program-year/objectives.js
```diff
@@ -21,9 +20,8 @@ export default class ProgramYearObjectivesComponent extends Component {
     return this.objectiveCount > 0;
   }
 
-  @action
-  load(element, [programYear]) {
-    this.objectiveCount = programYear.hasMany('programYearObjectives').ids().length;
+  get objectivesCount() {
```
stopfstedt, line 23: typo here, should be `objectiveCount`

```suggestion
  get objectiveCount() {
```

stopfstedt (dismissed):

packages/frontend/app/components/program-year/objectives.js
stopfstedt: @michaelchadwick my bad on this - i fat fingered the getter name in the suggestion. sorry about this. see new comment below.

jrjohnson (approved):

PR #8190, opened 2024-10-17
added a maxlength attribute option and default for SearchBox component

michaelchadwick:
Fixes ilios/ilios#3635

Added a `@maxlength` property to the `<SearchBox>` component and made the default 6000. This fixes any and all instances I found that have some 6000+ upper limit that makes the API return with a 414.


stopfstedt (approved):
LGTM

PR #8191, opened 2024-10-18
Add Code Coverage from Test App

jrjohnson:
We're only seeing coverage out of the frontend currently, let's combine it with what we get from the test app and send them both to codeclimate.


jrjohnson:
The upload works, but it's ultimately useless as we're not getting good stats out of the coverage. It only knows about the files within the test-app. Will need to keep an eye on the embroider addon issue tracker to determine when this might work.

jrjohnson:
Going to close this for now. We need lots of stuff upstream and I don't see a ton of value in trying to push that work forward.

PR #8192, opened 2024-10-18
adds unpublished icon to SingleEvent instances

michaelchadwick:
Fixes ilios/ilios#3473

Single Events need to show if they are unpublished just like when they're viewed on the calendar, so I added the missing icon. I also made sure to put missing mouseover text, and fixed up some tests.


packages/ilios-common/addon/components/single-event.hbs
```diff
@@ -1,26 +1,42 @@
...
-          {{#if (and this.sessionRouteExists this.canViewSession)}}
-            <LinkTo
-              @route="session"
-              @models={{array @event.course @event.session}}
-            >{{@event.name}}</LinkTo>
-          {{else}}
-            {{@event.name}}
-          {{/if}}
-        </em>
+        {{#if (not this.isPublished)}}
```
stopfstedt, line 13: should be checking `isPublished` on `@event`, not `this`. 

```suggestion
        {{#if (not @event.isPublished)}}
```

packages/ilios-common/addon/components/single-event.hbs
```diff
@@ -1,26 +1,42 @@
...
+          <FaIcon
+            @icon="file-signature"
+            @title={{t "general.notPublished"}}
+            data-test-draft-icon
+          />
+        {{else if this.isScheduled}}
```
stopfstedt, line 19: should be checking `isScheduled` on `@event`, not `this`.

```suggestion
        {{else if @event.isScheduled}}
```

stopfstedt (changes requested):
apart from the two issues mentioned in `single-event.hbs`, this looks good to me.

however, it's missing test coverage on these icons.

please add this appropriate tests.

for reference, i'd like to point you at existing page-object attributes and integration tests. please see https://github.com/ilios/frontend/blob/4718d3749813d2a3b96d063915db31febe7cd4f8/packages/test-app/tests/integration/components/weekly-calendar-event-test.js#L418-L486

michaelchadwick:
Thanks for the review, @stopfstedt! I'll get back to this soon.

stopfstedt (approved):
LGTM

PR #8203, opened 2024-10-30
Replace get-errors-for helper

stopfstedt:

i couldn't figure out how to salvage this `Resource`-based helper, so I'm attempting to side-step the issue by shifting its functionality into the `ValidationError` component.

this changes the interface of the `ValidationError` component.

from:

```hbs
<ValidationError @errors={{get-errors-for this.foo}} />
```

to:

```hbs
<ValidationError @validatable={{this}} @property="foo" />
```


michaelchadwick (approved):
Approval: granted

packages/test-app/tests/integration/components/validation-error-test.js
```diff
@@ -7,27 +7,43 @@ module('Integration | Component | validation-error', function (hooks) {
...
   });
 
   test('it renders with errors', async function (assert) {
-    await render(hbs`<ValidationError @errors={{array 'test 1' 'test 2'}} />`);
+    assert.expect(2);
 
-    assert.dom(this.element).hasText('test 1 test 2');
-  });
-
-  test('it responds to changes', async function (assert) {
```
jrjohnson, line 21: I think this test got left out and I'm worried about how this implementation will respond to new errors. Does it know to call the promise again, or will it never update?

jrjohnson (changes requested):

packages/test-app/tests/integration/components/validation-error-test.js
michaelchadwick: I'm actually glad you took a look at this PR, as I'm not the expert on this functionality :)

stopfstedt: we have plenty of indirect test coverage on state changes of the validated attribute, via integration tests for components that use the `<ValidationError>` component.

transforming/keeping this test-case would require a dummy object that's responsive. that seems like a lot of work for very little direct gain here.

jrjohnson: Summary of conversation: this isn't easily testable, but there is plenty of coverage for it being correctly reactive.

jrjohnson (approved):

PR #8206, opened 2024-10-31
Stop Autoupdate Check from Failure

jrjohnson:
I hate the big red x on this, I think this will help.


jrjohnson:
Before screen shot added. I can't do an after screenshot because this check only runs on dependabot PRs, we'll have to see if I flubbed it after merging.

PR #8207, opened 2024-11-01
Replace Pixel Device in Browser Tests

jrjohnson:
For some reason the pixel 9 has disappeared from the browserstack test API for Android 14. It seems like the 8 still works though so let's use that.


michaelchadwick (approved):
Not nine this time
Cuz eight is great

stopfstedt (approved):

PR #8213, opened 2024-11-04
replaces has-error-in Resource

stopfstedt:


michaelchadwick (approved):
@jrjohnson probably needs to look this over for a more detailed analysis, but everything looks in order (I at least manually tested the ProgramYears->Objective character limit and it works)

jrjohnson (approved):
I still don't understand how this is working. It just seems like it shouldn't auto track. It does though. Verbose, but works well and removes the helper. Approved.

PR #8214, opened 2024-11-05
removes persistance of user-context selections from calendar filters

stopfstedt:
fixes ilios/ilios#5725


jrjohnson (approved):
Code is good, will wait to hear that we really want to remove this feature. I was under the impression it was needed.

michaelchadwick (approved):
I'm actually surprised that Ilios _only_ used localStorage for this one thing, but when you have user accounts, there's far less reason to use it. Could we not move this feature into the user account itself?

dartajax:
I tested using the original case that the https://github.com/ilios/ilios/issues/5725 discussed - it is pretty unlikely that standard users would do things the way I do them - switching users for documentation context I mean. Anyway, the original situation reported by me has been resolved.

dartajax (dismissed):
As far as the original issue is concerned as referenced in https://github.com/ilios/ilios/issues/5725, this fixes the issue and that being the case can be approved.

michaelchadwick (approved):

PR #8215, opened 2024-11-06
synchronize Course/Session objective description/parentObjectives FadeText status

michaelchadwick:
Fixes ilios/ilios#5645

Whew, this was a long time coming for me. Basically, be it Courses or Sessions, their objectives (and their Description and Parent Objective columns) now synchronize in individual rows, so that if you expand one column, the other column expands, as well (and in the opposite direction, too). Had to learn how to send functions from parent objectives down to child objectives and all that, so fun! `ObjectiveSortManager` has to short-circuit this by having its own function that overrides the one sent by `ObjectiveListItem`.


packages/ilios-common/addon/components/course/objective-list-item.js
```diff
@@ -50,6 +51,10 @@ export default class CourseObjectiveListItemComponent extends Component {
     return this.isManagingParents || this.isManagingDescriptors || this.isManagingTerms;
   }
 
+  get fadeTextIsExpanded() {
```
jrjohnson, line 54: I think I caused this confusion because I was talking about a getter that would read both a local and a passed in property and decide which to use, but I'm not sure if this `get` is necessary. Couldn't the template just read `this.fadTextExpanded` directly?

packages/ilios-common/addon/components/editable-field.js
```diff
@@ -15,6 +15,10 @@ export default class EditableFieldComponent extends Component {
     return strippedText.length === 0;
   }
 
+  get fadeTextIsExpanded() {
```
jrjohnson, line 18: This doesn't appear to be used.

packages/test-app/tests/integration/components/fade-text-test.js
```diff
@@ -54,7 +54,13 @@ module('Integration | Component | fade-text', function (hooks) {
   test('it fades tall text given as component argument', async function (assert) {
     this.set('longHtml', this.longHtml);
 
-    await render(hbs`<FadeText @text={{this.longHtml}} />`);
+    this.set('expanded', false);
+    this.set('onExpandAll', (isExpanded) => {
+      this.set('expanded', isExpanded);
```
jrjohnson, line 59: An assertion here that the method is called as often as expected (counter or something else) would be welcome.

jrjohnson (changes requested):

jrjohnson:
Behavior looks great to me! There are some possibly artistic choices in what expands all. Clicking the expander catches everything, but clicking to edit a title or parents doesn't. I played with it for a while and I think I like this, but wanted to call it out for further testing by @dartajax.

packages/ilios-common/addon/components/course/objective-list-item.js
michaelchadwick: You're totally right. No getter needed. Just passing the property now.

packages/test-app/tests/integration/components/fade-text-test.js
michaelchadwick: Do you mean like an `assert.expect()`?

michaelchadwick: I interpreted this as putting an `assert.strictEqual(this.expanded, isExpanded)` in that `onExpandAll` definition, which then caused me to need `assert.expect()` calls :D

michaelchadwick:
@dartajax @jrjohnson I fixed the buggy icon we saw in the standup. I also fixed some icon inconsistencies I saw (per the light on dark vs dark on light styling). I also also fixed how the ObjectiveSortManager displayed so the draggable icon now sits in the center like before. I know that was for a future PR but I figured it out, so :D

michaelchadwick:
Still gotta fix the _Session_ version of these things, tho...

stopfstedt (dismissed):

michaelchadwick:
Got all the syncing stuff working, and fixed existing tests. Even without Code Climate warning me, I know new tests need to be written to account for the changes, but just want to make sure this is cool so far.

michaelchadwick:
@dartajax that's a valid visual point. I'm not sure yet how to improve the whitespace, but I will look into it.

michaelchadwick:
@dartajax that extra space should be removed now

dartajax (approved):
looks great to me

michaelchadwick:
Added Course Objective acceptance test (and improved FadeText integration test in the process). Now to work on the Session Objective acceptance test (which will be very similar, but still somewhat different).

michaelchadwick:
Session acceptance test now in and working(?). CodeClimate still complaining -_- @jrjohnson @stopfstedt any ideas?

jrjohnson:
OK to merge with failing code-climate test here.

dartajax (approved):
this looks good to me

dartajax (approved):

jrjohnson (approved):

PR #8217, opened 2024-11-07
re-route users without elevated privileges from course admin pages to error page.

stopfstedt:
this affects the course details page and sub-pages (like session management) that share the same root path `/courses/{id}/`, and other course-management related routes.

it's a foot in the door - i'll lock other admin areas such as instructor group/learner group management down in consecutive steps. 

refs https://github.com/ilios/ilios/issues/5721

### DIFF
--- a/packages/ilios-common/addon/routes/course/index.js
+++ b/packages/ilios-common/addon/routes/course/index.js
@@ -8,6 +8,8 @@ export default class CourseIndexRoute extends Route {
   @service store;
   @service dataLoader;
   @service session;
+  @service router;
+  @service currentUser;
 
   canCreateSession = false;
   canUpdateCourse = false;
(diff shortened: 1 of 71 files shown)


packages/ilios-common/addon/routes/course/index.js
```diff
@@ -19,6 +21,9 @@ export default class CourseIndexRoute extends Route {
 
   beforeModel(transition) {
     this.session.requireAuthentication(transition, 'login');
+    if (!this.currentUser.performsNonLearnerFunction) {
```
jrjohnson, line 24: I'm pretty sure the `beforeModel` hook in the course route will fire and redirect before this one is hit making this code unnecessary. The same is true of the `this.session.requireAuthentication(transition, 'login');` so maybe better to have it in both places and be sure?

jrjohnson (approved):
Seems like a net positive, LGTM.

michaelchadwick (approved):
Checked out the Netlify preview and it seems to do what it says on the tin (at least as far as Courses go).

PR #8218, opened 2024-11-07
adds downloadable data sheets for course visualizations

stopfstedt:


stopfstedt:
> I tried all four visualization types for https://deploy-preview-8218--ilios-frontend.netlify.app/data/courses/2224 and the download buttons all worked (in Chrome and Firefox) as expected. The Objectives one, however, put raw HTML markup into the Course Objective column, which may not be what you were intending.

Thank Mike, I'll take this back to the workbench for markup stripping. Should be back with it shortly.

michaelchadwick (approved):

PR #8228, opened 2024-11-14
remember Calendar courseFilters setting across hide/show filter toggle

michaelchadwick:
Fixes ilios/ilios#3430

Removed setting that was defaulting `this.courseFilters` to true when you hid and then showed the filters. Now it remembers!


stopfstedt (approved):
LGTM

PR #8229, opened 2024-11-14
Replace SASS imports

jrjohnson:
The `@import` syntax is deprecated in sass. If I'm understanding this correctly we should use `@use` when we want the imported thing to appear in this place, but `@forward` when we're saying when the containing file is `@use`'d we want to content to go there.

Did a little bit of clicking around, seems like this works. Will need a review by @dartajax and then we'll let Percy do it's thing as well.


dartajax (approved):
click tested - percy approved - I'm gonna merge it

PR #8231, opened 2024-11-14
Reports: Remove 'Anything' option for 'Instructors' and 'MeSH Terms' subjects

michaelchadwick:
Fixes ilios/ilios#4531

These two subjects can't be used with the "Anything" object, which is the default option for all subjects. I removed the "Anything" option from the `<select>` only when those subjects are chosen, and default them to their first prepositional object instead. Lastly, I removed the errors and tests that checked for the requirement as it is no longer user-selectable.

### DIFF
--- a/packages/frontend/app/components/reports/new-subject.js
+++ b/packages/frontend/app/components/reports/new-subject.js
@@ -285,7 +285,7 @@ export default class ReportsNewSubjectComponent extends Component {
         .map((item) => item.label)
         .sort()[0];
       const firstPrepositionalObject = this.fullPrepositionalObjectList.filter(
-        (item) => item.label == firstPrepositionalObjectLabel,
+        (item) => item.label === firstPrepositionalObjectLabel,
       )[0].value;
 
       this.changePrepositionalObject(firstPrepositionalObject);


packages/frontend/app/components/reports/new-subject.js
```diff
@@ -269,18 +269,33 @@ export default class ReportsNewSubjectComponent extends Component {
...
-    this.clearErrorDisplay();
+
+    if (this.includeAnythingObject) {
+      this.changePrepositionalObject(null);
+    } else {
+      const firstPrepositionalObjectLabel = Object.values(this.prepositionalObjectList)
+        .map((item) => item.label)
+        .sort()[0];
+      const firstPrepositionalObject = this.fullPrepositionalObjectList.filter(
+        (item) => item.label == firstPrepositionalObjectLabel,
```
stopfstedt, line 288: please use triple equal here for strict comparison, a bit surprised that the linter doesn't complain here.

```suggestion
        (item) => item.label === firstPrepositionalObjectLabel,
```

stopfstedt (changes requested):
one minor issue, the rest looks good to me.

stopfstedt (approved):
LGTM

PR #8232, opened 2024-11-14
make froala link edit popup fully visible

michaelchadwick:
Fixes ilios/ilios#5826

Made sidebar lower so popups show above it. There probably should be a test for this, but I'm not sure where in the suite this kind of thing would go (i.e. external dependency popup making).


stopfstedt (approved):
click tested in FF. LGTM

image

dartajax (approved):
new z-index > old z-index - works great

PR #8236, opened 2024-11-20
Validate Session Title with Yup

jrjohnson:
First step in making a transition away from our current classic decorator valdidation approach and into something more modern and less prone to blowing up horribly when [decorator](https://github.com/tc39/proposal-decorators) support lands in the language officially.

This has some significant advantages over our current approach:
1. We can declare many smaller validation subsets if we want as each is simple an object we can manipulate
2. The API on the component is simplified (I think it can get even better, but this was a good place to stop)
3. Less magic with creating and setup as you can see it all in the class

I'm not a *huge* fan of the `setupErrorMessages` function on file load invocation as I don't really know why that works and doesn't get called on every import, I wasn't able to get something going with an initializer, but optimization there may be possible.

This seemed like a natural stopping place for comments. I can either follow up in this PR with migrations for everything else, or more likely we'll tackle them one at a time.


michaelchadwick:
How would I test a "NewSession" to see if the validation works or not?

jrjohnson:
Testing:
Visit a course, scroll to the sessions list, click the `plus` to open the new session form, have at it!

michaelchadwick:
For some reason, I was thinking of a "NewSession" as a new user session, like when you log in /facepalm

jrjohnson:
Failing test is for real, I need to swap the way it debounces, it works for user input, but the test fails because async is hard and there seems to be a ordering issue with calling it this way.

Should be a pretty minor fix, this is still ready for comments.

michaelchadwick:
First thing I notice is that the form actually seems to get submitted before the validations are run now. Click submit button, see spinner, see validations. Before it would show up before spinning. Not sure if that's better or worse, but it does incur a minor speed decrease.

jrjohnson:
>the form actually seems to get submitted before the validations are run now.

I *think* you may have been seeing the same async weirdness that was breaking in the test, I've moved some stuff around.

I also reduced the debounce because I think that 250ms was actually too long and felt janky.

michaelchadwick:
@jrjohnson Your change fixed what I was seeing :D

packages/ilios-common/addon/components/new-session.js
```diff
@@ -1,27 +1,21 @@
...
 
-@validatable
 export default class NewSessionComponent extends Component {
   @service store;
 
-  @NotBlank() @Length(3, 200) @tracked title;
+  @tracked title;
   @tracked selectedSessionTypeId;
 
-  @cached
```
michaelchadwick, line 17: I always wondered how the `@cached` decorator worked for async data. Now that it's not being used, does that mean that the validation being done is no longer cached? Is that bad?

packages/ilios-common/addon/classes/yup-validations.js
```diff
@@ -0,0 +1,147 @@
...
+import { getProperties } from '@ember/object';
+import { object, setLocale } from 'yup';
+import { restartableTask, timeout } from 'ember-concurrency';
+
+const DEBOUNCE_MS = 100;
+
+/**
+ * Started with: https://mainmatter.com/blog/2021/12/08/validations-in-ember-apps/
+ */
+export default class YupValidations {
```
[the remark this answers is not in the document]
stopfstedt: i kinda like to have this in there, for grepping purposes. will make it easier to separate from the existing implementation, which also makes use of the term validate.

once we completely replaced the old stuff, we should make a pass over this and remove Yup from the class names, should be trivial to find and replace.

packages/ilios-common/addon/components/yup-validation-message.hbs
```diff
@@ -0,0 +1,5 @@
+{{#each this.messages as |m|}}
+  <span class="validation-error-message" ...attributes>
```
stopfstedt, line 2: why the splattributes here?

packages/ilios-common/addon/classes/yup-validations.js
```diff
@@ -0,0 +1,147 @@
...
+    await timeout(DEBOUNCE_MS);
+    const rhett = await this.#validate();
+    //return the result of the validation and not the promise itself
+    return rhett;
+  });
+
+  async #validate() {
+    try {
+      await this.schema.validate(this.#validationProperties(), {
+        abortEarly: false,
```
stopfstedt, line 62: how about setting this to `true`?

`abortEarly: false`

image



`abortEarly: true`

image


the latter would be consistent with the error output we're getting from the current implementation.

stopfstedt (commented):
i'm digging this. left a few comments.

packages/ilios-common/addon/components/new-session.js
stopfstedt: The data point under validation, `title`, is at no point in an async state. but since the `hasErrorFor` function is async, we currently have to do this dance in wrapping it. 
Getting rid of all this extra code is a huge improvement for the DX, at least. IMO.

jrjohnson: The validation is still an asynchronous process (because you can do asynchronous stuff in there, we're just not doing it here), however what happens now is that we-revalidate constantly if there are errors they get stored and then we display them, they can also go away, just constant shuffle, but all the async is elsewhere. The only part to worry about is validating before actually doing anything and that's where `isValid()` comes in. When you run that and wait for results you get a settled correct state.

packages/ilios-common/addon/classes/yup-validations.js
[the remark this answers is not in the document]
jrjohnson: +1 on what @stopfstedt has here, it's an interim solution and we'll probably rename everything when we finish ripping out the non-yup stuff.

packages/ilios-common/addon/components/yup-validation-message.hbs
jrjohnson: Just in case. You might toss a different class or a `data-FOO` on there and it will get pushed into every single one. I dunno if we'll use it, I just like to add them so we don't have to re-add them later.

packages/ilios-common/addon/classes/yup-validations.js
jrjohnson: This single field validation case doesn't show it, but what happens is validation dies when *any* field is wrong. So if you had two fields with an error only the first one would show up. I'm more inclined to just take the first error off the stack, but figured we'd see how people feel about seeing two.

stopfstedt: thanks for the explanation. i now recall you mentioned this before, but it didn't really register with me at the time.

packages/ilios-common/addon/components/yup-validation-message.hbs
stopfstedt: i accept this.

stopfstedt (approved):

PR #8237, opened 2024-11-20
Reports > Start tracking showNewReportForm as queryParam

michaelchadwick:
Spending a lot of time on the Reports page, and thought this could be useful.

This is not a high priority change, but it helped me reinforce the whole "sending properties and actions downstream" concept, and it keeps track of the New Report form's visibility instead of defaulting to off every time.

### DIFF
--- a/packages/frontend/app/components/reports/list.hbs
+++ b/packages/frontend/app/components/reports/list.hbs
@@ -17,18 +17,18 @@
     </h2>
     <div class="actions">
       <ExpandCollapseButton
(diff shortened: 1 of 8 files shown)


michaelchadwick:
@stopfstedt @jrjohnson I don't know why this has one failing test. It runs fine in the browser, but the CLI and CI fail.

michaelchadwick:
Oh, it's because a test had an errant `console.log` in it. Huh.

stopfstedt (approved):

PR #8238, opened 2024-11-22
cancel old graphql requests when new ones are run

michaelchadwick:
Reports are run one of two ways via the UI:

1. Going to `/reports`, choosing items from the dropdown inputs, and hitting the "Run Report" button
2. Going to an existing report at `/reports/subjects/:id`

Either way, it does a GraphQL lookup and the page waits for it to return a value. If you use method 1, and click the button before an existing request is done, it queues it, which can lead to potential long network requests running in the background, especially if you run a long report a bunch of times for either impatience or debugging reasons.

The ideal way to fix this is to improve the searches themselves so that they don't take so long. In the meantime, making sure new requests invalidate old requests is helpful.

This can be done by changing the method from an `async func()` call to an `ember-concurrency`-style `func = task(async() {})` call, which fixes the issue in case 1 (case 2 doesn't need a fix because you can't re-run the report from `ReportsSubjectRoute`).

However, there are still two cases that aren't yet covered:

1. Going to `/reports`, choosing items from the dropdown inputs, hitting the "Run Report" button, _and then hitting the "Cancel" button_.
2. Going to an existing report at `/reports/subjects/:id`..._and then clicking the "Back to Reports" link (or simply clicking back in the browser or somewhere else in the app)_.

I think I can figure both of these out, but need more time ;P


jrjohnson (commented):
I think we talked about this in person, but it was a few months ago. Wanted to put something official. Lots of great stuff in here and overall I'm in support of this. I don't like exposing the task in the `graphql` service to the consumers. It feels like API leakage. At some point we actually need to go back and make sure we're not calling `perform()` anywhere, but on the place a task lives.

What about putting the `AbortController` and the `cancelCurrentRun` into the `graphQL` service itself and making it cancelable from outside. Possibly by returning a reference to the promise? Not sure that feels right either as it's a lot of bookkeeping for short reports, but it's a start.

michaelchadwick:
@jrjohnson 
> What about putting the `AbortController` and the `cancelCurrentRun` into the `graphQL` service itself and making it cancelable from outside. Possibly by returning a reference to the promise? Not sure that feels right either as it's a lot of bookkeeping for short reports, but it's a start.

Sounds like a great idea. I'm kind of putting the breaks on this one until I can figure out a backend piece since, as you also mentioned, this only stops it on the client side. It would be more performant, and more complete, if we could cancel the server side, as well.

michaelchadwick:
Closing as the changes to the code base since I first created this are too onerous to rebase against, so just gonna start a new PR.

PR #8242, opened 2024-11-26
Use Yup Validations in Bulk New Users

jrjohnson:
Trying this out for the first time on the frontend, it seems to make a few things simpler, which is nice.

### DIFF
--- a/packages/frontend/app/components/bulk-new-users.js
+++ b/packages/frontend/app/components/bulk-new-users.js
@@ -1,17 +1,16 @@
 import Component from '@glimmer/component';
 import { cached, tracked } from '@glimmer/tracking';
-import { getOwner } from '@ember/application';
 import { action } from '@ember/object';
-import CoreObject from '@ember/object/core';
 import { service } from '@ember/service';
 import { isPresent } from '@ember/utils';
 import { all, filter } from 'rsvp';
 import { dropTask, restartableTask } from 'ember-concurrency';
 import PapaParse from 'papaparse';
 import { DateTime } from 'luxon';
-import { validatable, Length, NotBlank, IsEmail, Custom } from 'ilios-common/decorators/validation';
 import { findById, mapBy } from 'ilios-common/utils/array-helpers';
 import { TrackedAsyncData } from 'ember-async-data';
+import YupValidations from 'ilios-common/classes/yup-validations';
+import { string } from 'yup';
 
 export default class BulkNewUsersComponent extends Component {
   @service flashMessages;
@@ -169,7 +168,7 @@ export default class BulkNewUsersComponent extends Component {
     this.primaryCohortId = id;
   }
 
-  async existingUsernames() {
+  async getExistingUserNames() {
     const authentications = await this.store.findAll('authentication');
     return mapBy(authentications, 'username').filter(Boolean);
   }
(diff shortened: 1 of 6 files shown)


packages/frontend/app/components/bulk-new-users.js
```diff
@@ -182,26 +181,29 @@ export default class BulkNewUsersComponent extends Component {
...
     this.fileUploadError = false;
+    const existingUsernames = await this.getExistingUserNames();
     return new Promise((resolve) => {
       const allowedFileTypes = ['text/plain', 'text/csv', 'text/tab-separated-values'];
       if (!allowedFileTypes.includes(file.type)) {
-        const intl = this.intl;
-        this.set('fileUploadError', true);
-        throw new Error(intl.t('general.fileTypeError', { fileType: file.type }));
+        this.fileUploadError = true;
+        throw new Error(this.intl.t('general.fileTypeError', { fileType: file.type }));
```
stopfstedt, line 189: this is an improvement over the current implementation (which throws an error as well, but it's not visible to the end user - only to be found in the browser console).

however it's not ideal:
image

packages/frontend/app/components/bulk-new-users.js
```diff
@@ -234,18 +236,12 @@ export default class BulkNewUsersComponent extends Component {
...
-    const existingUsernames = await this.existingUsernames();
-    const filledOutUsers = proposedUsers.map((obj) => {
-      obj.existingUsernames = existingUsernames;
-
-      return obj;
-    });
-    this.validUsers = await filter(filledOutUsers, async (obj) => {
-      return await obj.isValid();
+    this.validUsers = await filter(proposedUsers, async (obj) => {
+      return obj.isValid();
```
stopfstedt, line 240: `ProposedUser.isValid()` is async, so we should await for promise resolution here.

```suggestion
      return await obj.isValid();
```

packages/frontend/app/components/bulk-new-users.js
michaelchadwick: Interesting filename, that `DOOM PILG..er`. I'm guessing https://boardgamegeek.com/boardgame/373657/doom-pilgrim has been visited recently.

stopfstedt (changes requested):
validation fails on a false positive here:

image

for  comparison from demo:
image

my test upload

```tsv
First	Last	Middle	Phone	Email	CampusID	OtherID	Username	Password
anton	zylinder	b	555-555-5555	anton@zylinder.com	xx9999991	xx9999991	xx9999991
anton	zylinder	b	notaphone	notanemail	xx9999992	xx9999992	xx9999992	
```

jrjohnson:
> ```generic-db
> First	Last	Middle	Phone	Email	CampusID	OtherID	Username	Password
> anton	zylinder	b	555-555-5555	anton@zylinder.com	xx9999991	xx9999991	xx9999991
> ```

I *think* this bad data and was incorrectly not caught in the previous validation when I refactored it for octane 4 years ago. The password for this user is blank which isn't allowed if the username is present. However currently we do `const stringValue = String(this.password).trim();` in [the validation callback](https://github.com/ilios/frontend/blob/master/packages/frontend/app/components/bulk-new-users.js#L453) which doesn't work. When `this.password` is `null` `stringValue becomes "null"!! 🫣

I suppose the good news is that `"null"` is never gonna cause an issue if we store it as a password because we do a hash of the user input and no value will ever end up hashed to `"null"`, still though 🤦 

I will add a test for sure.

packages/frontend/app/components/bulk-new-users.js
jrjohnson: I'm always on the fence about this. Technically we're returning a promise into the `filter` and that filter will await the promise for us and act on the resolved value. So adding the await is unnecessary, but it does make it clearer what is happening.

packages/frontend/app/components/bulk-new-users.js
michaelchadwick: https://github.com/ilios/frontend/pull/8254 should help with avoiding such an error from even being generated.

stopfstedt (approved):
re-reviewed and click-tested. i couldn't find any regressions in the re-review. LGTM.

packages/frontend/app/components/bulk-new-users.js
```diff
@@ -169,7 +168,7 @@ export default class BulkNewUsersComponent extends Component {
     this.primaryCohortId = id;
   }
 
-  async existingUsernames() {
+  async getExistingUserNames() {
```
michaelchadwick, line 171: Do we not want to be consistent with naming convention here? I see `UserName`, but a few lines down on 184 you're using `Username`.

jrjohnson: Damn it, I hate it when very good and valid points are made! Good catch. I will fix it.... tomorrow.

jrjohnson: Fixed in #8258

PR #8245, opened 2024-11-27
Remove UserProfile render modifiers

michaelchadwick:
Removed pesky `ember-render-modifiers` instances in the `UserProfileBio` and `UserProfileRoles` components.

Initial plan was to create separate `UserProfileBio[Details|Manager]` sub-components, but it's more involved and somewhat out of scope with this PR, so will be a followup.


packages/frontend/app/components/user-profile-bio.js
```diff
@@ -44,6 +45,40 @@ export default class UserProfileBioComponent extends Component {
...
+    this.otherId = user.otherId;
+    this.email = user.email;
+    this.displayName = user.displayName;
+    this.pronouns = user.pronouns;
+    this.preferredEmail = user.preferredEmail;
+    this.phone = user.phone;
+  }
+
+  getAuthUser = modifier(() => {
+    if (this.userAuthentication) {
```
stopfstedt, line 66: i don't understand how/why this would work, but it does.  seems like modifiers react to component  property changes, like getters.

any chance this could be made more explicit? 

perhaps `userAuthentication` could be made input to the modifier. 
something like 

`getAuthUser = modifier((userAuthentication))` and  `{{this.getAuthUser this.userAuthentication}}`.

stopfstedt (approved):
LGTM.
kudos for bringing the post-save visual indicator back.

PR #8252, opened 2024-12-02
Remove CurriculumInventoryReport render modifiers

michaelchadwick:
Refs ilios/ilios#5374

This _was_ going to remove __all__ the render modifiers in the `CurriculumInventory` section, but acceptance tests are not catching errors introduced by their removal, so scaled back to only affect the `VerificationPreview` component and sub-components, which improves on the use of `scrollIntoView`.


stopfstedt (approved):
LGTM

PR #8253, opened 2024-12-04
Replace render-modifiers in curriculum inventory reports component with getters

stopfstedt:


michaelchadwick (approved):
This improvement is much more insightful than my amateur one, so kudos! Did some click testing and it seems to work as before.

jrjohnson (approved):

PR #8254, opened 2024-12-04
Add template protection for filetypes on uploads with restrictions

michaelchadwick:
Fixes ilios/ilios#5875

Added `accept` attribute to `<input>` elements for file uploads in `BulkNewUsersComponent` and `LearnerGroupUploadDataComponent` components to restrict choosing invalid filetypes.


stopfstedt (approved):
nice one.

PR #8255, opened 2024-12-06
Add title tooltip to objective buttons

michaelchadwick:
Fixes ilios/ilios#5868

Edit a few months later: made it so the `title=` element property only shows up if new `@showTitle` flag is set, so it only affects objectives, and not every single `EditableField` instance ever. Also, created a `PageObject` for `EditableField` and updated its test to use it.

### DIFF
--- a/packages/ilios-common/addon-test-support/ilios-common/page-objects/components/editable-field.js
+++ b/packages/ilios-common/addon-test-support/ilios-common/page-objects/components/editable-field.js
@@ -1,0 +1,32 @@
+import {
+  clickable,
+  count,
+  create,
+  isVisible,
+  property,
+  text,
+  triggerable,
+} from 'ember-cli-page-object';
+import { hasFocus } from 'ilios-common';
+import fadeText from './fade-text.js';
+
+const definition = {
+  scope: '[data-test-editable-field]',
+  value: text(),
+  hasEditIcon: isVisible('[data-test-edit-icon]'),
+  editable: {
+    scope: '[data-test-edit]',
+    edit: clickable(),
+    iconCount: count(),
+    title: property('title'),
+  },
+  enter: triggerable('keyup', 'input', { eventProperties: { key: 'Enter' } }),
+  escape: triggerable('keyup', 'input', { eventProperties: { key: 'Escape' } }),
+  cancel: clickable('[data-test-cancel]'),
+  inputHasFocus: hasFocus('input'),
+  textareaHasFocus: hasFocus('textarea'),
+  fadeText,
+};
+
+export default definition;
+export const component = create(definition);
--- a/packages/ilios-common/addon-test-support/ilios-common/page-objects/components/editable-field.js
+++ b/packages/ilios-common/addon-test-support/ilios-common/page-objects/components/editable-field.js
@@ -1,9 +1,9 @@
 import {
+  attribute,
   clickable,
   count,
   create,
   isVisible,
-  property,
   text,
   triggerable,
 } from 'ember-cli-page-object';
@@ -18,7 +18,7 @@ const definition = {
     scope: '[data-test-edit]',
     edit: clickable(),
     iconCount: count(),
-    title: property('title'),
+    title: attribute('title'),
   },
   enter: triggerable('keyup', 'input', { eventProperties: { key: 'Enter' } }),
   escape: triggerable('keyup', 'input', { eventProperties: { key: 'Escape' } }),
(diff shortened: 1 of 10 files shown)


packages/ilios-common/addon/components/course/objective-list-item-parents.hbs
```diff
@@ -33,7 +33,14 @@
       >
         {{#if @editable}}
           <span data-test-parent>
-            <button type="button" class="link-button" {{on "click" @manage}} data-test-manage>
+            <button
+              type="button"
+              class="link-button"
+              aria-label={{t "general.edit"}}
```
jrjohnson, line 39: I *think* having a label is redundant if there is a title.

packages/test-app/tests/integration/components/editable-field-test.js
```diff
@@ -1,42 +1,42 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'test-app/tests/helpers';
-import { render, click, triggerKeyEvent, waitFor } from '@ember/test-helpers';
+import { render, waitFor } from '@ember/test-helpers';
 import { hbs } from 'ember-cli-htmlbars';
+import { setupIntl } from 'ember-intl/test-support';
+import { component } from 'ilios-common/page-objects/components/editable-field';
```
jrjohnson, line 6: Love this!

packages/test-app/tests/integration/components/editable-field-test.js
```diff
@@ -152,9 +150,22 @@ module('Integration | Component | editable field', function (hooks) {
...
-    await click('[data-test-edit]');
+    await component.editable.edit();
     assert.ok(this.status);
-    await click('[data-test-cancel]');
+    await component.cancel();
     assert.notOk(this.status);
   });
+
+  test('it has title property value only if @showTitle is true', async function (assert) {
+    this.intl = this.owner.lookup('service:intl');
```
jrjohnson, line 160: Nice, but not standard for our tests. If you use `setupIntl(hooks, 'en-us');` it will ensure the language is set to English and you can just test that value. Or if you want to get really fancy, use the [t test helper](https://ember-intl.github.io/ember-intl/docs/test-helpers/t)

packages/ilios-common/addon-test-support/ilios-common/page-objects/components/editable-field.js
```diff
@@ -0,0 +1,32 @@
...
+
+const definition = {
+  scope: '[data-test-editable-field]',
+  value: text(),
+  hasEditIcon: isVisible('[data-test-edit-icon]'),
+  editable: {
+    scope: '[data-test-edit]',
+    edit: clickable(),
+    iconCount: count(),
+    title: property('title'),
```
jrjohnson, line 21: I don't know that it makes a huge different, but this should probably pull from [attribute](http://ember-cli-page-object.js.org/docs/v2.3.x/api/attribute) instead of `property`.

packages/ilios-common/addon/components/session/objective-list-item-parents.hbs
```diff
@@ -33,7 +33,14 @@
       >
         {{#if @editable}}
           <span data-test-parent>
-            <button type="button" class="link-button" {{on "click" @manage}} data-test-manage>
+            <button
+              type="button"
+              class="link-button"
+              aria-label={{t "general.edit"}}
```
jrjohnson, line 39: Repeat comment, but this probably isn't needed with the title appearing for sure.

jrjohnson (changes requested):

jrjohnson:
Forgot in my review: There is a case difference between `Edit` and `expand`. Feels weird to me.

packages/ilios-common/translations/en-us.yaml
```diff
@@ -124,7 +124,7 @@ general:
   etAl: et al.
   eventNotFoundTitle: Event Not Found
   eventNotFoundExplanation: "The event you have tried to reach is either no longer available, or you have followed a personalized user link which is not shareable. If you continue to encounter issues, please contact your Ilios help desk."
-  expand: expand
```
jrjohnson, line 127: It looks like `general.expand` is used in a lot of places. Can they all by re-cased?

michaelchadwick: I checked all the places it's used (less than a dozen), and it's all labels for buttons, just like in this case. No `You sure you want to {{t "general.expand"}} that thing, buddy?` instances, if that's what you're thinking.

jrjohnson (approved):

PR #8258, opened 2024-12-06
Fix Method Case

jrjohnson:
For consistency re-casing this method name to match what we have everywhere else in the application.

This is cleanup from code review in https://github.com/ilios/frontend/pull/8242#discussion_r1872382009


michaelchadwick (approved):
:-D

stopfstedt (approved):
iApproveThisChange

PR #8259, opened 2024-12-06
Use Better CI Reporter

jrjohnson:
This is a version of what we're using for our browser tests and the output is way more readable in CI, making the failures easier to spot and fix. I like it. Let's use it on our other CI tests.


michaelchadwick (approved):
C aye aye

PR #8260, opened 2024-12-06
replaces ember render modifiers in curr inv new report component.

stopfstedt:
streamlines the initialization and tracking of report year selections.


michaelchadwick (approved):
LGTM, although it would be nice to better understand why we were using a `findById` method before, but now do not @stopfstedt

PR #8263, opened 2024-12-09
replaces usage of ember-render-modifiers in curr inv sequence block header component.

stopfstedt:


michaelchadwick (approved):
Un-render those modifiers

michaelchadwick (approved):

PR #8264, opened 2024-12-12
Remove PublicationMenuComponent render modifiers

michaelchadwick:
Refs ilios/ilios#5374

### DIFF
--- a/packages/ilios-common/addon/components/course/publication-menu.js
+++ b/packages/ilios-common/addon/components/course/publication-menu.js
@@ -2,6 +2,7 @@ import Component from '@glimmer/component';
 import { service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
+import { task, timeout } from 'ember-concurrency';
 
 export default class CoursePublicationMenuComponent extends Component {
   @service router;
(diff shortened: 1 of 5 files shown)


packages/ilios-common/addon/components/course/publication-menu.js
```diff
@@ -63,53 +64,67 @@ export default class CoursePublicationMenuComponent extends Component {
     return 'notpublished';
   }
 
-  focusOnFirstItem(menuElement) {
-    menuElement.querySelector('button:first-of-type').focus();
+  focusFirstLink = task(async () => {
+    await timeout(1);
+    document.querySelector('.publication-menu .menu button:first-of-type').focus();
```
stopfstedt, line 69: this works since there's only ever one menu open at the same time. 
however, querying from the doc root is brittle.

i'm suggesting bringing the menu element back as input argument, then making the DOM query from that element instead.

would require identifying the menu element from the event target in the click handler -for example,  if the clicked element is nested within the menu ("menu item"), you'll have to traverse the DOM tree up from it.

packages/ilios-common/addon/components/session/publication-menu.js
```diff
@@ -67,53 +68,67 @@ export default class SessionPublicationMenuComponent extends Component {
     return 'notpublished';
   }
 
-  focusOnFirstItem(menuElement) {
-    menuElement.querySelector('button:first-of-type').focus();
+  focusFirstLink = task(async () => {
+    await timeout(1);
+    document.querySelector('.publication-menu .menu button:first-of-type').focus();
```
stopfstedt, line 73: please see my comment about the course-variant of this component.

stopfstedt (approved):
click tested and reviewed. LGTM.

PR #8265, opened 2024-12-12
Replace render modifiers in curr inv sequence block overview component

stopfstedt:

### DIFF
--- a/packages/frontend/app/components/curriculum-inventory/sequence-block-overview.hbs
+++ b/packages/frontend/app/components/curriculum-inventory/sequence-block-overview.hbs
@@ -361,7 +361,7 @@
         </div>
         <div class="block starting-academic-level" data-test-starting-academic-level>
           <label for="academic-level-{{templateId}}">{{t "general.startLevel"}}:</label>
-          {{#if (and @canUpdate this.startingAndEndingAcademicLevelsLoaded)}}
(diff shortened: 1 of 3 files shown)


packages/frontend/app/components/curriculum-inventory/sequence-block-overview.hbs
```diff
@@ -1,498 +1,512 @@
...
+            {{else}}
+              <span data-test-course-title>{{if
+                  this.course
+                  this.course.title
+                  (t "general.notApplicableAbbr")
+                }}</span>
+            {{/if}}
+          </span>
+          {{#if this.selectedCourse}}
+            {{#if this.selectedCourse.id}}
```
michaelchadwick, line 38: If there is a `selectedCourse`, but no `selectedCourse.id`, then nothing appears?

stopfstedt: that's correct. if the user un-selects a previously attached course, i need to track this somehow. i'm using an empty object for this `{}`. this has to be different than no user selection, which is `null`.  triple state - empty object, null, or course model.
not sure how to do this otherwise, i made several passes over this. please let me know if you have a better idea here.

michaelchadwick (approved):
Click-tested and spot-checked. Seems good to me.

PR #8267, opened 2024-12-12
Remove VisualizerProgramYearObjectives render modifiers

michaelchadwick:
Refs ilios/ilios#5374

Two odd things I found:
1. The link to these visualizations don't show up on demo: https://demo.iliosproject.org/programs/15/programyears/21 (no link), but the same route on local there's a visualization link
2. The link only works if you click from the previous ProgramYear route: /programs/15/programyears/21 -> click link -> /data/programyears/21/objectives, but not directly. Or if you go there once, and then refresh, you get Zoinks. I'm assuming there's some routing shenanigans that is only working on transition, and not directly.

### DIFF
--- a/packages/frontend/app/components/visualizer-program-year-objectives.js
+++ b/packages/frontend/app/components/visualizer-program-year-objectives.js
@@ -2,9 +2,10 @@ import Component from '@glimmer/component';
 import { htmlSafe } from '@ember/template';
 import { filter, map } from 'rsvp';
 import { restartableTask, timeout } from 'ember-concurrency';
-import { tracked } from '@glimmer/tracking';
+import { cached, tracked } from '@glimmer/tracking';
 import { service } from '@ember/service';
 import { filterBy, uniqueValues } from 'ilios-common/utils/array-helpers';
+import { TrackedAsyncData } from 'ember-async-data';
 
 export default class VisualizerProgramYearObjectivesComponent extends Component {
   @service intl;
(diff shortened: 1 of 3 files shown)


packages/frontend/app/components/visualizer-program-year-objectives.js
```diff
@@ -16,15 +17,29 @@ export default class VisualizerProgramYearObjectivesComponent extends Component
...
+  }
+
+  load = restartableTask(async () => {
+    const cohort = await this.programYear.cohort;
+    const year = await this.programYear.getClassOfYear();
     const classOfYear = this.intl.t('general.classOf', { year });
     this.programYearName = cohort.title ?? classOfYear;
 
-    this.data = await this.getData(programYear);
+    this.data = await this.getData(this.programYear);
```
jrjohnson, line 31: This doesn't time out correctly if the program year isn't already loaded. It *could* return null.  This fix is a bit more complicated, but worth the time:
1. Remove the `load` entirely
2. Create a `TrackedAsyncData` that calls `getData` using the program year. Until program year exists it should return null.
3. Probably lots of adjustment and cleanup to get that to work

jrjohnson (changes requested):

packages/frontend/app/components/visualizer-program-year-objectives.js
michaelchadwick: @jrjohnson Any idea why the links to these visualizations only exist on local, but not dev?

stopfstedt: @michaelchadwick the visibility of these links is controlled via a feature flag. 

for reference, see:
https://github.com/ilios/frontend/blob/master/packages/frontend/app/components/program-year/overview.hbs#L7

PR #8276, opened 2024-12-18
get rid of render-modifiers in sequence-block-session-manager component.

stopfstedt:
this removes all async loading stuff from the component. instead, the parent component ensures to pre-load all necessary data upfront as well as guarding against incompletely loaded data from being passed down.


michaelchadwick (approved):
Opa!

PR #8279, opened 2024-12-19
replaces ember-render-modifiers with getters in program-year components.

stopfstedt:


jrjohnson (approved):
Spent a while trying to get that loading spinner to show up, doesn't seem like it ever did anything. Begone!

PR #8280, opened 2024-12-19
Remove CourseRollover render modifiers

michaelchadwick:
Refs ilios/ilios#5374

I could not find a way to get a reference to the root DOM element of a component, which is needed for some loading, so I made a new custom modifier, `get-element`, which _feels_ like a side-step from `{did-insert}` instead of a true replacement, but is also [what Ember recommends creating](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-at-ember-render-modifiers.md), so...all good? :D

### DIFF
--- a/packages/ilios-common/addon/components/course/rollover-date-picker.hbs
+++ b/packages/ilios-common/addon/components/course/rollover-date-picker.hbs
@@ -1,8 +1,6 @@
 <input
   aria-label={{t "general.pickADate"}}
   data-test-course-rollover-picker
-  {{did-insert (perform this.showPicker) @course @value}}
-  {{did-update (perform this.showPicker) @course @value}}
-  {{on "click" (if this.isOpen this.close (noop))}}
+  {{get-element (perform this.showPicker @course @value)}}
   ...attributes
 />
--- a/packages/ilios-common/addon/components/course/rollover.hbs
+++ b/packages/ilios-common/addon/components/course/rollover.hbs
@@ -1,12 +1,6 @@
-<div
-  class="course-rollover"
-  {{did-insert (perform this.load) @course}}
-  {{did-update (perform this.load) @course}}
->
+<div class="course-rollover">
   {{#if this.save.isRunning}}
-    <WaitSaving
-      @showProgress={{false}}
-    />
+    <WaitSaving @showProgress={{false}} />
   {{/if}}
   {{#let (unique-id) as |templateId|}}
     <div class="backtolink">
@@ -34,7 +28,7 @@
           disabled={{this.save.isRunning}}
           placeholder={{t "general.courseTitlePlaceholder"}}
           data-test-title
-        >
+        />
         <ValidationError @validatable={{this}} @property="title" />
       </div>
       <div class="item year-select">
@@ -46,6 +40,7 @@
             id="year-{{templateId}}"
             data-test-year
             {{on "change" this.setSelectedYear}}
+            {{get-element this.loadValidYear}}
           >
             {{#each this.years as |year|}}
               <option
@@ -54,7 +49,9 @@
                 selected={{is-equal year this.selectedYear}}
               >
                 {{#if this.academicYearCrossesCalendarYearBoundaries}}
-                  {{year}} - {{add year 1}}
+                  {{year}}
+                  -
+                  {{add year 1}}
                 {{else}}
                   {{year}}
                 {{/if}}
@@ -73,16 +70,18 @@
         <div class="item">
           <label>
             {{t "general.startDate"}}:
-            <br>
+            <br />
             <small>
               ({{t "general.mustBeSameDayOfWeekAsCurrentStartDate"}})
             </small>
           </label>
-          <Course::RolloverDatePicker
-            @course={{@course}}
-            @value={{this.startDate}}
-            @onChange={{set this "startDate"}}
-          />
+          {{#if this.allCourses}}
+            <Course::RolloverDatePicker
+              @course={{@course}}
+              @value={{this.startDate}}
+              @onChange={{set this "startDate"}}
+            />
+          {{/if}}
         </div>
         <div class="included">
           <label class="title" for="included-{{templateId}}">
@@ -95,7 +94,7 @@
               checked={{not this.skipOfferings}}
               data-test-skip-offerings
               {{on "click" (toggle "skipOfferings" this)}}
-            >
+            />
             <span>
               {{t "general.offerings"}}
             </span>
@@ -117,10 +116,7 @@
         <button
           type="button"
           disabled={{if
-            (or
-              (not this.selectedYear)
-              (includes this.selectedYear this.unavailableYears)
-            )
+            (or (not this.selectedYear) (includes this.selectedYear this.unavailableYears))
             true
           }}
           class="done text"
--- a/packages/ilios-common/addon/components/course/rollover.hbs
+++ b/packages/ilios-common/addon/components/course/rollover.hbs
@@ -75,7 +75,7 @@
               ({{t "general.mustBeSameDayOfWeekAsCurrentStartDate"}})
             </small>
           </label>
-          {{#if this.allCourses}}
+          {{#if (is-array this.allCourses)}}
             <Course::RolloverDatePicker
               @course={{@course}}
               @value={{this.startDate}}
--- a/packages/ilios-common/addon/components/course/rollover.hbs
+++ b/packages/ilios-common/addon/components/course/rollover.hbs
@@ -35,7 +35,7 @@
         <label for="year-{{templateId}}">
           {{t "general.year"}}:
         </label>
-        {{#if (is-array this.allCourses)}}
+        {{#if this.allCoursesData.isResolved}}
           <select
             id="year-{{templateId}}"
             data-test-year
@@ -75,7 +75,7 @@
               ({{t "general.mustBeSameDayOfWeekAsCurrentStartDate"}})
             </small>
           </label>
-          {{#if (is-array this.allCourses)}}
+          {{#if this.allCoursesData.isResolved}}
             <Course::RolloverDatePicker
               @course={{@course}}
               @value={{this.startDate}}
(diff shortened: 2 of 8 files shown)


packages/test-app/tests/integration/modifiers/get-element-test.js
```diff
@@ -0,0 +1,14 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'test-app/tests/helpers';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Modifier | get-element', function (hooks) {
+  setupRenderingTest(hooks);
+
+  // Replace this with your real tests.
```
jrjohnson, line 9: Should go away

packages/ilios-common/addon/components/course/rollover-date-picker.hbs
```diff
@@ -1,8 +1,6 @@
 <input
   aria-label={{t "general.pickADate"}}
   data-test-course-rollover-picker
-  {{did-insert (perform this.showPicker) @course @value}}
-  {{did-update (perform this.showPicker) @course @value}}
-  {{on "click" (if this.isOpen this.close (noop))}}
+  {{get-element (perform this.showPicker @course @value)}}
```
jrjohnson, line 4: I've been leaving these with a `did-insert` and figured we'd clean them up. I have no problem with this approach though, let's do it! We have a few places where we *do* need the element for something and this is a good solution.

packages/ilios-common/addon/components/course/rollover.hbs
```diff
@@ -46,6 +40,7 @@
             id="year-{{templateId}}"
             data-test-year
             {{on "change" this.setSelectedYear}}
+            {{get-element this.loadValidYear}}
```
jrjohnson, line 43: This usage doesn't track for me though. There isn't anything here that needs access to the DOM. This data should be loaded in a TrackedAsyncData and processed in a getter.

packages/ilios-common/addon/components/course/rollover.hbs
```diff
@@ -73,16 +70,18 @@
...
             <small>
               ({{t "general.mustBeSameDayOfWeekAsCurrentStartDate"}})
             </small>
           </label>
-          <Course::RolloverDatePicker
-            @course={{@course}}
-            @value={{this.startDate}}
-            @onChange={{set this "startDate"}}
-          />
+          {{#if this.allCourses}}
```
jrjohnson, line 78: If `this.allCourses` what? It's working off of the null and that's valid, but it's hard to read it here. Would prefer `if this.allCoursesData.isResolved`.

packages/ilios-common/addon/components/course/rollover.js
```diff
@@ -31,27 +30,44 @@ export default class CourseRolloverComponent extends Component {
...
+  }
+
+  get academicYearCrossesCalendarYearBoundaries() {
+    return this.academicYearCrossesCalendarYearBoundariesData.isResolved
+      ? this.academicYearCrossesCalendarYearBoundariesData.value
+      : false;
+  }
+
+  @cached
+  get allCoursesData() {
```
jrjohnson, line 50: This is OK. It works. But it will force a reload of all courses (`store.query` is never cached).  My approach here would be to create a `get schoolData` that loads `this.args.course.school` and then have the `allCoursesData` getter to `this.school.courses`.

packages/ilios-common/addon/components/course/rollover.js
```diff
@@ -31,27 +30,44 @@ export default class CourseRolloverComponent extends Component {
...
+  }
+
+  get allCourses() {
+    return this.allCoursesData.isResolved ? this.allCoursesData.value : null;
+  }
 
-    // set selectedYear to next valid year (or current if none found)
+  @action
+  async loadValidYear() {
+    await this.allCourses;
```
jrjohnson, line 67: This value can't be awaited. It will always be `null` and resolve instantly. I think you'll need to refactor this out into getters anyway, based on using the element modifier to load this, but worth nothing this troublespot I've hit before. The `TrackedAsyncData` pattern is nice, but it has this gotcha built in.

packages/ilios-common/addon/modifiers/get-element.js
```diff
@@ -0,0 +1,20 @@
...
+  <div {{get-element this.doStuffWithElement}}>
+    <!-- Content -->
+  </div>
+  ```
+*/
+export default modifier(function getElement(element, [callback]) {
+  if (typeof callback === 'function') {
+    callback(element); // Pass the element to the provided callback
+  } else {
+    console.warn('get-element modifier expects a callback as the first positional argument.');
```
jrjohnson, line 18: Let's be harsher here. Grab `import { assert } from '@ember/debug';` and use it to ensure the callback is a function. That will throw nicely in development mode.

packages/test-app/tests/integration/modifiers/get-element-test.js
```diff
@@ -0,0 +1,14 @@
...
+  test('it renders', async function (assert) {
+    await render(hbs`<div {{get-element this.setRootElement}}></div>`);
+    assert.ok(true);
```
jrjohnson, line 12: Needs a real test here to ensure that `this.setRootElement` gets called. Also needs to test the assertion failure when a non-callback is passed. I didn't check, but I bet this test is currently throwing a warning in the console when it runs.

jrjohnson (changes requested):
Great idea to sub in out own modifier where it is very much needed here. Some other cleanup needed as well though before this is ready.

packages/ilios-common/addon/components/course/rollover.hbs
michaelchadwick: Changing it to `{{#if this.allCoursesData.isResolved}}` just makes the DatePicker not show up at all. If I change it to `{{#if (is-array this.allCourses)}}`, like the usage for the other `<select>` that holds `this.years`, then it works. It seems they both check to make sure it's not null.

packages/ilios-common/addon/components/course/rollover.hbs
michaelchadwick: The reason for adding that there is that the `Course::RolloverDatePicker` is waiting for the **Year** `<select>` element to load before it gets the `selectedYear` + 1 value. If I don't have that there, then the **Year** `<select>` loads, say, `2025`, and the date picker remains on `11/26/2024` instead of rolling over to `11/25/2025` like it currently does on demo.

jrjohnson: How about wrapping everything in an `if` so it only tries to render once the data needed is resolved?

packages/ilios-common/addon/modifiers/get-element.js
michaelchadwick: I tried this approach, but couldn't get the new real test to pass for when the modifier is given an invalid/missing callback. I get a `global error` from `qunit` on the `render(hbs``)` call, and no mangling of the global window error event listeners seemed to quell it. There's no example in the codebase where we do this, so I'm currently out of ideas.

packages/ilios-common/addon/components/course/rollover.js
michaelchadwick: Good point. Removed.

packages/test-app/tests/integration/modifiers/get-element-test.js
michaelchadwick: Added a real set of tests, but see https://github.com/ilios/frontend/pull/8280#discussion_r1894261100 as to why I have not been able to test a "failing" assertion yet.

packages/ilios-common/addon/components/course/rollover.js
michaelchadwick: I'm not exactly sure how to refactor this per your suggestion. Could you supply some code to help guide me? 😃

jrjohnson: Absolutely. Sort of free hand though so might have syntax errors:

Add a school data property:
```javascript
@cached
get schoolData() {
  return new TrackedAsyncData(this.args.course.school);
}
```

Then consume it to load all the courses like

```javascript
@cached
get coursesData() {
  return new TrackedAsyncData(this.schoolData.isResolved ? this.schoolData.value.courses : null);
}
```

Then use that to get the list of courses for the current school:

```javascript
get allCourses() {
if (!this.coursesData.isResolved || this.coursesData.value === null) {
return [];
}

return this.courses.data.value;
}
```

This will use ember-data's built in cache because all the courses in the school are already fetched from the API.

packages/ilios-common/addon/modifiers/get-element.js
jrjohnson: I haven't done this one either, fun new stuff! Looks like maybe you can use `assert.Throws` from [qunit](https://qunitjs.com/api/assert/throws/).

Possibly like:

```javascript
assert.throws(await render(hbs`<div id='root-element' {{get-element}}></div>`));
```

If that doesn't work I'd be ok with skipping the second test.

packages/ilios-common/addon/components/course/rollover.hbs
michaelchadwick: Is not the whole thing already wrapped in an `if`?

```
{{#if this.allCoursesData.isResolved}}
  <select
    id="year-{{templateId}}"
    data-test-year
    {{on "change" this.setSelectedYear}}
    {{get-element this.loadValidYear}}
  >
    {{#each this.years as |year|}}
      ...
    {{/each}}
  </select>
{{else}}
  <LoadingSpinner />
{{/if}}
```

packages/ilios-common/addon/components/course/rollover.hbs
michaelchadwick: After changing the getters, this is now serviced by the `{{#if this.allCoursesData.isResolved}}`.

packages/ilios-common/addon/modifiers/get-element.js
michaelchadwick: I tried my hardest, but all I get in the end is a `global failure` in `qunit` due to the throwing of an Error. I put in a `todo` for now.

michaelchadwick:
This has been superseded so closing.

PR #8282, opened 2024-12-21
Replace ember render modifier in learner group instructor manager component

stopfstedt:


michaelchadwick (approved):
This looks like the exact model I need to use for https://github.com/ilios/frontend/pull/8245

PR #8288, opened 2024-12-30
Fix New Year Errors

jrjohnson:
Discovered several interconnected issues around the end of 2024 and the start of 2025. The most serious of these was that opening week at a glance all weeks for 2025 blew up as we were asking for 53 weeks in a 52 week year. This fixes ilios/ilios#5908.

I also discovered that the wrong title was appearing at the top of the dashboard week view during this time period as well.

PR #8291, opened 2024-12-30
correct variable names - these are not dates, but week-of-the-year numbers

stopfstedt:
follow-up to #8288


jrjohnson (approved):
Good change!

michaelchadwick (approved):

PR #8292, opened 2024-12-30
break style declarations for heading-title and button-title apart.

stopfstedt:
the SCSS compiler in prod evidently works differently than in non-prod, so we've been getting different CSS output here.
this makes things consistent again in both types of envs.

fixes https://github.com/ilios/ilios/issues/5912


stopfstedt:
this did not do what i was hoping for. the issue seems to be `font: inherit` not being correctly overridden when one of the sub-declarations, like `font-size` is provided.

PR #8294, opened 2025-01-06
Removes dependency on ember-cli-clean-css.

stopfstedt:
from this plugin's [README](https://github.com/ember-cli/ember-cli-clean-css?tab=readme-ov-file):

> Note
>
> This addon is not needed and is NOT used when [embroider](https://github.com/embroider-build/embroider) is enabled.

we're building with embroider in frontend, so let's remove this dependency.

refs https://github.com/ilios/ilios/issues/5912


jrjohnson (approved):
Excellent!

PR #8299, opened 2025-01-10
Remove DetailCohortsComponent render modifier for detail learning materials header

michaelchadwick:
Refs ilios/ilios#5374


stopfstedt (approved):
click tested and code reviewed. the scroll back to the LM management component header after adding a new LM still works as intended.

LGTM.

PR #8308, opened 2025-01-11
Remove MeshManagerComponent input render modifiers

michaelchadwick:
Refs ilios/ilios#5374

### DIFF
--- a/packages/ilios-common/addon/components/mesh-manager.hbs
+++ b/packages/ilios-common/addon/components/mesh-manager.hbs
@@ -1,15 +1,32 @@
 <section class="mesh-manager" data-test-mesh-manager>
-  {{#if @targetItemTitle}}
-    <h2 class="target-title">
-      {{t "general.selectMeshFor" title=@targetItemTitle}}
-    </h2>
-  {{/if}}
-  {{#if this.sortedTerms.length}}
-    <ul class="selected-terms">
-      {{#each this.sortedTerms as |term|}}
-        <li>
-          {{#if @editable}}
-            <button type="button" {{on "click" (fn @remove term)}}>
+  {{#let (unique-id) as |templateId|}}
+    {{#if @targetItemTitle}}
+      <h2 class="target-title">
+        {{t "general.selectMeshFor" title=@targetItemTitle}}
+      </h2>
+    {{/if}}
+    {{#if this.sortedTerms.length}}
+      <ul class="selected-terms">
+        {{#each this.sortedTerms as |term|}}
+          <li>
+            {{#if @editable}}
+              <button type="button" {{on "click" (fn @remove term)}}>
+                <span class="term-title">
+                  {{term.name}}
+                </span>
+                <span class="term-details">
+                  {{term.id}}
+                  {{#if term.deleted}}
+                    -
+                    <span class="deprecated">({{t "general.deprecatedAbbreviation"}})</span>
+                  {{else if term.trees.length}}
+                    -
+                    <MeshDescriptorLastTreeNumber @descriptor={{term}} />
+                  {{/if}}
+                </span>
+                <FaIcon @icon="xmark" class="remove" />
+              </button>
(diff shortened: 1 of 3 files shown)


packages/ilios-common/addon/components/mesh-manager.hbs
```diff
@@ -1,139 +1,131 @@
...
+      <input
+        id={{this.meshSearchInputId}}
+        autocomplete="off"
+        class="search-input"
+        type="search"
+        value={{this.query}}
+        placeholder={{t "general.search"}}
+        {{on "input" this.update}}
+        {{on "keyup" this.keyUp}}
+      />
```
stopfstedt, line 59: XHTML :fist_oncoming:

stopfstedt (changes requested):
this `moveFocus()` action doesn't seem to be wired up.

```bash
stefan@nichtsnutz: ~/dev/projects/frontend on remove-mesh-manager-input-render-modifiers[$]
$ grep -r 'moveFocus' packages --exclude-dir=dist --exclude-dir=node_modules
packages/ilios-common/addon/components/search-box.js:  moveFocus() {
packages/ilios-common/addon/components/search-box.js:    this.moveFocus();
packages/ilios-common/addon/components/mesh-manager.js:  moveFocus() {
packages/frontend/app/components/locale-chooser.js:  moveFocus(event) {
packages/frontend/app/components/locale-chooser.hbs:          {{on "keyup" this.moveFocus}}
```

proposing complete removal of this action and the horse that it rode in, including the id/element stuff.

michaelchadwick:
> this `moveFocus()` action doesn't seem to be wired up.
> 

@stopfstedt Excellent catch. I think I just assumed there was a magnifying glass icon in the Mesh Manager search input just like there is one in the Learning Materials one (and the general search box at the top). It's like that method was _never_ wired up, actually.

However, since we already use the magnifying glass in other search inputs, should we not just add one here so it has one? To add an icon in one place or to remove the method across the board seems like a global question outside the scope of this PR.

michaelchadwick:
@stopfstedt Well, after trying to implement the magnifying glass icon in the search input, I realized why it probably isn't there: the input just autocompletes as you type. There's no reason to ever click an icon to "submit", so adding it is worthless at best, and potentially confusing at worst. I'm now not sure why we have it on the other search boxes, except for maybe accessibility reasons?

packages/ilios-common/addon/components/mesh-manager.js
```diff
@@ -16,7 +17,14 @@ export default class MeshManagerComponent extends Component {
...
   @tracked hasMoreSearchResults = false;
-  @tracked searchInput;
+
+  get meshSearchInputId() {
+    return `mesh-search-input-${guidFor(this)}`;
+  }
+
+  get meshSearchInputElement() {
+    return document.getElementById(this.meshSearchInputId);
+  }
```
stopfstedt, line 27: this can be removed again.

packages/ilios-common/addon/components/mesh-manager.hbs
```diff
@@ -1,139 +1,131 @@
...
+              {{/if}}
+            </span>
+          {{/if}}
+        </li>
+      {{/each}}
+    </ul>
+  {{/if}}
+  {{#if @editable}}
+    <div class="mesh-search" data-test-mesh-search>
+      <label for={{this.meshSearchInputId}}>{{t "general.searchMesh"}}:</label>
```
stopfstedt, line 49: this change has become unnecessary again, you can revert to the previous form of using the `unique-id as |templateId|` approach.

stopfstedt (changes requested):
sorry for giving you the runaround on this, but the since the need for getting a hold of this element within the component class is gone, some of this can be scrapped and the template can be reverted to the previous state (but leaving the nice cleanup intact).

stopfstedt (approved):
LGTM

PR #8320, opened 2025-01-17
Ember cli string helpers replacement

stopfstedt:
fixes https://github.com/ilios/ilios/issues/5927

### DIFF
--- a/packages/test-app/tests/integration/helpers/capitalize-test.js
+++ b/packages/test-app/tests/integration/helpers/capitalize-test.js
@@ -12,7 +12,7 @@ module('Integration | Helper | {{capitalize}}', function (hooks) {
 
     let expected = 'Hi';
 
-    assert.dom('*').hasText(expected, 'capitalizes a single string');
+    assert.dom().hasText(expected, 'capitalizes a single string');
   });
 
   test('It leaves the capitalization unchanged with correctly capitalized string', async function (assert) {
@@ -20,7 +20,7 @@ module('Integration | Helper | {{capitalize}}', function (hooks) {
 
     let expected = 'Harry';
 
-    assert.dom('*').hasText(expected, 'leaves capitalization when string is already capitalized');
+    assert.dom().hasText(expected, 'leaves capitalization when string is already capitalized');
   });
 
   test('It correctly capitalizes an uncapitalized sentence', async function (assert) {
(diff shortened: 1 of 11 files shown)


packages/ilios-common/addon/-private/create-string-helper.js
michaelchadwick, line 1: An empty private file?

stopfstedt: baleted.

packages/test-app/tests/integration/helpers/capitalize-test.js
```diff
@@ -0,0 +1,59 @@
...
+
+module('Integration | Helper | {{capitalize}}', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('It capitalizes a single string', async function (assert) {
+    await render(hbs`{{capitalize 'hi'}}`);
+
+    let expected = 'Hi';
+
+    assert.dom().hasText(expected, 'capitalizes a single string');
```
michaelchadwick, line 15: I like seeing test messages like these!

packages/test-app/tests/integration/helpers/html-safe-test.js
```diff
@@ -0,0 +1,35 @@
...
+module('Integration | Helper | {{html-safe}}', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('It html-safes the html string', async function (assert) {
+    await render(hbs`{{html-safe '<h1>Hello World</h1>'}}`);
+
+    assert.dom('h1').hasText('Hello World', 'html string is correctly rendered');
+  });
+
+  test('It safely renders CSS classes from a property', async function (assert) {
```
michaelchadwick, line 16: What's an example where this would be used? I've never used any `htmlSafe` method on a CSS class before.

jrjohnson: Something like https://github.com/ilios/frontend/blob/eabdb58011518b7173d460ae80cbcbb3c0d3661c/packages/ilios-common/addon/components/progress-bar.js#L5-L9 where if we don't mark the string as `safe` Ember will escape it to prevent [css injection](https://owasp.org/www-project-web-security-testing-guide/v41/4-Web_Application_Security_Testing/11-Client_Side_Testing/05-Testing_for_CSS_Injection). If you put the class into the template that's fine, but if you attempt to do work on these properties in a getter or some other JS property Ember will call shenanigans.

jrjohnson (approved):

PR #8321, opened 2025-01-18
Remove @ember/edition-utils dependency

michaelchadwick:
This dependency is included, but seems to not be used, and hasn't had an update in years. Let's remove it.


jrjohnson (approved):
🗑️

PR #8328, opened 2025-01-23
corrects prefix for brand icons to get the Black Tie icon back on screen.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/5944


michaelchadwick (approved):
Not sure how this was incorrect for so long, and only just noticed, but maybe UCSF doesn't hold many fancy shindigs?

PR #8330, opened 2025-01-24
Removed 'MeSH Term' as object option when 'Term' is selected as subject

michaelchadwick:
Fixes ilios/ilios#4546


stopfstedt (approved):
LGTM

PR #8333, opened 2025-01-30
Correct Rollover Year Options

jrjohnson:
Courses can always be rolled over into the previous, current, and several future years.

The academic year doesn't map to the calendar year, so we always need to ensure we're starting at the right place otherwise the current year will disappear in January.

From January through the end of June we need to back up even further because the previous academic year is now two years behind the current calendar year.

Fixes ilios/ilios#5963


dartajax (approved):
looks good to me - I approve

PR #8334, opened 2025-01-31
Attempt to make all top/left/right margin/padding consistent (eek)

michaelchadwick:
Fixes ilios/ilios#5112

Through the changing of many `.scss` files, I've attempted to normalize the top/left/right margin/padding of Ilios's main layout and components.

I'm sure I missed something somewhere, but I tried to do as an exhaustive job as I could. I know this is a huge set of changes, and there was plenty of "what I think it should look like" at play, so expecting pushback, but let's get this one going!

### DIFF
--- a/packages/ilios-common/app/styles/ilios-common/components/print-course.scss
+++ b/packages/ilios-common/app/styles/ilios-common/components/print-course.scss
@@ -4,7 +4,7 @@
 .print-course {
   .header {
     background-color: c.$culturedGrey;
-    padding: 0 1rem;
+    padding: 0 0.5rem;
     width: 100%;
 
     h2 {
@@ -27,7 +27,7 @@
   .block {
     @include m.detail-container(c.$orange);
     background-color: c.$white;
-    padding: 0 1rem;
+    padding: 0.5rem;
 
     .title {
       @include m.detail-container-title;
@@ -43,6 +43,14 @@
 
     .static-list {
       @include m.ilios-static-list;
+    }
+  }
+
+  .print-course-session {
+    .session-objective-list {
+      .fade-text-control {
+        background-image: linear-gradient(to bottom, transparent, c.$white);
+      }
     }
   }
 
--- a/packages/frontend/app/components/curriculum-inventory/new-sequence-block.hbs
+++ b/packages/frontend/app/components/curriculum-inventory/new-sequence-block.hbs
@@ -3,7 +3,7 @@
   data-test-curriculum-inventory-new-sequence-block
   ...attributes
 >
-  <h2 class="title" data-test-title>
+  <h2 class="new-sequence-block-title" data-test-title>
     {{t "general.newSequenceBlock"}}
   </h2>
   {{#let (unique-id) as |templateId|}}
--- a/packages/frontend/app/components/new-user.hbs
+++ b/packages/frontend/app/components/new-user.hbs
@@ -1,8 +1,8 @@
 {{#let (unique-id) as |templateId|}}
   <div class="new-user" data-test-new-user ...attributes>
-    <h3>
+    <h4 class="title">
       {{t "general.newUser"}}
-    </h3>
+    </h4>
     <div class="new-user-form">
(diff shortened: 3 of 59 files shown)


packages/ilios-common/app/styles/ilios-common/components/print-course.scss
```diff
@@ -46,6 +46,14 @@
     }
   }
 
+  .print-course-session {
```
jrjohnson, line 49: Did this come from somewhere else?

michaelchadwick:
@dartajax Can you please give this a visual spot-check? Many things were touched, and I need a second pair of eyes.

jrjohnson (approved):

packages/ilios-common/app/styles/ilios-common/components/print-course.scss
michaelchadwick: I'm not exactly what sure what you mean, but Course Summary reports have many `.print-course-session`-classed `<div>` sections, and they needed some space above their titles imo.

dartajax:
image

well they almost line up in the shot above - I noticed this in the PR preview and I can see why it occurred and I'm not sure if I care although I kind of like it better the way it is in the first (demo current ver) screen shot ...

image

My $.02 is that it's the padding applied inside the other screen - again not sure if I care but just something I noticed - other than that my Space Committee (tm) brain likes the slight space saving that we get with the more standardized padding.

dartajax:
image

The lower "Actions" is fine so applying that formatting to the above one so they match is my suggestion here.

dartajax:
The same is true on the left -  make the padding on the learning materials (left and right) match the padding of the offerings level (left and right) - it'd be smoother and more consistent.

dartajax:
image

dartajax:
one follow up - well one more I should say - while navigating the page, it is sometimes necessary to click on the outside of any given control to skip over that list and go to the next one - with such a thin amount of padding this is a bit more difficult - users might get confused if they are used to how it currently is - maybe we could put a slight border on it so they know where the control ends and the "neutral zone" begins

dartajax:
I'd recommend @saschaben to check this out too when time is available or I can demo these comments (and maybe more) at our next meeting -

michaelchadwick:
@dartajax here're some responses:
* Fixed the Session->LMs table padding so every element in the Session section has a uniform left/right padding
* Re: Published buttons differing on this PR - I don't think it's possible to fix as the session info has the extra padding so it doesn't bump up against sides of its container, and that just doesn't match the outer Course header stuff anymore with all the changes
* I'm not clear on what you meant by "click on the outside of any given control to skip over that list and go to the next one". Will you please screenshot an example?

packages/ilios-common/app/styles/ilios-common/components/print-course.scss
jrjohnson: Of course, you are completely allowed to add new styles. For some reason I thought everything *must* be an adjustment. 🤦‍♂️

dartajax:
> @dartajax here're some responses:
> 
>     * Fixed the Session->LMs table padding so every element in the Session section has a uniform left/right padding
> 
>     * Re: Published buttons differing on this PR - I don't think it's possible to fix as the session info has the extra padding so it doesn't bump up against sides of its container, and that just doesn't match the outer Course header stuff anymore with all the changes
> 
>     * I'm not clear on what you meant by "click on the outside of any given control to skip over that list and go to the next one". Will you please screenshot an example?

I don't think a screen shot would be helpful but a demonstration might. I find Ilios a bit confusing to navigate when you can't tell if your focus is inside a control or outside - especially since some of our lists can be long and you might want to go right past them to another section. It's hard to find the "border" or when you are outside of a control if the padding is super small.

michaelchadwick:
@dartajax I think doing a quick demo of the site tomorrow will work wonders.

michaelchadwick:
@dartajax @jrjohnson I added the background-color and header consistency of the New [X] widgets all over the site, too. Hopefully that still aligns with the general feel of this PR.

PR #8348, opened 2025-02-03
Sunburst our Loading

jrjohnson:
There are a few routes where we don't have a good loading state, plus some inner route transitions where it's hard to tell if clicking on the link did anything. Let's add a white screen with a slowly visible sunburst so the user knows Ilios is doing something.

Ideally we'll do loading routes for everywhere this shows up, but we have a gap right now we need to fill.

### DIFF
--- a/packages/frontend/app/components/loading-sunburst.hbs
+++ b/packages/frontend/app/components/loading-sunburst.hbs
@@ -1,0 +1,1 @@
+<div class="loading-sunburst"></div>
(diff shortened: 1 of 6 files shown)


michaelchadwick:
Is this a formal componentization of the `<div id="ilios-loading-indicator">` sunburst?

jrjohnson:
After team discussion, this isn't better. Will work more on individual loaders for each route.

PR #8350, opened 2025-02-05
Don't search on empty string Global Search Box

michaelchadwick:
Fixes ilios/ilios#5972

Chrome and Safari seem to takeover the `esc` key and just clear the search box (which is what I'd expect), but Firefox actually runs our custom code and does a search on an empty string. Leaving the hook in there, but removing the search, makes the experience the same across the board.

### DIFF
--- a/packages/frontend/app/components/global-search-box.js
+++ b/packages/frontend/app/components/global-search-box.js
@@ -96,7 +96,6 @@ export default class GlobalSearchBox extends Component {
 
     if (this.isEscapeKey(keyCode)) {
       this.clear();
-      this.args.search('');
     }
   }
 


dartajax (changes requested):
I noticed when searching for "psyll" = no results that when you use the `esc` key on Chrome, it clears out the search criteria box for global search. On Firefox, it does not do this. I like this behavior and plus we are trying to standardize, eh? Can we make this happen on Firefox as well? Let me test some more browsers - if confusion ensues, I will remove this change request since it works better than before that's for sure.

dartajax:
I tested on Safari and Opera - `esc` works with a search string of "psyll" (0 records returned) so hopefully we can make the experience the same on Firefox.

jrjohnson:
I *think* the intention of sending the empty string is to bubble up to the controller and clear the URL of the search string. I think that may be currently broken as well. Makes me wonder if this code path is getting called at all when escape is pressed.

michaelchadwick:
There's definitely a bit more here than I thought. It seems like, currently, Chrome/Safari don't run that code path at all, but Firefox does. However, running the current code path in Firefox triggers a search when not on the `search` route, but does nothing on it (whereas Chrome/Safari continue to override it by just clearing the search input's value.

michaelchadwick:
I did a bit more research on this, and we can force Chrome/Safari to respect our custom keyboard handler for the `esc` key, but it doesn't seem to make a difference.

@dartajax I tried both immediately pressing `esc` and waiting for results to come up before pressing it, and on all browsers this new code fixes the "press `esc` and get a page full of no results issue".

@jrjohnson The codepath to clear the search query is, in fact, broken.

michaelchadwick:
The logic to clear both the search input and the querystring should now be working.

jrjohnson (approved):
LGTM, didn't test behavior, just checked code.

dartajax (approved):
works for me - tested

PR #8351, opened 2025-02-05
Disable Route Splitting

jrjohnson:
This is a broken features for us currently as it doesn't include loading routes when it splits breaking our loading experience in a confusing way.


jrjohnson:
Code climate can be ignored here. This is ready for review.

michaelchadwick:
Nitpick: if you're going to comment out the whole feature, why not leave in the actual, but commented-out, routes so they can be more easily commented in by someone else who doesn't know them as well? And yes I know one could just look at the git history, but just sayin'.

michaelchadwick (approved):
See one minor nitpick, but otherwise good.

jrjohnson:
> why not leave in the actual, but commented-out, routes

Good call.
The current list of routes isn't doing anything good, but there is some possibility we'll get this enabled again. Once this is merged I'll follow up with a restoration PR that we can leave in draft that has a better list of routes in it and some tests for the loading routes in particular.

PR #8352, opened 2025-02-06
Remove Reporting render modifiers

michaelchadwick:
Refs ilios/ilios#5374

This is the first of seven `ember-render-modifiers` to remove in reporting, but wanted to see if this is a good approach. Could potentially abstract the `modifier` into its own file and pull it in, but dunno.


michaelchadwick:
Superseded by #8363

PR #8353, opened 2025-02-07
Fix Token Generation

jrjohnson:
At some specific time points (such as between 11pm and midnight we were generating negative numbers which were invalid ISO 8601 periods and would be rejected by the API. Instead of rolling our own we can now use [Luxon's built in toIso() ](https://moment.github.io/luxon/api-docs/index.html#duration)method to gather this data.

Fixes ilios/ilios#5976


stopfstedt (approved):
LGTM

PR #8364, opened 2025-02-11
Remove Reporting Competency render modifier

michaelchadwick:
Refs ilios/ilios#5374

After much consternation and gnashing of teeth, this PR has become the potential prototype for updates that may come to #8363, #8366, #8367, and #8368 (and another one for `Terms` eventually).

Major change in that when the `prepositionalObject` of `competency` is chosen, the array of `prepositionalObjectId`s no longer have a default selection as the options to implement it are heady and most likely not worth the effort. This does mean that someone has to manually choose something from the `<select>` in order to run the report. If they don't, an error message will appear (which already happens for `Instructor` and `MeSH Term` object types), and the `<select>` will be focused.

### DIFF
--- a/packages/frontend/app/components/reports/new-subject.js
+++ b/packages/frontend/app/components/reports/new-subject.js
@@ -340,7 +340,9 @@ export default class ReportsNewSubjectComponent extends Component {
     const isValid = await this.isValid();
     if (!isValid) {
       if (this.prepositionalObject === 'competency' && !this.prepositionalObjectId) {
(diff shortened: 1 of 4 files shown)


stopfstedt:
during click-testing, i realized that the the reports won't run if the user has not made a selection. we may have to force the user to make a selection here, by introducing a default/non-value "select competency..." option to the dropdown.

stopfstedt (approved):

stopfstedt:
@dartajax please click-test as well. thanks!

dartajax (approved):
seems fine to me

PR #8371, opened 2025-02-19
Make UserProfileCalendar event blocks routable

michaelchadwick:
Fixes ilios/ilios#4295

Clicking on an event in the calendar does nothing, but clicking on an event when on the `Dashboard::Calendar` component routes the user to an event detail, so I changed it so they work the same way.

### DIFF
--- a/packages/frontend/app/components/user-profile-calendar.hbs
+++ b/packages/frontend/app/components/user-profile-calendar.hbs
@@ -32,9 +32,10 @@
     <IliosCalendarWeek
       @calendarEvents={{this.calendarEvents}}
       @date={{this.date}}
-      @areEventsSelectable={{false}}
+      @areEventsSelectable={{true}}
       @areDaysSelectable={{false}}
       @isLoadingEvents={{this.eventsData.isPending}}
+      @selectEvent={{@selectEvent}}
     />
   </div>
 </div>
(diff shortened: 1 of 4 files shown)


michaelchadwick:
@dartajax Needs some discussion over the issue I found that events not connected to the current logged in user (e.g. demo_dev1 looking at non-demo_dev1 events) go to a 404 Event Not Found page instead of the expected event detail page.

dartajax:
> @dartajax Needs some discussion over the issue I found that events not connected to the current logged in user (e.g. demo_dev1 looking at non-demo_dev1 events) go to a 404 Event Not Found page instead of the expected event detail page.

I think my gut instinct on this one was that a user like `demo_dev1` and basically any user who has access to the admin console would have enough permissions to be routed to the event.

jrjohnson:
This may prove tricky because it's not an event for the user. The event detail routes expect to know and load the current user unless the event it a school event. Wondering if we change the way we load events on this calendar to load all school events and the filter for the user we're interested in. Or something else complicated.

dartajax:
not sure how I feel about this one -- I am logged in as `Nicholas Gutierrez` who has permissions to see anything as that user is a user with many permissions indeed 

as Nicholas, I look up the student learner "demo_student28" in the Admin Console and click to view this user's calendar - the events can now be clicked upon but always I "demo_dev1" end up with something like the following after clicking ...

See image comments below ...

image

Why would power user `demo_dev1` not have access to this event? Better just close the whole thing probably and leave the events not clickable I guess or ...?

dartajax (approved):
seems to be working as intended - great job

michaelchadwick:
NTS: make a followup to better handle this without the `isUserProfileCalendar` flag.

PR #8372, opened 2025-02-20
Change Reporting->[Subject]->Academic Year to work like new Competency

michaelchadwick:
Refs #8363, #8364

The change in #8363, while successfully removing the `ember-render-modifier` usage, introduced a new bug: a report won't run unless a selection is manually made for the specific Academic Year, as the default selection is no longer made when the options are loaded.

The change made in #8364 does a similar removal of an `ember-render-modifier` for the Competency option, but also changes the way the dropdown default selection works, in that a manual selection _needs_ to be made, and it's more obvious now.

This PR does the same thing for Academic Year.

### DIFF
--- a/packages/frontend/app/components/reports/new-subject.js
+++ b/packages/frontend/app/components/reports/new-subject.js
@@ -339,7 +339,9 @@ export default class ReportsNewSubjectComponent extends Component {
     this.addErrorDisplaysFor(['title', 'prepositionalObject', 'prepositionalObjectId']);
     const isValid = await this.isValid();
     if (!isValid) {
-      if (this.prepositionalObject === 'competency' && !this.prepositionalObjectId) {
+      const dropdownObjectTypes = ['academic year', 'competency'];
+
+      if (dropdownObjectTypes.includes(this.prepositionalObject) && !this.prepositionalObjectId) {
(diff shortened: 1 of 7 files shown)


stopfstedt (approved):

dartajax (dismissed):
any chance of having the sort order be in descending order and the possibility of having the current year "2024" which = "2024-25" be selected by default? if not, I will merge either way and dismiss my change request - at least make the sort order descending instead of starting at "2009" if possilbe

dartajax:
> any chance of having the sort order be in descending order and the possibility of having the current year "2024" which = "2024-25" be selected by default? if not, I will merge either way and dismiss my change request - at least make the sort order descending instead of starting at "2009" if possible

I am amending my request - I realized the default selection is not necessary but the sort order should still be flipped to Z-A by year. Cool?

michaelchadwick:
@dartajax I can't make it choose anything by default anymore, because of the generalized object Reporting change introduced in #8364. There will be 4 other objects, Instructor Group, Program Year, Session Type, and Term, that will have the same thing done to them, in order to remove `ember-render-modifiers` from their templates.

However, I did reverse the year values, as you requested.

dartajax (approved):

PR #8377, opened 2025-02-22
Fix Courses by Mesh Term Report

jrjohnson:
The SQL query we need to run on the API has way too many joins and performs terribly for the courses by meshDescriptors filter. Instead we can reverse the lookup here and first pull the descriptor and all of it's relationships and then turn that into course ids and query them like we do any other filter.

Makes for messy tests because the graphQL responses have to use a counter, but it's 100x faster and won't crash the API DB anymore so that seems worth it.

Refs ilios/ilios#5650
Fixes ilios/ilios#4540

### DIFF
--- a/packages/frontend/app/components/reports/subject/course.js
+++ b/packages/frontend/app/components/reports/subject/course.js
@@ -74,15 +74,38 @@ export default class ReportsSubjectCourseComponent extends Component {
       rhett.push(`schools: [${school.id}]`);
     }
     if (prepositionalObject && prepositionalObjectTableRowId) {
-      let what = pluralize(camelize(prepositionalObject));
       if (prepositionalObject === 'mesh term') {
-        what = 'meshDescriptors';
-        prepositionalObjectTableRowId = `"${prepositionalObjectTableRowId}"`;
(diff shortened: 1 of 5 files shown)


michaelchadwick:
This doesn't seem to change anything?
Screenshot 2025-02-24 at 10 01 48 AM

jrjohnson:
Added this to the MeshTerms by course report as well, so both ways are now fixed.

michaelchadwick:
Screenshot 2025-02-25 at 8 13 15 AM
Still seems like a problem :-/

stopfstedt (approved):
LGTM

jrjohnson:
Pulling this back into Draft mode as there appear to be two issue:
1. Unable to create a new Mesh Terms for Course report, getting a 400 from the API. Weird.
2. Running this report for a large course (like Ground School 2022) isn't as performant as it should be. May need to make some more adjustments for this case.

jrjohnson:
Back out of draft mode. The issue with running the report found by @michaelchadwick are when a school filter was added, I've fixed that. I'm unable to reproduce the saving issues I'm seeing in the netlify against demo, even running against demo on my local. I think it may be something unrelated.

michaelchadwick:
@jrjohnson Woo! That report now runs quickly.

PR #8381, opened 2025-02-26
Limit Report results for better browser performance

michaelchadwick:
Fixes ilios/ilios#6002

Some reports, regardless of how long they take, can return many (hundreds, sometimes thousands) results, which can tax browser rendering.

I've added max results functionality (currently ~~100~~200), and any report done with Courses that returns more than that number of results will engage it. This PR still has the other subject types to implement, some usability bugs (like if you use the Years Filter), and tests to fix, but I wanted to get this out there for initial poking and prodding.

- [x]  Competencies
- [x]  Courses
- [x]  Instructors
- [x]  Instructor Groups
- [x]  Learning Materials
- [x]  MeSH Terms
- [x]  Program Years
- [x]  Programs
- [x]  Session Types
- [x]  Sessions
- [x]  Terms


jrjohnson:
Early thoughts, didn't spend a lot of time in the code.

Works great and way better! 

Instead of setting the results count as a tracked property I think it should be a getter. And I know it's the UI that was requested, but I don't like using the course expanded here. I don't think we have a good pattern yet. We use a `Show More` button on Mesh Search. I'm ok with that here. Maybe with a down arrow icon. I just think the orange bar is too much.

I love the results count. I don't know if it should show `100/667` when we hide things because of the show more. I do think it makes sense when items are hidden due to the year filter, but then maybe it doesn't go in that position. 

I'm not a fan of the scrolling. It's think `scrollIntoView` is too slow here. Would rather just see `element.scrollTo` directly.

michaelchadwick:
NTS: Change max to 200; remove expand/collapse button and replace with "download to see all results" functionality.

jrjohnson (changes requested):
I'm not sure how I feel about passing around the subject header instance here because I think it would work just as well to extract the download button into it's own component and allow it to be used in two places.

packages/frontend/app/components/reports/subject-download.js
```diff
@@ -0,0 +1,42 @@
...
+import { TrackedAsyncData } from 'ember-async-data';
+import createDownloadFile from 'ilios-common/utils/create-download-file';
+import PapaParse from 'papaparse';
+
+export default class ReportsSubjectDownload extends Component {
+  @service reporting;
+  @tracked finishedBuildingReport = false;
+
+  @cached
+  get reportTitleData() {
```
jrjohnson, line 14: Looks like this is getting passed in. Hopefully it can go away here and remove all the async from this component?

packages/frontend/app/controllers/reports/subject.js
```diff
@@ -6,6 +6,7 @@ export default class ReportsSubjectController extends Controller {
   @service store;
 
   queryParams = ['report', 'reportYear'];
+  resultsLengthMax = 200;
```
jrjohnson, line 9: Since each report results display has to manage this it makes more sense to me to have it be a value within that component and not something that gets passed down. Maybe make it a constant for easier search / readability, but I don't see us wanting to make this configurable or anything special so let's drop the mental overhead of passing it around.

packages/frontend/app/components/reports/subject/course.hbs
```diff
@@ -1,20 +1,23 @@
...
-  @prepositionalObject={{@prepositionalObject}}
-  @prepositionalObjectTableRowId={{@prepositionalObjectTableRowId}}
-  @year={{@year}}
-  @changeYear={{@changeYear}}
-  @showYearFilter={{not-eq @prepositionalObject "academic year"}}
-  @description={{@description}}
-  @fetchDownloadData={{this.fetchDownloadData}}
-  @readyToDownload={{this.data.isResolved}}
-/>
+{{#if this.allCoursesData.isResolved}}
```
jrjohnson, line 1: I'm not seeing the reason to hide this here, especially with the `@readyToDownload` argument that prevents any shenanigans. I, for sure, don't think we need both.

packages/frontend/tests/integration/components/reports/subject-download-test.js
```diff
@@ -0,0 +1,119 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'frontend/tests/helpers';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import { setupMirage } from 'frontend/tests/test-support/mirage';
+import { setupAuthentication } from 'ilios-common';
+import { component } from 'frontend/tests/pages/components/reports/subject-download';
+import a11yAudit from 'ember-a11y-testing/test-support/audit';
+
+module('Integration | Component | reports/subject-download', function (hooks) {
```
jrjohnson, line 10: Solid test!

jrjohnson (changes requested):

packages/frontend/app/components/reports/subject-download.js
michaelchadwick: Looking over my code, I can see what you mean. There's a lot of stuff that doesn't need to be passed in at this point, so reconfiguring a bunch now...

packages/frontend/app/components/reports/subject/course.js
```diff
@@ -12,13 +12,16 @@ export default class ReportsSubjectCourseComponent extends Component {
   @service iliosConfig;
   @service currentUser;
   @service intl;
+  @service reporting;
```
jrjohnson, line 15: Left behind in refactor?

packages/frontend/translations/en-us.yaml
```diff
@@ -352,6 +352,7 @@ general:
   reportDisplayTitleWithoutObject: "All {subject} in {school}"
   reportName: Report Name
   reportNamePlaceholder: Please enter a report name.
+  reportResultsExceedMax: Only the first 200 results shown. Download report to see all results.
```
jrjohnson, line 355: The 200 is hardcoded here, but it's also hardcoded in the course results. For maximum refactorability this should get passed in as `{resultsLimit}` when this translation is invoked in the `course.hbs` template.

packages/frontend/app/components/reports/subject-results.hbs
```diff
@@ -8,5 +8,6 @@
     @changeYear={{@changeYear}}
     @year={{@year}}
     @description={{@description}}
+    @resultsLength={{@resultsLength}}
```
jrjohnson, line 11: I think this is leftover refactor and not longer used.

jrjohnson (changes requested):
Everything is mostly looking good and ready for other reports. Couple of small changes needed.

packages/frontend/app/components/reports/subject/mesh-term.js
```diff
@@ -22,12 +24,20 @@ export default class ReportsSubjectMeshTermComponent extends Component {
...
   }
 
-  get sortedData() {
-    return this.data.value.sort((a, b) => {
+  get allMeshTerms() {
+    return this.allMeshTermsData.isResolved ? this.allMeshTermsData.value : [];
+  }
+
+  get sortedMeshTerms() {
+    return this.allMeshTermsData.value.sort((a, b) => {
```
jrjohnson, line 32:
```suggestion
    return this.allMeshTerms.sort((a, b) => {
```

packages/frontend/app/components/reports/subject/competency.js
```diff
@@ -45,8 +55,12 @@ export default class ReportsSubjectCompetencyComponent extends Component {
...
   }
 
+  get reportResultsExceedMax() {
+    return this.allCompetencies.length > this.resultsLengthMax;
+  }
+
   @action
   async fetchDownloadData() {
-    return [[this.intl.t('general.competencies')], ...this.sortedData.map((v) => [v])];
+    return [[this.intl.t('general.competencies')], ...this.allCompetencies.map((v) => [v])];
```
jrjohnson, line 64:
```suggestion
    return [[this.intl.t('general.competencies')], ...this.sortedCompetencies.map((v) => [v])];
```

packages/frontend/app/components/reports/subject/term.js
```diff
@@ -23,8 +25,16 @@ export default class ReportsSubjectTermComponent extends Component {
...
   }
 
-  get sortedData() {
-    return sortBy(this.data.value, ['vocabulary.title', 'title']);
+  get allTerms() {
+    return this.allTermsData.isResolved ? this.allTermsData.value : [];
+  }
+
+  get sortedTerms() {
+    return sortBy(this.allTermsData.value, ['vocabulary.title', 'title']);
```
jrjohnson, line 33:
```suggestion
    return sortBy(this.allTerms, ['vocabulary.title', 'title']);
```

jrjohnson (changes requested):
Couple of tiny changes, but it also needs the tests in session-results `it renders all results when resultsLengthMax is not reached` and `it renders limited results and an extra download button when resultsLengthMax is eclipsed` for all the reports as that behavior is in each component. It's boilerplate, but worth it I think.

PR #8385, opened 2025-03-03
prevent Percy from being built during installation.

stopfstedt:
see https://pnpm.io/package_json#pnpmignoredbuiltdependencies


michaelchadwick (approved):
Woot!

PR #8395, opened 2025-03-04
replaces rollover-specific date picker with our generic one.

stopfstedt:
cramming it all in there

partially supersedes https://github.com/ilios/frontend/pull/8280

while in there, i reworked the tracking of some of the modifier attributes and their corresponding input changes, to keep setting/re-setting attributes on the picker object to an absolute minimum. (comparing object values, like Dates, in JS is still  _ze würst_)


packages/ilios-common/addon/components/course/rollover.js
```diff
@@ -71,6 +79,12 @@ export default class CourseRolloverComponent extends Component {
     this.selectedCohorts = this.selectedCohorts.filter((obj) => obj !== cohort);
   }
 
+  @action
+  updateStartDate(newStartDate) {
+    // if a date is forced that isn't allowed
```
jrjohnson, line 84: I'm not sure what this means?

packages/ilios-common/addon/modifiers/date-picker.js
```diff
@@ -24,29 +28,49 @@ export default class DatePickerModifier extends Modifier {
...
-    if (this.flatpickr.selectedDates[0] !== value) {
-      this.flatpickr.setDate(value);
-    }
-    if (this.flatpickr.minDate !== minDate) {
-      this.flatpickr.set('minDate', minDate);
-    }
-    if (this.flatpickr.maxDate !== maxDate) {
-      this.flatpickr.set('maxDate', maxDate);
+    if (Array.isArray(allowedWeekdays)) {
+      this.allowedWeekdays = allowedWeekdays;
```
jrjohnson, line 38: Edge case, but if we switch from a list to null passed in value this won't reset back to an empty array.

jrjohnson (changes requested):

packages/ilios-common/addon/modifiers/date-picker.js
stopfstedt: fair point. i'll put an `else`  clause in, and default to empty array for anything else but array input.

packages/ilios-common/addon/components/course/rollover.js
stopfstedt: i copied that verbatim from the old rollover date picker component.

the only way i'm able to trigger an `undefined` value to be passed here is by interacting with the input field that the picker is tethered to. i.e. attempting to use backspace to clear out the field, or try copy/paste a new value in.  it doesn't change the input in the UI, but will fire the onChange event on the picker for some reason.

packages/ilios-common/addon/modifiers/date-picker.js
stopfstedt: handled :heavy_check_mark:

jrjohnson (approved):

PR #8399, opened 2025-03-04
Remove OfferingCalendar render modifiers

michaelchadwick:
Refs ilios/ilios#5374

### DIFF
--- a/packages/ilios-common/addon/components/offering-calendar.js
+++ b/packages/ilios-common/addon/components/offering-calendar.js
@@ -1,7 +1,6 @@
 import Component from '@glimmer/component';
 import { cached, tracked } from '@glimmer/tracking';
 import { map } from 'rsvp';
-import { deprecate } from '@ember/debug';
 import { DateTime } from 'luxon';
 import { TrackedAsyncData } from 'ember-async-data';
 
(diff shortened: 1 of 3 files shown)


packages/ilios-common/addon/components/offering-calendar.js
```diff
@@ -12,59 +12,19 @@ export default class OfferingCalendar extends Component {
...
-      const eventIdentifier = event.name + event.startDate + event.endDate;
-      return eventIdentifier !== currentEventIdentifier;
-    });
-
-    return [...filteredEvents, this.currentEvent];
+  @cached
+  get calendarEventsData() {
+    return new TrackedAsyncData(
+      this.loadData(
+        this.args.startDate,
```
jrjohnson, line 19: This could be `this.startDate` so it passes through the getter, or just delete the `startDate` getter because that deprecation is long handled and not needed anymore.  Same for `endDate`. As far as I can tell they're otherwise unused.

jrjohnson (changes requested):

packages/ilios-common/addon/components/offering-calendar.js
michaelchadwick: Excellent catch. I decided to remove the `get`ters since that was using `@ember/debug` and it seemed more prudent to rid the codebase of such things.

jrjohnson (approved):

PR #8400, opened 2025-03-05
Update Ember to v6.2.2 

jrjohnson:
Using ember-cli-update which includes a blueprint update to the latest
eslint config flat file format. I ported over a minimal set of
configuration options from our old config, but kept it as stock as
possible.

Still had to update some stuff for the new config.

WIP:
- [x] There is an issue with `ember-exam` and this updated `test-helpers.js` blueprint. It appears that every test is loaded into every browser. Investigating. Resolved in 803ca07.

Fixes ilios/ilios#5921


stopfstedt (approved):
LGTM

PR #8401, opened 2025-03-05
Remove OfferingForm render modifiers

michaelchadwick:
Refs ilios/ilios#5374

### DIFF
--- a/packages/ilios-common/addon/components/offering-form.js
+++ b/packages/ilios-common/addon/components/offering-form.js
@@ -56,7 +56,6 @@ export default class OfferingForm extends Component {
   @tracked savedOfferings = 0;
   @tracked recurringDayOptions = null;
   @tracked availableInstructorGroups = [];
-  @tracked loaded = false;
   @tracked saveProgressPercent;
   @tracked urlChanged = false;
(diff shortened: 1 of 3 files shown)


packages/ilios-common/addon/components/offering-form.js
```diff
@@ -149,19 +150,77 @@ export default class OfferingForm extends Component {
...
+      associatedSchools = uniqueValues(cohortSchools);
+    }
+    const allInstructorGroups = await map(associatedSchools, (school) => school.instructorGroups);
+    return allInstructorGroups.reduce((flattened, arr) => {
+      return [...flattened, ...arr];
+    }, []);
+  }
+
+  loadDefaultAttrs() {
+    if (this.loaded) {
```
stopfstedt, line 187: I think this `loaded` attribute can be removed entirely - it looks like we're tracking state for no good reason with this.

stopfstedt (commented):
please see my comment re: `this.loaded` and check if this attribute can be removed.

otherwise, this looks good to me.

stopfstedt (approved):
LGTM

PR #8404, opened 2025-03-06
add share button for Curriculum Report URL

michaelchadwick:
Fixes ilios/ilios#6030

### DIFF
--- a/packages/frontend/app/components/reports/curriculum/header.hbs
+++ b/packages/frontend/app/components/reports/curriculum/header.hbs
@@ -34,11 +34,11 @@
         aria-label={{t "general.copyCurriculumReportUrl"}}
         id={{this.copyButtonId}}
         data-test-copy-url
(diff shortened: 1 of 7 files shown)


jrjohnson (changes requested):
I think it should only show up if the report has been wrong. I'm not against it also showing up after some courses are picked, but it shouldn't show up when the report isn't runnable.

jrjohnson (approved):
Code looks good. Going to ask @dartajax for feedback on user experience as it was his request originally.

dartajax (dismissed):
image

<>

image

I prefer the formatting of the first mouse over callout but matching formats would be nice.

michaelchadwick:
@dartajax Made the tooltip style the same.

@jrjohnson Might want to just glance at the added code.

dartajax (approved):
looks good to me

jrjohnson (approved):

PR #8405, opened 2025-03-07
Remove Classic Classes from Tests

jrjohnson:
We had this linting exception in place for tests where we were still using this old pattern. Removed the exception and cleaned up the invocations. I also removed a comment that keeps bugging me and I never remember to remove. Little bonus!


michaelchadwick (approved):
Approval: granted

PR #8408, opened 2025-03-08
Remove HtmlEditor render modifier

michaelchadwick:
Refs ilios/ilios#5374

I'm very aware that what I'm doing with the template is...unorthodox...but I can't figure out how else to do it, so please enlighten me! It _does_ work, though, even if I don't know why.

### DIFF
--- a/packages/ilios-common/addon/components/html-editor.js
+++ b/packages/ilios-common/addon/components/html-editor.js
@@ -1,9 +1,10 @@
 import Component from '@glimmer/component';
 import { service } from '@ember/service';
-import { cached, tracked } from '@glimmer/tracking';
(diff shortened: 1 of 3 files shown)


packages/ilios-common/addon/components/html-editor.js
```diff
@@ -28,12 +28,25 @@ export default class HtmlEditorComponent extends Component {
     this.editorId = guidFor(this);
   }
 
+  @cached
+  get createEditor() {
```
jrjohnson, line 32: I'm having second thoughts about this name. Should probably be `loadFroala` `createdEditor` is a little difficult to understand when reading, especially the template.

jrjohnson (approved):

PR #8418, opened 2025-03-10
Update @ember/test-waiters

jrjohnson:
Unpin our locked versions and update this in common where we actually use it.

~I expect tests to fail here. Waiting on `ember-test-waiters #512`~

Unlocked! By getting everything up to 4.0 we seem to be ready to go here.


stopfstedt:
@jrjohnson needs rebasing. otherwise, this is good to go.

PR #8427, opened 2025-03-11
added label for AcademicYear report filter

michaelchadwick:
Fixes ilios/ilios#6004

### DIFF
--- a/packages/frontend/app/components/reports/subject-year-filter.hbs
+++ b/packages/frontend/app/components/reports/subject-year-filter.hbs
@@ -1,11 +1,6 @@
 <div class="years-filter" data-test-report-subject-year-filter>
   <label>{{t "general.academicYear"}}
-    <select
-      aria-label={{t "general.academicYears"}}
-      id="subject-year-filter"
-      data-test-year-filter
-      {{on "change" (pick "target.value" @changeYear)}}
-    >
+    <select data-test-year-filter {{on "change" (pick "target.value" @changeYear)}}>
       <option selected={{is-empty @selectedYear}} value="">
         {{t "general.allAcademicYears"}}
       </option>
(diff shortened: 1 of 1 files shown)


packages/frontend/app/components/reports/subject-year-filter.hbs
```diff
@@ -1,16 +1,19 @@
...
-  >
-    <option selected={{is-empty @selectedYear}} value="">
-      {{t "general.allAcademicYears"}}
-    </option>
-    {{#each (sort-by "title:desc" this.allAcademicYears) as |year|}}
-      <option value={{year.id}} selected={{eq year.id @selectedYear}}>
-        {{year.title}}
+  <label>{{t "general.academicYear"}}
+    <select
+      aria-label={{t "general.academicYears"}}
```
jrjohnson, line 4: This can go away now the two have a real label.

packages/frontend/app/components/reports/subject-year-filter.hbs
```diff
@@ -1,16 +1,19 @@
...
+      id="subject-year-filter"
```
jrjohnson, line 5: Not sure what this line is doing?

jrjohnson (changes requested):

packages/frontend/app/components/reports/subject-year-filter.hbs
michaelchadwick: Ah, yes, I needed an id when I first added the label because I was using `<label for="">` but apparently just putting the form element inside a `<label>` is the proper way to do things now (so says linting), so I removed this.

packages/frontend/app/components/reports/subject-year-filter.hbs
michaelchadwick: I didn't realize you could remove an `aria-label` if there was an associated `<label>`, but TIL.

dartajax (approved):
works for me

PR #8429, opened 2025-03-11
Replace render modifiers in course rollover component

stopfstedt:

i'm still a bit iffy by setting tracked props in the `save()` task, but don't see another way to shoe-horn these in if the user hasn't changed the defaults _before_ form validation runs.


jrjohnson (approved):
😤 love it.

PR #8430, opened 2025-03-11
for vocabulary selection, filter out vocabularies where all top-level terms are inactive.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6033

### DIFF
--- a/packages/ilios-common/addon/components/taxonomy-manager.js
+++ b/packages/ilios-common/addon/components/taxonomy-manager.js
@@ -55,8 +55,23 @@ export default class TaxonomyManager extends Component {
     }
   }
 
+  @cached
+  get assignableVocabulariesData() {
+    return new TrackedAsyncData(this.getAssignableVocabularies(this.nonEmptyVocabularies));
+  }
+
   get assignableVocabularies() {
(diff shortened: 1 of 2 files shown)


jrjohnson (approved):

dartajax:
I take it these terms are set up so if a higher level term is inactive, any sub terms are also assumed to be inactive.

dartajax:
> I take it these terms are set up so if a higher level term is inactive, any sub terms are also assumed to be inactive.

Check "OEQ" for reference - the higher level terms are all inactive but the sub-terms are still active. I am fine with that - just want to make sure.

stopfstedt:
> > I take it these terms are set up so if a higher level term is inactive, any sub terms are also assumed to be inactive.
> 
> Check "OEQ" for reference - the higher level terms are all inactive but the sub-terms are still active. I am fine with that - just want to make sure

yes, that's what Sascha asked for. from the corresponding ticket.

> When all terms within a vocabulary are disabled at the top level (ie, if there is nothing a user can pick from the term list when tagging), that vocabulary should NOT be available in the drop down select list

dartajax (approved):

PR #8432, opened 2025-03-12
Remove Get

jrjohnson:
Using this utility from ember is now linted against. The functionality it used to provide is either unnecessary or replaced with optional chaining.

### DIFF
--- a/packages/ilios-common/eslint.config.mjs
+++ b/packages/ilios-common/eslint.config.mjs
@@ -51,7 +51,6 @@ export default [
       reportUnusedDisableDirectives: 'error',
     },
     rules: {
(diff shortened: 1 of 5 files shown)


packages/ilios-common/eslint.config.mjs
```diff
@@ -51,7 +51,6 @@ export default [
       reportUnusedDisableDirectives: 'error',
     },
     rules: {
-      'ember/no-get': 'off',
```
michaelchadwick, line 54: Ember? No. Get off!

michaelchadwick (approved):
```
/--==--\
| 0  0 |
|  /\  |
| \__/ |
\======/

PR #8436, opened 2025-03-13
show inactive label on terms regardless of edit or view mode.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6048

this changes the output in a couple of places, if applicable:

in course/session objectives management:

image

in program year objectives management: 

image

in program year details:

image

in course details:

image

in session details:

image

in course print view:

image

### DIFF
--- a/packages/frontend/tests/acceptance/course/objectivelist-test.js
+++ b/packages/frontend/tests/acceptance/course/objectivelist-test.js
@@ -26,7 +26,7 @@ module('Acceptance | Course - Objective List', function (hooks) {
       competency: competencies[0],
     });
     const programYearObjectiveWithoutCompetency = this.server.create('program-year-objective');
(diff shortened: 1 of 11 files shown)


jrjohnson:
Code is good, but I'm going to put a needs team discussion on this, this closes the ticket, but I'm not sure we want to make it so visibly difficult to look at courses that have historically inactive terms. Those terms were active when applied and are probably correct. The large red inactive is pretty menacing and hard to look at.

stopfstedt:
> Code is good, but I'm going to put a needs team discussion on this, this closes the ticket, but I'm not sure we want to make it so visibly difficult to look at courses that have historically inactive terms. Those terms were active when applied and are probably correct. The large red inactive is pretty menacing and hard to look at.

per team discussion - we landed on red text, but not bolded.

michaelchadwick (approved):

PR #8437, opened 2025-03-14
prevent contents in removable tag list item from running together by removing flex display.

stopfstedt:
during review, please check percy output.

fixes https://github.com/ilios/ilios/issues/6047


dartajax:
Oh I get it ... maybe. The blank space I was requesting to have it be like `Complicated Beats (_inactive_)` rather than `Complicated Beats(_inactive_)` has to be applied whether "(_inactive_)" is there or not - is that what it is? I think my original ticket was hoping to only put the blank space in when the term that had been selected was inactive not every time. Is that the case?

PR #8441, opened 2025-03-17
Use a Modifier for Validation

jrjohnson:
We had several different patterns here to both start showing the error message and to ensure that the message gets updated when the value changes. I've replaced them all with a modifier on the validation class that doesn't show any errors until the user focuses out of the input, but does re-run the validations anytime the value changes.

Refs ilios/ilios#6055

### DIFF
--- a/packages/test-app/tests/integration/components/course/rollover-test.js
+++ b/packages/test-app/tests/integration/components/course/rollover-test.js
@@ -468,6 +468,7 @@ module('Integration | Component | course/rollover', function (hooks) {
     const input = `${title} input`;
 
     await fillIn(input, '');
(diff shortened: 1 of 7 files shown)


packages/test-app/tests/integration/components/course/rollover-test.js
```diff
@@ -468,6 +468,7 @@ module('Integration | Component | course/rollover', function (hooks) {
     const input = `${title} input`;
 
     await fillIn(input, '');
+    await emberBlur(input);
```
stopfstedt, line 471: today i learned about the `blur` test helper.

stopfstedt (approved):
LGTM

PR #8448, opened 2025-03-18
Remove SessionCopy render modifiers

michaelchadwick:
Refs ilios/ilios#5374

### DIFF
--- a/packages/ilios-common/addon/components/session-copy.js
+++ b/packages/ilios-common/addon/components/session-copy.js
@@ -1,11 +1,12 @@
 import Component from '@glimmer/component';
 import { service } from '@ember/service';
 import { hash, all, filter } from 'rsvp';
-import { dropTask, restartableTask, timeout } from 'ember-concurrency';
+import { dropTask, timeout } from 'ember-concurrency';
 import { action } from '@ember/object';
-import { tracked } from '@glimmer/tracking';
+import { cached, tracked } from '@glimmer/tracking';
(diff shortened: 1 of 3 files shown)


packages/ilios-common/addon/components/session-copy.js
```diff
@@ -17,7 +18,12 @@ export default class SessionCopyComponent extends Component {
   @tracked years;
   @tracked allCourses;
```
stopfstedt, line 19: you can replace this tracked property with a getter altogether.

something like:

```
get allCourses() {
   return this.allCoursesData.isResolved ? this.allCoursesData.value : [];
}
```

then change `loadCourses()` to return on its last line instead of setting the property.

stopfstedt (changes requested):

packages/ilios-common/addon/components/session-copy.js
michaelchadwick: I thought that the async method setting the tracker was not as good as it could be, but for some reason `return`-ing an `await` method did not seem right. I was wrong! Thanks.

stopfstedt (approved):
LGTM

stopfstedt:
@dartajax please click-test session copying. seems to be working as intended to me, but a second pair of eyes might be good in any case.

dartajax (approved):
copy works fine - I found another unrelated bug so there's that - the unrelated bug occurs in the PR and on demo

PR #8454, opened 2025-03-18
Replace Validator in New Learning Material

jrjohnson:
Replacing the class validator with yup.

Using the .test validator we can write custom tests and return custom
messages for any field.

Refs ilios/ilios#6055

### DIFF
--- a/packages/ilios-common/addon-test-support/ilios-common/page-objects/components/new-learningmaterial.js
+++ b/packages/ilios-common/addon-test-support/ilios-common/page-objects/components/new-learningmaterial.js
@@ -23,7 +23,9 @@ const definition = {
   agreement: clickable('[data-test-copyright-agreement]'),
   save: clickable('.done'),
   cancel: clickable('.cancel'),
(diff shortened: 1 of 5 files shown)


stopfstedt (approved):
first into the breach! thanks for tackling a whole bunch of these tricky ones. will use these as templates going forward.
LGTM

stopfstedt:
@dartajax please give all three variations (link, citation, file) of adding a new learning material a round of click-testing for possible regressions. i couldn't find any myself. for context: we're changing form validation libraries here.

dartajax:
I managed to raise this error ...

image

I was entering a ticket https://github.com/ilios/ilios/issues/6074 testing between my local proxied to demo and testing this PR - I am trying to determine the exact steps to reproduce this issue - staleness is part of it - I navigated away after raising an error message and then came back and fixed the issue and attempted to save the learning material attachment to the course - course level LM

jrjohnson:
That's almost certainly a bug here @dartajax, if you can figure out how to trigger it I'd love to fix it. I'm guessing we're not validating something and then sending bad data.

PR #8455, opened 2025-03-18
Silence Remaining Build Scripts

jrjohnson:
Not sure why only built doesn't silence the others, but we seem to need to list them all. So... I've done that.


stopfstedt (approved):
yes!

PR #8461, opened 2025-03-20
Validate Titles With Yup

jrjohnson:
Knocked out all the components where we *only* do a title validation.

Refs ilios/ilios#6055


packages/frontend/app/components/reports/subject-header.js
```diff
@@ -3,13 +3,17 @@ import { cached, tracked } from '@glimmer/tracking';
...
+import { string } from 'yup';
 
-@validatable
 export default class ReportsSubjectHeader extends Component {
   @service reporting;
-  @tracked @Length(1, 240) title = '';
+  @tracked title = '';
+
+  validations = new YupValidations(this, {
+    title: string().max(240),
```
michaelchadwick, line 15: Should this not have a `.min(1)` as part of the chain?

packages/frontend/app/components/school-manager.js
```diff
@@ -2,20 +2,24 @@ import Component from '@glimmer/component';
...
-  @tracked @NotBlank() @Length(1, 60) title;
+  @tracked title;
 
   constructor() {
     super(...arguments);
     this.title = this.args.school.title;
   }
 
+  validations = new YupValidations(this, {
+    title: string().required().max(60),
```
michaelchadwick, line 20: Same comment as before: should this have a `.min(1)`?

packages/frontend/app/components/reports/subject-header.js
jrjohnson: Good catch, but I think it's ok this way. In our old validator we'd ignore empty in this case, but our new validator catches an empty as less than 1. Empty is OK for the subject header so less than one is also ok. That's my take anyway.

packages/frontend/app/components/school-manager.js
jrjohnson: In these validators `string()` plus `required()` has the same effect as an empty string is either blank or has less than one character. So I removed it for brevity. As far as I could test this is identical to the current validation.

michaelchadwick (approved):

PR #8462, opened 2025-03-21
Add School Filter to Subject Reports

jrjohnson:
When All Schools is selected for a report add a school filter to the results so they can be refined.

Fixes ilios/ilios#4912

wip: I've only done competencies to validate the approach.

### DIFF
--- a/packages/frontend/app/components/reports/subject-header.hbs
+++ b/packages/frontend/app/components/reports/subject-header.hbs
@@ -36,5 +36,11 @@
   {{#if @showYearFilter}}
     <Reports::SubjectYearFilter @selectedYear={{@year}} @changeYear={{@changeYear}} />
   {{/if}}
+  {{#if @showSchoolFilter}}
+    <Reports::Subject::SchoolFilter
+      @selectedSchool={{@filterSchool}}
+      @changeSchool={{@changeFilterSchool}}
+    />
+  {{/if}}
   <Reports::SubjectDescription @description={{@description}} @resultsLength={{@resultsLength}} />
 </div>
--- a/packages/frontend/app/components/reports/subject-results.hbs
+++ b/packages/frontend/app/components/reports/subject-results.hbs
@@ -7,6 +7,8 @@
     @school={{@school}}
     @changeYear={{@changeYear}}
     @year={{@year}}
+    @filterSchool={{@filterSchool}}
+    @changeFilterSchool={{@changeFilterSchool}}
     @description={{@description}}
   />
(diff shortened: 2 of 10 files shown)


jrjohnson:
@dartajax please take a look at the subject competency report and see if this is the behavior you had in mind for this ticket. I'm not convinced we aren't better served by having users go back and choose a school to run the report on, but I imagine there are some users who want to toggle around. I'm ok either way, only competencies is done, but if it looks good to you let me know and I'll add this to the rest of the subject report types.

dartajax:
> @dartajax please take a look at the subject competency report and see if this is the behavior you had in mind for this ticket. I'm not convinced we aren't better served by having users go back and choose a school to run the report on, but I imagine there are some users who want to toggle around. I'm ok either way, only competencies is done, but if it looks good to you let me know and I'll add this to the rest of the subject report types.

I guess the drop-down would appear only if someone happens to run a report with `"All Schools"` selected and results from more than one school displayed (this second part is preferred by me but not required by any means). The drop-down selection is not activated yet - doesn't do anything but I bet you knew that. Your logic of choosing a school to run the report on seems valid to me too. I am not against that either. However, functionality as described in the ticket (when drop-down appears not that it is active) seems fine.

jrjohnson:
Forgot it was slightly different, you can try out the fully functioning dropdown in a saved report. Doesn't work yet in the "run report" type.

dartajax:
> Forgot it was slightly different, you can try out the fully functioning dropdown in a saved report. Doesn't work yet in the "run report" type.

That click event only runs the first time it is used - not click but select from drop-down using mouse. Also the count should update too.

dartajax:
I ran one for ...

`All Competencies for 2024 - 2025 in All Schools`

... where 2024 - 2025 was an academic year I selected - don't see the school filter there but I assume maybe because only one school was returned or that is out of scope for this PR is my other more likely assumption

dartajax:
oh yeah I guess we need to pre-pend the school for the output report of the one I specified above when acad year is selected - follow-up ticket possibly

dartajax (approved):
It works for me on saved reports of the "All Schools" "Competencies" "Anything" report - newly generated reports show the drop-down but there is no functionality there - no prob since that is not supposed to work with the PR I assume - also if a specific academic year is selected, the filters don't appear - again out of scope for this PR - so all good on what was specified at the beginning of this now lengthy comment.

jrjohnson:
This is super stale, going to close it for now.

PR #8463, opened 2025-03-21
Improved Course Loading

jrjohnson:
Adding a loading route for the courses list so it renders instantly without a delay. Also replaced the pulse loader on the course list with a version of the loading list instead.

### DIFF
--- a/packages/frontend/app/styles/components/courses/loading.scss
+++ b/packages/frontend/app/styles/components/courses/loading.scss
@@ -1,0 +1,15 @@
+@use "../../ilios-common/mixins" as cm;
+@use "../../ilios-common/colors" as c;
+
+.courses-loading {
+  select,
+  input {
+    @include cm.loading-text;
+  }
+
+  .toggle-mycourses .toggle-buttons label {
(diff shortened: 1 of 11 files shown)


packages/frontend/app/styles/components/courses/loading.scss
michaelchadwick, line 1: I know these are being "shimmered" since they aren't ready to be used until the courses are loaded, but I feel like the effect looks...off?...when compared to the table. I tried removing these styles and for the most part things looked and worked fine (the only issue is the `ToggleButtons` as they can be toggled while courses are loaded, and then they might snap to the other option depending on what's in the querystring).

This is all to say that this PR is a huge improvement, but could maybe be iterated on a little more. I'm cool with whatever the consensus is, though.

michaelchadwick (approved):
```
|==\      /\    |==\
|==/     /  \   |   \
|  \    /====\  |   /
|   \  /      \ |==/
```

dartajax (approved):
pretty slick - I like it

PR #8465, opened 2025-03-21
Use Corepack to Determine pnpm Version

jrjohnson:
Using the packageManager field in our package.json file activates corepack from Node and ensures we're always in alignment on what version of pnpm we're using and what is used in our workflows.

I've added a workflow that will update this to the latest version once a week.

I initially set the version to 10.0.0 so we'll have something to update right away.


stopfstedt (approved):
LGTM, thanks for the detailed description.

PR #8475, opened 2025-03-25
Use ember-keyboard addon for GlobalSearchBox

michaelchadwick:
The addon [ember-keyboard](https://adopted-ember-addons.github.io/ember-keyboard/) makes it a lot easier to explicitly denote which keyboard shortcuts you want a component to react to. I've used it on the `GlobalSearchBox` component as an example, but it could be used anywhere on the site.


packages/frontend/app/components/global-search-box.hbs
```diff
@@ -14,7 +14,10 @@
         (perform this.autocomplete)
       )
     }}
-    {{on "keyup" this.keyboard}}
+    {{on-key "Escape" this.onEscapeKey event="keyup"}}
```
jrjohnson, line 17: Any reason not to use the default `keydown` and remove passing the event type?

michaelchadwick: I looked right past that in the documentation. You're totally right. This also allows for "holding down the key to get warp speed arrow-age".

jrjohnson: Didn't even think of that, but holding down should absolutely work, nice!

PR #8478, opened 2025-03-26
Fix reporting double-<title> bug

michaelchadwick:
Fixes ilios/ilios#6078

The reporting area has an exception where it tries to set the page title twice due to the new Subject/Curriculum Reports switching. I fixed the `{{page-title}}` directives and added an exception to avoid a bug in how the `<title>` gets set.

### DIFF
--- a/packages/frontend/app/services/page-title.js
+++ b/packages/frontend/app/services/page-title.js
@@ -10,11 +10,6 @@ export default class HeaderTitleService extends EmberPageTitleService {
   @tracked title = '';
 
   titleDidUpdate(title) {
-    // exception to avoid double-titling in reports
-    if (title == 'Ilios Curriculum Reports Subject Reports') {
-      this.title = 'Ilios Curriculum Reports';
-    } else {
-      this.title = title;
-    }
+    this.title = title;
   }
 }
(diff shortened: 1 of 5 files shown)


jrjohnson (changes requested):
This solution will only work for English, but it's also not the best path forward.

What happens with `{{page-title` is i[t loads each route template](https://github.com/ember-cli/ember-page-title) in order so here it's loading `reports.hbs` and then `reports/subject.hbs`.

Instead of checking this value on update one or both of the `{{page-title` helpers should be modified to provide different data. One possible solution is to use [replace](https://github.com/ember-cli/ember-page-title?tab=readme-ov-file#page-title-helper) to drop the titles further up the chain.

michaelchadwick:
> This solution will only work for English, but it's also not the best path forward.

My English-centric brain dropped the ball on that one -_-

> Instead of checking this value on update one or both of the `{{page-title` helpers should be modified to provide different data.

I tried this approach first, but it was replacing more than I wanted. However, it's definitely a better solution, so will revisit.

jrjohnson (approved):

dartajax (approved):
works for me

dartajax:
tested French and Spanish too - all good once tests pass

PR #8480, opened 2025-03-27
Improve Loading on Instructors for Year

jrjohnson:
Loading all the users for a year with the API filter can take a long time with all of the joins. Instead I'm loading the courses for the year and then using our existing system of loading users for a course to get the data. Loading all the users for all the courses was also a bit of a resource hog so I chunked it up into 5 at a time requests. This seems to be the right compromise for optimal loading speed and reasonable request size.

Fixes ilios/ilios#6083


stopfstedt (approved):
LGTM

PR #8491, opened 2025-04-02
use the correct getter to get the start-date when rolling a course over.

stopfstedt:


jrjohnson (approved):

jrjohnson:
While it is only in a single browser the test failure here seems related. May need to change the way test data is calculated to avoid adding a flakey test.

stopfstedt:
> While it is only in a single browser the test failure here seems related. May need to change the way test data is calculated to avoid adding a flakey test.

handled. using  a string instead of a JS date for `startDate` on the course fixture seems to have done the trick. i copied and adjusted that change from the  `rollover course with new start date`  scenario in the same test.

PR #8494, opened 2025-04-03
Remove ember-mirage

jrjohnson:
Turns our we don't need this and can interact with mirage directly with very little needed in our apps. I've enhanced our existing setupMirage to startup and shutown using our config. There was a hidden `environment` option being created somewhere in ember-mirage that we needed to provide ourselves.


stopfstedt (approved):
nice one. approved!

PR #8498, opened 2025-04-08
Remove ember-event-helpers Addon

jrjohnson:
We're only using this in one place. It's a perfectly nice addon, but ancient and never going to be updated. Much easier to reach for the event itself and not add this complication.


michaelchadwick (approved):
Maybe we had higher hopes for using this at some point, but with only one instance so far, yup...good to yank.

PR #8502, opened 2025-04-08
redirect non-privileged users to 404 page on reports routes.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6090


michaelchadwick (approved):
Tried it out and it works great. However, `/reports/subject` doesn't seem to work, regardless. Just goes to a 404 even using my own account...?

PR #8505, opened 2025-04-10
Log Invisible Validation Errors

jrjohnson:
To prevent bugs such as we had with validating the session overview (#8492), where an error in a field that isn't currently open or showing errors prevents validation, this now logs those errors to the console. This isn't visible much to users, as it's in the console, but I used the intl service to translate them into nice errors anyway.


jrjohnson:
I'm keeping this in draft mode, not sure if it's a good idea. It makes tests noisy. Also, is it overly complicated?

jrjohnson:
We haven't had another repeat of this issue, closing this as not needed.

PR #8507, opened 2025-04-14
Adds better multi-school support to Curriculum Reports

michaelchadwick:
Fixes ilios/ilios#6028
Fixes ilios/ilios#6029

Curriculum Reports don't currently display anything that indicates a report spans more than one school, and downloads also don't include that information. This PR fixes that.

### DIFF
--- a/packages/frontend/app/components/reports/curriculum.js
+++ b/packages/frontend/app/components/reports/curriculum.js
@@ -41,6 +41,25 @@ export default class ReportsCurriculumComponent extends Component {
     return years.some((year) => year !== years[0]);
   }
 
+  get selectedSchoolIds() {
+    if (!this.selectedCourses) {
+      return [];
+    }
+    const schools = this.store.peekAll('school');
+    let schoolIds = [];
+    this.selectedCourses.map((course) => {
+      const schoolForCourse = schools.find((school) =>
+        school.hasMany('courses').ids().includes(course.id),
+      );
+
+      if (schoolForCourse) {
+        schoolIds = [...schoolIds, schoolForCourse.id];
+      }
+    });
+    //use a Set to remove duplicates
+    return [...new Set(schoolIds)];
+  }
+
   get reportResultsComponent() {
     switch (this.selectedReportValue) {
       case 'sessionObjectives':
--- a/packages/frontend/app/components/reports/curriculum.js
+++ b/packages/frontend/app/components/reports/curriculum.js
@@ -47,7 +47,7 @@ export default class ReportsCurriculumComponent extends Component {
     }
     const schools = this.store.peekAll('school');
     let schoolIds = [];
-    this.selectedCourses.map((course) => {
+    this.selectedCourses.forEach((course) => {
       const schoolForCourse = schools.find((school) =>
         school.hasMany('courses').ids().includes(course.id),
       );
--- a/packages/frontend/app/components/reports/curriculum.js
+++ b/packages/frontend/app/components/reports/curriculum.js
@@ -17,9 +17,6 @@ export default class ReportsCurriculumComponent extends Component {
   get passedCourseIds() {
     return this.args.selectedCourseIds?.map(Number) ?? [];
   }
-  get passedSchoolIds() {
-    return this.args.selectedSchoolIds?.map(Number) ?? [];
-  }
 
   get selectedReportValue() {
     return this.args.report ?? 'sessionObjectives';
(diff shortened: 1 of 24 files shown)


michaelchadwick:
Thanks for the approval, @dartajax. There's a better way to keep track of the schoolIds that doesn't require storing them in the URL like I currently do, so I'm going to refactor and ask for another review later.

dartajax:
I put a `DO NOT MERGE` on there for now to make sure I don't merge it too soon.

packages/frontend/app/components/reports/curriculum.js
```diff
@@ -41,6 +41,25 @@ export default class ReportsCurriculumComponent extends Component {
     return years.some((year) => year !== years[0]);
   }
 
+  get selectedSchoolIds() {
+    if (!this.selectedCourses) {
+      return [];
+    }
+    const schools = this.store.peekAll('school');
+    let schoolIds = [];
+    this.selectedCourses.map((course) => {
```
jrjohnson, line 50: `.map` returns the value. This could be `let schoolIds = course.map` or else `this.selectedCourses.forEach`.

Though what you maybe want here is a reduce. That way you can only add schools as they're needed and it's semantically better as you're reducing the list of courses to their school ids.

packages/frontend/app/components/reports/curriculum/learner-groups.hbs
```diff
@@ -31,6 +35,9 @@
       {{#if this.reportRunning}}
         {{#each (sort-by "courseTitle" @courses) as |c|}}
           <tr>
+            {{#if this.hasMultipleSchools}}
+              <td>{{11}}</td>
```
jrjohnson, line 39: This should be the name of the school for this course. Will required doing a local getter for `@courses` and then augmenting it with the school data, but since this can be looked up without async loading it makes more sense to have a real name here and it will space out better so when the load finished there won't be a re-flow of the table.

packages/frontend/app/components/reports/curriculum/session-objectives.hbs
```diff
@@ -31,6 +35,9 @@
       {{#if this.reportRunning}}
         {{#each (sort-by "title" @courses) as |c|}}
           <tr>
+            {{#if this.hasMultipleSchools}}
+              <td>{{11}}</td>
```
jrjohnson, line 39: Same comment as the `11` on learner groups report.

jrjohnson (changes requested):
Looks awesome, great tests, couple of small changes.

packages/frontend/app/components/reports/curriculum.js
michaelchadwick: Good catch on the `.map` vs `.forEach`. I tried changing this to a `.reduce`, but I need to make sure the school ids aren't duplicated, and couldn't find a nice way to keep track of that, so I feel like having an array of ids still works.

packages/frontend/app/components/reports/curriculum/learner-groups.js
```diff
@@ -135,9 +143,46 @@ export default class ReportsCurriculumLearnerGroupsComponent extends Component {
     return a.sessionTitle.localeCompare(b.sessionTitle);
   };
 
+  schoolTitle = (courseId) => {
+    const schools = this.store.peekAll('school');
+    const school = schools.find((school) => school.hasMany('courses').ids().includes(courseId));
+
+    return school ? school.title : '11';
```
stopfstedt, line 150: i don't understand this fallback - why `11`?

packages/frontend/app/components/reports/curriculum/session-objectives.js
```diff
@@ -128,9 +131,46 @@ export default class ReportsCurriculumSessionObjectivesComponent extends Compone
     return a.sessionTitle.localeCompare(b.sessionTitle);
   };
 
+  schoolTitle = (courseId) => {
+    const schools = this.store.peekAll('school');
+    const school = schools.find((school) => school.hasMany('courses').ids().includes(courseId));
+
+    return school ? school.title : '11';
```
stopfstedt, line 138: same here - why is it `11`?

michaelchadwick: This has been changed to 'School' and made into a getter.

stopfstedt (approved):
LGTM

dartajax (approved):
looks fine to me

PR #8512, opened 2025-04-18
Add labels for course visualizations that don't have data (yet)

michaelchadwick:
Fixes ilios/ilios#4328

Added some display labels when data isn't present for course visualizations.

### DIFF
--- a/packages/ilios-common/addon/components/course/visualizations.hbs
+++ b/packages/ilios-common/addon/components/course/visualizations.hbs
@@ -35,6 +35,7 @@
             @isIcon={{true}}
             @course={{@model}}
             @showDataTable={{false}}
+            @showNoChartDataError={{true}}
           />
         </LinkTo>
       </div>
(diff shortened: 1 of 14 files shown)


michaelchadwick:
Example course with no objectives, session types, vocabularies, or instructors
Screenshot 2025-04-18 at 8 21 25 AM

stopfstedt (approved):
looks good enough for me. really don't like further nesting conditionals within conditionals here to make this happen, but this may be addressed as a follow-up.

dartajax (changes requested):
I found the image below in Percy - not digging it. 

image

michaelchadwick:
@dartajax excellent find, Dave. I will fix this.

michaelchadwick:
NTS: Course Objectives, Visualizations Root, and Visualizations Detail all need to be logicked correctly.

dartajax (changes requested):
Slight change request (sorry about that) ...

I would suggest for the caption under Vocabularies that it read like "Instructors" stating `"No vocabularies have been linked to any sessions in this course."` rather than stating that the course has no sessions since this is probably not true.

dartajax (approved):
looks fine now

stopfstedt (approved):

PR #8520, opened 2025-04-22
use Yup for form validation in curriculum-inventory/new-report component.

stopfstedt:

not as straightforward as i thought this would be. turns out there is no drop-in replacement for the `@IsBlank` validation decorator. so I'm working around this by throwing [`trim`](https://github.com/jquense/yup?tab=readme-ov-file#stringtrimmessage-string--function-schema) into the YUP validation schema _before_ [`required`](https://github.com/jquense/yup?tab=readme-ov-file#schemarequiredmessage-string--function-schema) is declared in the attribute validation chain.
seems to work, but please give it a closer look during review/testing. thanks.


packages/frontend/app/components/curriculum-inventory/new-report.js
```diff
@@ -23,6 +23,11 @@ export default class CurriculumInventoryNewReportComponent extends Component {
     this.selectedYear = this.currentYear;
   }
 
+  validations = new YupValidations(this, {
+    name: string().trim().required().max(60),
```
michaelchadwick, line 27: The `CurriculumInventoryReportHeaderComponent` is currently being validated to require 3 characters minimum for a report title: https://github.com/ilios/frontend/blob/master/packages/frontend/app/components/curriculum-inventory/report-header.js#L9

There's no problem with the shift to Yup, but these two requirements don't match.

jrjohnson: I like trim, didn't know about it. I used [transform()](https://github.com/ilios/frontend/blob/4c7dc7e44b5ad963e27607a1962e7f5c14642991/packages/ilios-common/addon/components/course/overview.js#L42C8-L42C17) to do some processing on null as well.

stopfstedt: re: 3  char limit - i'm realigning it here: 736ea011add6f432bd868e31af17cccbedca091d

michaelchadwick (approved):

PR #8523, opened 2025-04-22
use Yup for form validation in curriculum-inventory/report-rollover component.

stopfstedt:
while at it, i removed the 3 chars minimum from the form validations on the name attribute.
this realigns the form validation with the corresponding new-report component.


packages/frontend/app/components/curriculum-inventory/report-rollover.hbs
```diff
@@ -19,10 +19,14 @@
             disabled={{this.save.isRunning}}
             placeholder={{t "general.reportNamePlaceholder"}}
             {{on "input" (pick "target.value" this.changeName)}}
-            {{on "keyup" (queue (fn this.addErrorDisplayFor "name") this.saveOnEnter)}}
+            {{on "keyup" this.saveOnEnter}}
```
jrjohnson, line 22: Did you leave the `{{this.validations.attach "name"}}` off on purpose because the field is pre-filled with data? Same for `description`? Otherwise you have to click save to get errors, which maybe is good in this case? Not sure.

stopfstedt: good catch. fixed. :heavy_check_mark:

jrjohnson (approved):

PR #8527, opened 2025-04-24
turn off UserNameInfo filter and always show info icon with more data

michaelchadwick:
Fixes ilios/ilios#6138


michaelchadwick:
There are reasons why this info is hidden, and so this is not going to be pursued at this time.

PR #8528, opened 2025-04-24
UserProfileBio Sub-Componetization

michaelchadwick:
Creates two sub-components for `UserProfileBio`: `UserProfileBioDetails` and `UserProfileBioManager`. Currently, the `UserProfileBio` component's template has markup for both viewing and editing a user profile bio's fields. Because of a bug, if you search for a user after loading one, it doesn't properly load the correct new user's data into the fields when `isManaging` is true.

Breaking up the components into a read/write pair forces the `UserProfileBioDetails` component to load every time, ensuring the correct data is available when switching to `Manager`-mode.

Fixes ilios/ilios#6139


packages/frontend/app/components/user-profile-bio-details.gjs
```diff
@@ -0,0 +1,178 @@
...
+import { service } from '@ember/service';
+import t from 'ember-intl/helpers/t';
+import { on } from '@ember/modifier';
+import { fn, uniqueId } from '@ember/helper';
+import FaIcon from 'ilios-common/components/fa-icon';
+
+export default class UserProfileBioDetailsComponent extends Component {
+  @service iliosConfig;
+
+  @tracked firstName;
```
jrjohnson, line 12: As far as I can tell none of these need to be tracked or setup anymore as this component is now read only and the template works off of the user so they can all go.

jrjohnson (changes requested):
One thing to remove, otherwise looks awesome! Good fix and I think it's way easier to understand this way.

packages/frontend/app/components/user-profile-bio-details.gjs
michaelchadwick: Indeed! I had to look at another `.gjs` file to find out how to remove all that and still have a functioning file. You actually end up taking out the whole `Component` part and just have `import`s and a `<template>`. It's still weird getting used to this format, but it works!

jrjohnson (approved):

stopfstedt (changes requested):
based on Dave's out-of-band feedback:

image (5)

the button should be _above_ the error message.

dartajax (approved):
back in business

PR #8534, opened 2025-04-25
Format Code in Tagging Workflow

jrjohnson:
When we tag the newline in the package.json files gets dropped, adding a step to run prettier and format these again. This requires installing pnpm and our packages. Just in case other stuff gets formatted and we don't want it as part of the commit I'm only adding and committing the two files we expect to change.


stopfstedt (approved):
LGTM

PR #8535, opened 2025-04-26
clear form validation errors on cancel.

stopfstedt:
while at it, appropriately scope down validation function calls to name attribute when saving name changes.

companion to https://github.com/ilios/frontend/pull/8532


michaelchadwick (approved):
Nice work

PR #8539, opened 2025-04-30
use Yup to validate curriculum-inventory/new-sequence-block component.

stopfstedt:


packages/frontend/app/components/curriculum-inventory/new-sequence-block.hbs
```diff
@@ -120,10 +135,14 @@
           type="text"
           value={{this.duration}}
           disabled={{this.save.isRunning}}
-          {{on "keyup" (queue (fn this.addErrorDisplayFor "duration") this.saveOrCancel)}}
-          {{on "input" (pick "target.value" (set this "duration"))}}
+          {{on "input" (pick "target.value" this.setDuration)}}
```
jrjohnson, line 138: `{{on "keyup" this.saveOrCancel}}` got dropped here.

packages/ilios-common/addon/classes/yup-validations.js
```diff
@@ -232,3 +235,37 @@ function moreThan() {
     };
   };
 }
+
+function lessThan() {
+  return (validationParams) => {
+    //our current translations expects this key to be named gt and not more as it is in yup
```
jrjohnson, line 241:
```suggestion
    //our current translations expects this key to be named lt and not less as it is in yup
```

packages/ilios-common/addon/classes/yup-validations.js
```diff
@@ -232,3 +235,37 @@ function moreThan() {
...
+    const lt = validationParams.less;
+    return {
+      path: validationParams.path,
+      messageKey: 'errors.lessThan',
+      values: { lt },
+    };
+  };
+}
+
+function minimum() {
```
jrjohnson, line 251: This one seems like It might not be needed, is it min in both places?

jrjohnson (changes requested):

packages/ilios-common/addon/classes/yup-validations.js
stopfstedt: [string.min()](https://yup-docs.vercel.app/docs/Api/types#stringminlimit-number--ref-message-string--function-schema)
 is different from [number.min()](https://yup-docs.vercel.app/docs/Api/types#numberminlimit-number--ref-message-string--function-schema) - min chars  vs lowest number in range. the error message is different, too.
 
agreed, the naming of these two functions is not ideal. as a follow-up item, i'd like to clean up the yup-validations class - perhaps inline all these one-off functions, or scope/namespace them better. not entirely sure yet.

jrjohnson (approved):

PR #8545, opened 2025-05-05
Added zebra striping to subject report detail route

michaelchadwick:
Fixes ilios/ilios#5241

### DIFF
--- a/packages/frontend/app/styles/components/reports/subject.scss
+++ b/packages/frontend/app/styles/components/reports/subject.scss
@@ -1,4 +1,3 @@
-@use "../../ilios-common/colors" as c;
 @use "../../ilios-common/mixins" as cm;
 
 .reports-subject {
@@ -7,6 +6,12 @@
   .content-container {
     .report-results {
       @include cm.ilios-list-reset;
+
+      li {
+        margin-bottom: 0;
+        padding: 0.25rem 0.1rem;
+      }
+
       @include cm.ilios-zebra-list;
     }
   }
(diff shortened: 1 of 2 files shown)


packages/ilios-common/app/styles/ilios-common/mixins/ilios-list.scss
```diff
@@ -34,6 +35,17 @@
   }
 }
 
+@mixin ilios-zebra-list() {
+  li {
+    margin-bottom: 0;
```
jrjohnson, line 40: I'm not sure this margin and padding belong in this mixin as it confuses the responsibility of color vs layout. Talk me into it?

packages/frontend/app/styles/components/reports/subject.scss
```diff
@@ -1,11 +1,15 @@
+@use "../../ilios-common/colors" as c;
 @use "../../ilios-common/mixins" as cm;
 
+@use "sass:color";
```
jrjohnson, line 4: Used anywhere?

michaelchadwick: Good catch. Vestige will be removed.

packages/ilios-common/app/styles/ilios-common/mixins/ilios-list.scss
michaelchadwick: You're absolutely right. Moved it out of here and into the component style instead.

jrjohnson (approved):

PR #8546, opened 2025-05-05
Add Copy Report button to Subject Report detail page

michaelchadwick:
Fixes ilios/ilios#4914

Created a link-disguised-as-button that "edits" a Subject Report by routing to a new report with the arguments of the saved report, effectively allowing you to "change" its arguments.

Update: changing mechanism from "Edit" to "Copy Report" so it makes more sense

### DIFF
--- a/packages/frontend/app/components/reports/new-subject.gjs
+++ b/packages/frontend/app/components/reports/new-subject.gjs
@@ -316,7 +316,7 @@ export default class ReportsNewSubjectComponent extends Component {
 
     const report = this.store.createRecord('report', {
       title: this.title,
-      user: this.userModel,
+      user: this.userModel ?? (await this.currentUser.getModel()),
       subject: this.subject,
       prepositionalObject: this.prepositionalObject,
       prepositionalObjectTableRowId: this.prepositionalObjectId,
--- a/packages/frontend/app/components/reports/new-subject.gjs
+++ b/packages/frontend/app/components/reports/new-subject.gjs
@@ -461,11 +461,7 @@ export default class ReportsNewSubjectComponent extends Component {
     {{#let (uniqueId) as |templateId|}}
       <div class="reports-new-subject" data-test-reports-new-subject ...attributes>
         <div class="title" data-test-component-title>
-          {{#if @editReport}}
-            {{t "general.editReport"}}
-          {{else}}
-            {{t "general.newReport"}}
-          {{/if}}
+          {{t "general.newReport"}}
         </div>
         <div class="new-subject-content">
           <p data-test-title>
--- a/packages/frontend/app/components/reports/new-subject.gjs
+++ b/packages/frontend/app/components/reports/new-subject.gjs
@@ -23,7 +23,6 @@ import { uniqueId, fn } from '@ember/helper';
 import t from 'ember-intl/helpers/t';
 import { on } from '@ember/modifier';
 import pick from 'ilios-common/helpers/pick';
-import set from 'ember-set-helper/helpers/set';
 import ValidationError from 'ilios-common/components/validation-error';
 import eq from 'ember-truth-helpers/helpers/eq';
 import sortBy from 'ilios-common/helpers/sort-by';
@@ -36,8 +35,6 @@ export default class ReportsNewSubjectComponent extends Component {
   @service intl;
   @service store;
   @service dataLoader;
-
-  @tracked selectedTitle;
 
   @tracked isSaving = false;
   @tracked schoolChanged = false;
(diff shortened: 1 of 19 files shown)


stopfstedt (approved):
LGTM.
cancel behaviour is now a bit wonky, since it always lands you back in list view and not in details even if you came from there . might be a worth a convo and follow-up item.

stopfstedt (approved):

dartajax (changes requested):
I believe the title of the report you decide to edit should appear in the "Report Title" field - shouldn't it?

If possible, it should probably say "Edit Report" rather than "New Report" as well. I love the button though.

michaelchadwick:
@dartajax I believe I've fixed the issues you brought up.

dartajax:
I hate to do this but if the title is one of the auto-generated titles, shouldn't the title refresh with new selection? I will team discussion it up until tomorrow since I don't even know if it is possible to determine whether a title was auto-generated or entered manually. What I am suggesting may be impossible - if so, we can dismiss this comment and I will approve my change request.

dartajax:
Maybe a "refresh title" button if user wants that to happen.

michaelchadwick:
@dartajax I feel like bringing in the existing title, even if it's auto-generated, and filling the "Report Title" value with it is good enough in this case. If the user changes a parameter, it won't update the title unless they blank it out and save again. Keeping track of whether it's auto-generated or not feels like unnecessary overhead. Thoughts?

dartajax:
> @dartajax I feel like bringing in the existing title, even if it's auto-generated, and filling the "Report Title" value with it is good enough in this case. If the user changes a parameter, it won't update the title unless they blank it out and save again. Keeping track of whether it's auto-generated or not feels like unnecessary overhead. Thoughts?

Well I am not surprised by this but when we perform the original save on these reports, we must check to see if the user has entered anything or else the saved title is generated by accepting the values specified by the drop-downs or search selection. We can't do that on an update? I'd think the same logic might be applied but I understand if that is a scope creep.

dartajax:
> > @dartajax I feel like bringing in the existing title, even if it's auto-generated, and filling the "Report Title" value with it is good enough in this case. If the user changes a parameter, it won't update the title unless they blank it out and save again. Keeping track of whether it's auto-generated or not feels like unnecessary overhead. Thoughts?
> 
> Well I am not surprised by this but when we perform the original save on these reports, we must check to see if the user has entered anything or else the saved title is generated by accepting the values specified by the drop-downs or search selection. We can't do that on an update? I'd think the same logic might be applied but I understand if that is a scope creep.

Ah but they can blank it out and save it again to auto-generate a new title? That might work. Let me test that.

michaelchadwick:
@dartajax Once the Report is saved, the auto-generated title _becomes_ the title, so its source is lost. Thus, when editing it's as if they named the report what the auto-generated title came up with. If they change params, they're changing params in a report with title "What Auto-Generate Created". We could add a flag to the report model that remembers if it was auto-generated, but yeah...it feels like major scope creep at this point.

dartajax:
Well I don't know still and I do apologize for being difficult. When you open an existing report, pick or search for new criteria, blank out the title in order to generate a new one, it does save but it also saves the old one so we are getting more and more reports. It is not doing an update but rather an add new. Not sure if I care that much but it should update the old one, no?

dartajax:
> Well I don't know still and I do apologize for being difficult. When you open an existing report, pick or search for new criteria, blank out the title in order to generate a new one, it does save but it also saves the old one so we are getting more and more reports. It is not doing an update but rather an add new. Not sure if I care that much but it should update the old one, no?

I can document my way around this but it should be pointed out to users that they are not saving and updating the old report but generating another one. The old one can be manually deleted then I guess.

michaelchadwick:
It's definitely an odd mechanism that is _like_ an edit, but not quite.

michaelchadwick:
@dartajax @stopfstedt Relevant points about this change have come up, so back to draft until we can come to a consensus as a team.

michaelchadwick:
NTS: Changed button/icon from "Edit" to "Copy Report". Don't bring report title if auto-generated (`title = null`), as it will naturally auto-generate again. If custom title is used, add ` (Copy)` to end of report title.

dartajax (approved):
I took a note of what @stopfstedt mentioned as well but I approve of this as is.

dartajax (changes requested):
image

any chance of making the fonts match a bit better? I am sorry about this and my overall pickiness

dartajax (approved):
thanks for checking and fixing those font weights @michaelchadwick - I approve now yet again

PR #8547, opened 2025-05-06
Template Tags Integration Tests

jrjohnson:
First step to a full application move to this format.
Start with moving to the `<template>` tag format and `.gjs` files for all of our
integration tests.


michaelchadwick:
Also, the whole `this` vs `context` thing is interesting, and _more_ interesting that it was only in two files.

PR #8277, opened 2024-12-18
Improve Ilios Loading Error div

michaelchadwick:
Currently, the styling of the `ilios-loading-error` `div` that gets inserted on the page if there is a loading error leaves much to be desired.

Screenshot 2024-12-18 at 3 51 05 PM

I decided to spruce it up, and hopefully everyone thinks it to be an improvement.

Screenshot 2024-12-18 at 3 49 14 PM

Although there are many ways to do it, for an example, I tested triggering the error by adding a `constructor` to [`packages/ilios-common/addon/components/course/rollover-date-picker.js`](https://github.com/ilios/frontend/blob/master/packages/ilios-common/addon/components/course/rollover-date-picker.js):

```
constructor() {
  super(...arguments);
  this.showPicker.perform({}, null);
}
```

PR #8548, opened 2025-05-07
Hide Ilios error overlay when testing

michaelchadwick:
While [improving](https://github.com/ilios/frontend/pull/8277) the rarely-seen-if-everything-is-going-fine `ilios-loading-error` overlay dialog was _good_ (it can't be ignored, and it usually means BAD THING HAPPEN), it can be a hindrance when running tests in the browser, as it just...sits there, obscuring the test information. This change adds CSS support back to the testing area and hides it. Errors still appear in the dev console, but the overlay won't hinder things anymore.

### DIFF
--- a/packages/frontend/public/assets/test-support.css
+++ b/packages/frontend/public/assets/test-support.css
@@ -1,0 +1,4 @@
+/* add styles here specifically for testing */
+#ilios-loading-error {
+  display: none !important;
+}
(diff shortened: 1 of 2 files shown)


jrjohnson (changes requested):
Big fan, but instead of adding this so far from it's source, I think adding a block to https://github.com/ilios/frontend/blob/master/packages/frontend/lib/ilios-error/index.js#L16 for `test-body` and putting the `display: none` there. Or adding a tests/style directory to that in-repo addon which will make it's way into `test-support` through the build is a better path.

stopfstedt (approved):
LGTM

jrjohnson (approved):

PR #8550, opened 2025-05-08
use Yup to validate learner-group management components.

stopfstedt:


dartajax (approved):
seems fine to me

packages/frontend/app/components/learner-group/root.gjs
```diff
@@ -218,12 +221,6 @@ export default class LearnerGroupRootComponent extends Component {
   });
 
   changeLocation = restartableTask(async () => {
-    this.addErrorDisplayFor('location');
```
jrjohnson, line 221: 👋

jrjohnson (approved):

PR #8551, opened 2025-05-08
Apply Template Tage Codemod to Components

jrjohnson:
Used both the convert and merge-history parts of this codemod to transform all of our components into `gjs` and preserve the blame for each line.

Some cleanup after the codemod and also discovered the [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view) file which then hides the codemod changes in blame view so it's easier to see other changes. Cool!


stopfstedt:
re: https://github.com/ilios/frontend/pull/8551/commits/99ea78964dfac1f7e47175ff0e1813fbaa1a285b

i wonder if we can set this right again with handlebar's whitespace control https://handlebarsjs.com/guide/expressions.html#whitespace-control 
will try this as a follow-up item.

stopfstedt (approved):

PR #8552, opened 2025-05-09
Validate Before Displaying Errors

jrjohnson:
When we show error messages because of a blur event we can sometimes show messages that aren't valid for the current state. By running the validator we ensure this doesn't happen.

The blur event that displays the validator only needs to fire when focus leaves the element the first time, on subsequent focusout we'd be doing extra validation work. Adding the once option here causes the event to be removed as soon as it is called, which is nice because it's one less thing.

Fixes ilios/ilios#6159

### DIFF
--- a/packages/ilios-common/addon/classes/yup-validations.js
+++ b/packages/ilios-common/addon/classes/yup-validations.js
@@ -128,9 +128,10 @@ export default class YupValidations {
     element.addEventListener(
       'focusout',
       () => {
+        this.#validate();
         this.makeErrorsVisibleFor.perform([field]);
       },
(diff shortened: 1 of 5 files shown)


stopfstedt (approved):
LGTM

packages/ilios-common/addon/classes/yup-validations.js
```diff
@@ -128,9 +128,10 @@ export default class YupValidations {
     element.addEventListener(
       'focusout',
       () => {
+        this.#validate();
```
stopfstedt, line 131: :dart:

PR #8553, opened 2025-05-10
use Yup to validate objective-list-item components and new-objective component.

stopfstedt:
couple of things going on here.

1. the @HtmlNotBlank decorator has been replaced with the imported [`striptags()`](https://www.npmjs.com/package/striptags). Yup validates against a wrapper getter around the actual description attribute. good enough.
2. i renamed the title attribute and associated methods to description, since this is what the corresponding input form is labeled at. we've been using a mix of title and description before, this consolidates the nomenclature as much as possible.
3. i yanked the code that was responsible for disabling the submit button in the edit-in-place component since we lost reactiveness here. I tried to get there via an "hasErrorFor()" in the YupValidations class, but it didn't fire on changes to the internal error tracking structure. Not a big loss here - the form still won't submit if validation does not pass.


michaelchadwick (approved):
Click-tested course, session, and program objectives. All seems good.

stopfstedt:
@dartajax  - for click testing purposes. this affects components that deal with creating and updating program-year/course/session objectives. should work as before, but the validation error messages verbiage has changes slightly.

dartajax (approved):
seems fine to me

PR #8564, opened 2025-05-14
Fix Subject Report parameter language loading refresh

michaelchadwick:
Fixes ilios/ilios#6170

Subjects and Objects were properties of the component, so they only got refreshed on load. Changed them to `get`ters and now they refresh on language change. Also fixed `userModelData` loading to be normalized with our current `TrackedAsyncData` pattern.


stopfstedt (approved):
excellent. getters FTW!

PR #8565, opened 2025-05-14
New User form now uses same validation and password strength meter as Edit User

michaelchadwick:
Fixes ilios/ilios#4303
Supersedes #8176

Had a lot of trouble actually getting a `PasswordValidator` component to work with the form, so I decided to finally make my life a lot easier and just make the schema the same for New and Edit User forms, using the existing form validation framework. I moved the styles into a shared SASS file, though, so all my work wasn't totally for naught.

But hey...over a year later this PR actually does something useful!


stopfstedt (approved):
LGTM.

as an aside - i'm still disliking the fact that we have to lug around browser-specific CSS pseudo elements, but checking the MDN docs it looks like they're still necessary.

anyway, this is good to go.

dartajax (approved):
works for me - I approve

PR #8566, opened 2025-05-15
use Yup to validate new-user component.

stopfstedt:


packages/frontend/app/components/new-user.gjs
```diff
@@ -3,49 +3,75 @@ import { cached, tracked } from '@glimmer/tracking';
...
   @tracked nonStudentMode = true;
-  @tracked showUsernameTakenErrorMessage = false;
+
+  validations = new YupValidations(this, {
+    firstName: string().ensure().trim().required().max(50),
+    middleName: string().ensure().trim().max(20),
+    lastName: string().ensure().trim().required().max(50),
+    campusId: string().ensure().trim().max(16),
+    otherId: string().ensure().trim().max(16),
+    email: string().ensure().trim().required().max(100).email(),
```
michaelchadwick, line 51: Whatever regex Yup is using for email is different than our [current one](https://github.com/ilios/frontend/blob/master/packages/ilios-common/addon/decorators/validation/is-email.js), and is looser. It allows me to put in `a@a` as a valid email address. When submitted to the backend, I get a 400 Bad Request error.

michaelchadwick (changes requested):
Yup's default email validation is broken and allows for bad requests to be submitted to the backend.

packages/frontend/app/components/new-user.gjs
stopfstedt: from https://github.com/jquense/yup?tab=readme-ov-file#stringemailmessage-string--function-schema

> WATCH OUT: Validating email addresses is nearly impossible with just code. Different clients and servers accept different things and many diverge from the various specs defining "valid" emails. The ONLY real way to validate an email address is to send a verification email to it and check that the user got it. With that in mind, yup picks a relatively simple regex that does not cover all cases, but aligns with browser input validation behavior since HTML forms are a common use case for yup.
>
> If you have more specific needs please override the email method with your own logic or regex:

i'll do just that then.

stopfstedt: since the email validation rules provided by Yup don't cut it, we'll have to use the ones provided by validator.js.
which is what the current validation library - class-validator - is using under the hood. 

these are _elaborate_ and cannot/should not be copy/pasted into our codebase.

adding validator.js as direct dependency and use [ `isEmail()`](https://github.com/validatorjs/validator.js/blob/master/src/lib/isEmail.js) from there.

michaelchadwick (approved):
```
[#####]
[ o o ]
[  ~  ]
[ \_/ ]
[=====]
```

PR #8567, opened 2025-05-16
singularize/pluralize character/characters in error messages.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6178


stopfstedt:
thinking about this some more - i'm starting to wonder if we should ever hit that 1 character minimum scenario. in that case, the error message should really be "is required", not "is too short".

michaelchadwick (approved):

PR #8568, opened 2025-05-16
Yupdate CourseHeader validation

michaelchadwick:
Refs ilios/ilios#6055

In order to avoid multiple error messages, I removed an obvious `required()` validator and used a `min().max()` approach that would still make it required and the correct length.

### DIFF
--- a/packages/ilios-common/addon/components/course/header.gjs
+++ b/packages/ilios-common/addon/components/course/header.gjs
@@ -31,9 +31,13 @@ export default class CourseHeaderComponent extends Component {
     this.courseTitle = this.args.course.title;
   }
 
-  validations = new YupValidations(this, {
-    courseTitle: string().trim().required().min(3).max(200),
-  });
+  validations = new YupValidations(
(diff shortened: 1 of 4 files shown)


packages/ilios-common/addon/components/course/header.gjs
```diff
@@ -2,17 +2,18 @@ import Component from '@glimmer/component';
 import { cached, tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { restartableTask } from 'ember-concurrency';
-import { validatable, Length, NotBlank } from 'ilios-common/decorators/validation';
+import { validatable } from 'ilios-common/decorators/validation';
```
jrjohnson, line 5: The `@validatable` decorator can go as well now that this is yuppified.

jrjohnson (changes requested):
One more thing to remove, and the description for the PR needs an update with the new approach.

jrjohnson (approved):

PR #8569, opened 2025-05-16
remove doubling up of form validation error messages on blank title.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6177

### DIFF
--- a/packages/frontend/app/components/curriculum-inventory/sequence-block-header.gjs
+++ b/packages/frontend/app/components/curriculum-inventory/sequence-block-header.gjs
@@ -24,7 +24,7 @@ export default class CurriculumInventorySequenceBlockHeaderComponent extends Com
   }
 
   validations = new YupValidations(this, {
-    title: string().required().min(3).max(200),
+    title: string().ensure().trim().min(3).max(200),
   });
 
   @action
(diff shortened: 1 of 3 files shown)


packages/frontend/app/components/curriculum-inventory/sequence-block-header.gjs
```diff
@@ -24,7 +24,7 @@ export default class CurriculumInventorySequenceBlockHeaderComponent extends Com
   }
 
   validations = new YupValidations(this, {
-    title: string().required().min(3).max(200),
+    title: string().ensure().trim().min(3).max(200),
```
michaelchadwick, line 27: Is using `ensure()...min(x)` over `required()` the preferred method over using the `abortEarly` option?

stopfstedt: i'm aiming to reduce the number of potential validation error messages with this.

blank and character count < 3 would now spew out the same message instead of separate ones.

abortEarly is still desirable.

michaelchadwick: I like this approach better. I updated my CourseHeader PR to use it, too.

michaelchadwick (approved):

PR #8573, opened 2025-05-19
School->Vocabs->Terms->New remove status message width restriction on large screens

michaelchadwick:
Fixes ilios/ilios#6157


jrjohnson (approved):

dartajax (changes requested):
I hate to do this but it's what I do ... sometimes. See the screen shot below. Horizontal alignment issue - can we line it up better?

image

dartajax (approved):

PR #8575, opened 2025-05-19
use Yup to validate user-profile-bio-manager component.

stopfstedt:
password validation is a bit weird around the edges - minimum length is 5 _and_ it cannot be a string containing only whitespace. unlike most other text data points, the password doesn't get trimmed on the server-side when writing, so we can't trim here.


packages/frontend/tests/integration/components/user-profile-bio-manager-test.gjs
```diff
@@ -409,7 +409,18 @@ module('Integration | Component | user profile bio manager', function (hooks) {
     assert.notOk(component.password.hasError);
     await component.password.set('');
     await component.save();
-    assert.ok(component.password.hasError);
+    assert.strictEqual(component.password.errors.length, 2);
```
michaelchadwick, line 412: I know you mentioned password validation being trickier than other fields. Is this why we get two error messages on a blank string instead of just one (`Password can not be empty`)?

stopfstedt: aye, i can't seem to short circuit this down to one error message when the input is blank. would require reconfiguring Yup as previously discussed.

michaelchadwick (approved):
Tried in vain to make it short-circuit on first error message, so that'll have to be a global `YupValidations` config later. Otherwise, good stuff!

PR #8581, opened 2025-05-21
corrects init on title buffer.

stopfstedt:
override the title buffer when the input's title changes.

fixes https://github.com/ilios/ilios/issues/6184


stopfstedt:
withdrawn in favor of https://github.com/ilios/frontend/pull/8580

PR #8582, opened 2025-05-21
corrects init on url and location buffer.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6185


stopfstedt:
withdrawn in favor of https://github.com/ilios/frontend/pull/8583

PR #8585, opened 2025-05-22
Disable ILM toggle if content exists

michaelchadwick:
Fixes ilios/ilios#4316

`<ToggleYesNo>` was only disabled conditionally in one other part of the site, but the component itself has no code to use a `@disabled` argument, thus this fixes the issue and adds it.

However, what should a disabled toggle look like? And what should the tooltip/message say to warn users when it's disabled?


saschaben:
Hmm. a couple points. First, this is going to wildly irritate any user who has activated the ILM toggle but needs to flip it back to a regular session. Next, the error message will need to state clearly all the values / attributes that need removal in order to deactivate the toggle. Ideally, the message will also include a "Click [here]() to remove all groups and instructors" type link in order to keep the users from having to navigate to user management and clickety click. That said, the visual toggle is fine, and we can discuss and address the above thoughts at jam.

PR #8586, opened 2025-05-22
Remove School Results Search Counts

jrjohnson:
As we work to optimize search we need to limit the returned results, making these counts misleading. Remove them.


michaelchadwick (approved):
💀 to result counts

PR #8592, opened 2025-05-27
use Yup to validate room prop.

stopfstedt:


michaelchadwick (changes requested):
Screenshot 2025-05-28 at 8 53 34 AM
Another instance of the "double error message" event. Demo doesn't fall prey to this particular data input, but it _does_ trigger if you put in something like `0.A`.

PR #8595, opened 2025-05-29
by default, only render the first validation error message.

stopfstedt:


stopfstedt:
the remaining two from the list above are legit lists of errors and will remain collections.

```
$ grep -r 'error' packages --exclude-dir=dist --exclude-dir=node_modules --include=*.js | egrep 'page' | grep collection
packages/frontend/tests/pages/components/learner-group/bulk-assignment.js:  finalErrorData: collection('[data-test-final-error-data] tbody tr', {
packages/frontend/tests/pages/components/login-form.js:  errors: collection('[data-test-error]'),
```

PR #8596, opened 2025-05-30
Fixed School Vocab Manager title init/cleanup

michaelchadwick:
Refs ilios/ilios#6186

Somehow this instance was not in the list from ilios/ilios#6186, but I noticed it while trying to fix something else.


stopfstedt (approved):
LGTM

PR #8602, opened 2025-05-31
Subject Report object types that use search inputs now focused when validation fails

michaelchadwick:
Fixed ilios/ilios#6216


packages/frontend/app/components/reports/new-subject.gjs
```diff
@@ -447,6 +447,15 @@ export default class ReportsNewSubjectComponent extends Component {
         select.classList.add('error');
         select.focus();
       }
+
+      const searchObjectTypes = ['course', 'instructor', 'learning material', 'mesh term'];
+
+      if (searchObjectTypes.includes(this.prepositionalObject) && !this.prepositionalObjectId) {
+        const search = document.querySelector('[data-test-search-box] input');
```
stopfstedt, line 454: the targeted element is not guaranteed to be there, and will break the site when not.

running the report with this configuration will blow up, for example:

image

stopfstedt (changes requested):

stopfstedt (approved):
LGTM

PR #8605, opened 2025-06-02
Add missing icon for unpublished ILMs on WaaG and single events

michaelchadwick:
Fixes ilios/ilios#4225

Added missing icon, but also added page-object for `FaIcon` and rewrote tests to use it.

Ended up not using the `FaIconStack` implementation that was written for this PR, but will add it to a separate one to keep concerns, uh, separate.


michaelchadwick:
Update on this: added the infrastructure for a `stacked` or `layered` Font Awesome icon, as well as an integration test for it.

Next: actually add the unpublished icon.

PR #8610, opened 2025-06-05
indicate inactive users with icon.

stopfstedt:
colors are all over the place as-is -in some cases this icon is rendered in red, in others it's grey. i went with grey for these on this first iteration. to be reviewed and decided on.

fixes https://github.com/ilios/ilios/issues/6140


dartajax:
I put `NEEDS TEAM DISCUSSION` label on this one so we can discuss what color to standardize the disabled account icon to be - this PR has a dark grey color rather than screaming red but we should discuss which is best and apply them everywhere. Unrelated account anomalies were showing up too - not related to this ticket - `user_id 4530`

stopfstedt:
> I'm good with merging this as is, great tests!
> 
> There is a lot of duplication of the `{{unless}}` block and I wonder if that would be easier to read as a single `<UserStatus @enabled={{user.enabled}} />` component?

great idea for a follow-up item. this will also help to scope the icon style a bit better - right now it just "hangs out" as component style without an actual corresponding component. 

incoming later today.

dartajax (approved):
I was able to test this a bit but not quite as much as would have been optimal. Search is out of whack currently but will be back soon to normal. I was able to view a few of the new disabled icons and it looks fine to me. I am approving this PR as a result. More telling, less yelling. I like it.

PR #8620, opened 2025-06-12
better placement of search box during learner-group user management

stopfstedt:
refs https://github.com/ilios/ilios/issues/3778

image


dartajax (approved):
I approve of this. I may have yet still another follow-up ticket but we'll see. This is nice.

jrjohnson (approved):

PR #8622, opened 2025-06-13
space saver - moves search box into section header.

stopfstedt:
refs https://github.com/ilios/ilios/issues/3778

image


dartajax (approved):
It's almost as if you knew what my follow-up ticket was going to be - this one. Nailed it. This is excellent.

jrjohnson (approved):
Took me a while to find the filter, that was moved... but I persevered!

PR #8636, opened 2025-06-21
render the name of the current learner group, not its top-level parent, in the component title.

stopfstedt:
the current learner group is the correct context to put on the screen.

fixes https://github.com/ilios/ilios/issues/6210


dartajax (approved):
Top level - bah. Not that you needed my approval for this, but I am glad to do the honors.

PR #8639, opened 2025-06-23
consider selected courses input when picking the right school in dropdown.

stopfstedt:
this fixes https://github.com/ilios/ilios/issues/6241

but doesn't add additional output mentioned in the ticket's comments (this is to be further discussed).


michaelchadwick (approved):
Wonderful change. Good test. Sweet.

PR #8641, opened 2025-06-24
escape user input before using it as regular expression.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6280


michaelchadwick (approved):
YHDW,S

PR #8644, opened 2025-06-25
display sessions associated with a given learner-group.

stopfstedt:
refs ilios/ilios#6187


stopfstedt:
> This is an awesome addition! However, as we all discussed previously, the potential for taking up a lot of vertical space due to many associated sessions makes me think this requires a default collapsible version on load, with the expanded version after a click.

expando added :heavy_check_mark:

michaelchadwick (changes requested):
Sorry to ask for more, but methinks that if there are no Associated Sessions, it should just show "None" and be expanded, without having to expand to see that message.

stopfstedt:
per team discussion today
1. change the component title to `Associated Courses`
2. tally up the number of linked courses and append them to the component title, e.g. `Associated Courses (5)`
3. if there are no associated courses/sessions, then only display the title `Associated Courses (0)`, without the expando-button.

michaelchadwick (dismissed):
Minor class style change noted, but I also have a major functionality change request: track whether this component is expanded in the `LearnerGroupController` and pass setter/getter methods down to `CourseAssociations` so that you can click on a course/session and then hit the browser back button and still have it expanded (currently it always defaults to collapsed)

stopfstedt:
> but I also have a major functionality change request: track whether this component is expanded in the LearnerGroupController and pass setter/getter methods down to CourseAssociations so that you can click on a course/session and then hit the browser back button and still have it expanded (currently it always defaults to collapsed)

handled. please re-review, thanks!

michaelchadwick (approved):
Thanks for adding that extra URL functionality! Love this new addition.

PR #8646, opened 2025-06-27
corrects lang string key.

stopfstedt:
better scopes toggle-all element in component page object while at it.

fixes https://github.com/ilios/ilios/issues/6286


michaelchadwick (changes requested):
I'm a fan of both changes, but they don't seem to be related, despite being in the same component. I think this should be two PRs.

stopfstedt:
> I'm a fan of both changes, but they don't seem to be related, despite being in the same component. I think this should be two PRs.

i yanked the changes that reorganized the page object structure, and everything that came with it.
now, it's just the lang key correction and the associated test case additions.

PR #8647, opened 2025-06-27
ensure that the program year model is always loaded in route.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6294


michaelchadwick (approved):
Great catch and fix.

PR #8648, opened 2025-06-27
consider learner-groups associated with ILM when calculating session-to-learnergroup totals.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4945


dartajax (approved):
I tested this as well. Looks good to me.

PR #8654, opened 2025-07-01
V2 Search API

jrjohnson:
Align our UI with the new search API in https://github.com/ilios/ilios/pull/6295 where schools and years are part of filtering on the server and not on the frontend, autocomplete is removed, and the API version is bumped.

With this new interface we get much faster search results at a significant performance improvement on Opensearch and much better pagination.


stopfstedt (approved):
LGTM

PR #8665, opened 2025-07-03
Make all `<title>`values unique

michaelchadwick:
Fixes ilios/ilios#6304


stopfstedt (dismissed):
great improvement. 
i clicked through most (all?) screens, and only found two pages in the admin console where the page title still seems static - that's Updates from the Campus Directory and Students Requiring Cohort Assignments.

@michaelchadwick - please add prefixes to the page title on these as well.

image

image

jrjohnson (approved):

PR #8670, opened 2025-07-07
Curriculum Reports download now has correct school column header and value

michaelchadwick:
Fixes ilios/ilios#6318

### DIFF
--- a/packages/frontend/app/components/reports/curriculum/learner-groups.gjs
+++ b/packages/frontend/app/components/reports/curriculum/learner-groups.gjs
@@ -89,7 +89,7 @@ export default class ReportsCurriculumLearnerGroupsComponent extends Component {
   get summary() {
     return this.reportWithLearnerGroups.map((c) => {
       return {
-        schoolTitle: c.school?.title,
+        schoolTitle: c.school.title,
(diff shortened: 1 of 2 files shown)


packages/frontend/app/components/reports/curriculum/learner-groups.gjs
```diff
@@ -89,7 +89,7 @@ export default class ReportsCurriculumLearnerGroupsComponent extends Component {
   get summary() {
     return this.reportWithLearnerGroups.map((c) => {
       return {
-        schoolTitle: c.school.title,
+        schoolTitle: c.school?.title,
```
stopfstedt, line 92: @michaelchadwick - these changes don't track. the backend requires school title to be at [least one character long](https://github.com/ilios/ilios/blob/ed369b294bb831a69989b45c0120e16c4a1e2601/src/Entity/School.php#L57),  so i don't see why we would treat this as an potentially nullish value.

michaelchadwick: Sounds good. Changed to non-potential-nullish.

jrjohnson (approved):

PR #8673, opened 2025-07-08
Fixed incorrect translation key capitalization method

michaelchadwick:
Fixes ilios/ilios#6324

https://github.com/adopted-ember-addons/ember-cli-flash?tab=readme-ov-file#arbitrary-options


stopfstedt (approved):
LGTM

PR #8674, opened 2025-07-09
Label Interactive Elements

jrjohnson:
Each of these has visual clues, but not labels to say what they do. For accessibility we need both. Most got an aria-label, but in some cases I thought a title with a tooltip would work better.

Refs ilios/ilios#6322


stopfstedt (approved):
LGTM

PR #8675, opened 2025-07-10
Adjust Header Levels

jrjohnson:
These headers weren't at the correct level based on their position in the DOM, for accessibility headers should descend in order and not skip any steps.

Refs ilios/ilios#6322


packages/frontend/app/styles/components/user-profile-cohorts.scss
```diff
@@ -4,6 +4,10 @@
   .user-profile-cohorts-header {
     @include m.detail-container-header;
 
+    h2 {
+      @include m.ilios-heading-h3;
+    }
```
michaelchadwick, line 9: There are several examples like this, but just picking one of them: I'm assuming it was easier to do it this way, but there is cognitive dissonance seeing a mixin called `ilios-heading-h3` applied to an `h2` element.

jrjohnson: Interesting. Would it make more sense to see an include header-style. And a second include font-size(xl)? I read these as style this h2 like and h3. But I'm good with the other approach. I think we have this all over though so I'd rather take this as an item for discussion and follow up with a PR for the new pattern everywhere.

michaelchadwick: I understand how you could read this particular CSS declaration that way, but it doesn't make the same sense to me as the following:
```
h2 {
  @include m.ilios-heading-h2;
}
```

jrjohnson: Yes. I agree! Just exploring what that change would look like. Because we're not using the h2 size in most places. It's h2 for accessibility, but we want it smaller in practice. I like being explicit, but don't want to be too verbose.

michaelchadwick: OK, cool. I think the _real_ change is to re-evaluate our H1->H6 hierarchy across the board, and re-label things accordingly, and then update our mixins to match.

However, that sounds like a LOT of work, and I'm not sure it's worth it at this PR juncture. I'm curious what @stopfstedt has to say about this.

jrjohnson: This PR fixes all of our headers in HTML to be the right semantic choice, but then the default styles are wrong for the context visually. So I applied a different header style to them. I *think* to get at the core of your question we'll need to replace our `header-h*` mixins with something else, I have no problem with this, but it will need to be a followup PR for sure to do it everywhere.

michaelchadwick (approved):
Just to reiterate: this is a good change, the cognitive dissonance of the HTML tag and a mixin with different header versions notwithstanding.

dartajax:
image

dartajax:
This PR has conflicts anyway so not ready yet.

jrjohnson:
Nice find, pushed a fix (and the merge conflict fix as well).

PR #8676, opened 2025-07-10
Limit Integration a11y Tests

jrjohnson:
Setting a global run option for all integration tests to disable this color contrast check. For many of our components they're styled in place within the context of other components and inherit those styles and background. This makes tests of the color contrast here fail, we'll rely on acceptance tests to catch this issue.

Refs ilios/ilios#6322


michaelchadwick (approved):
CB,Y

PR #8677, opened 2025-07-10
Improve Print Course Accessibility

jrjohnson:
We had two issues here, one - the passed in templateId was unique to the course, but not to the session so we had many elements with the same id, one for each session. Second - we weren't actually passing anything, there was a variable name mismatch.


michaelchadwick (approved):
RJ, D!

PR #8678, opened 2025-07-10
Eighty Six Code Climate

jrjohnson:
Bye bye dear friend, value was high, but migration to their new service is annoying. Maybe we'll bring this back someday, but for now, toodles.

Fixes ilios/ilios#6317


stopfstedt (approved):

michaelchadwick:
I love how this change doesn't pass the Code Climate check ;P Also, love that someone else uses the term "86" to mean getting rid of something!

PR #8680, opened 2025-07-11
Replace Term Picker Interaction

jrjohnson:
Tying a click to a list item isn't a good idea, replaced everything with buttons more suited to the purpose.


michaelchadwick (changes requested):
Code looks good, but the visual representation could be improved.

Current PR change:
Screenshot 2025-07-11 at 9 02 02 AM

Improvement #1 (Remove gray list background):
Screenshot 2025-07-11 at 9 05 52 AM

Improvement #2 (Change color of X because red-on-dark-blue is not very contrasty:
Screenshot 2025-07-11 at 9 06 02 AM

PR #8687, opened 2025-07-15
Replace sass Color Variables with css Custom Properties

jrjohnson:
One step closer to plain css and easier to see what's going on as the name gets output and can be read in the inspector.

Recommend reviewing in individual commits, most of the changes are find/replace.


packages/frontend/app/styles/components/api-version-check.scss
```diff
@@ -1,10 +1,6 @@
 @use "../ilios-common/colors" as c;
 @use "../ilios-common/mixins" as m;
 
-.api-version-check-warning {
```
stopfstedt, line 4: just double-checking on this style class removal - is this intentional?

jrjohnson: Yes. Forgot to put that in the commit. We're not using it anywhere.

PR #8689, opened 2025-07-15
Re-add some important QUnit Theme CSS changes to testing

michaelchadwick:
Refs ilios/frontend#8668

Ever since we removed the custom QUnit Theme, I have missed it! This uses the native Ember `test-support.css` hook to add some of the colors and spacing fixes that were most useful (imo). Hopefully it still passes a11y mustard.

### DIFF
--- a/packages/frontend/public/assets/test-support.css
+++ b/packages/frontend/public/assets/test-support.css
@@ -1,0 +1,66 @@
+:root {
+  --color-gray-100: #f4f6f8;
+  --color-gray-700: #42474f;
+  --color-gray-900: #1c1e24;
+  --color-red: #e84646;
+  --color-green: #41c351;
+  --color-success: #d9f9e3;
+  --color-danger: #ffd8e1;
+
+  --spacing-2: 1rem;
+}
+
+#qunit {
+  margin: 0;
+  padding: 0 !important;
+}
+
+#qunit-header {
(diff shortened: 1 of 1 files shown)


packages/frontend/public/assets/test-support.css
```diff
@@ -0,0 +1,66 @@
+:root {
```
jrjohnson, line 1: These need to be scoped to the test container at least.

jrjohnson (changes requested):
I'd much rather see this go into the add-on upstream not into Ilios directly. If we're going to add it to Ilios all the color's probably need to go. Or some investigation into why it was preventing failures from showing in our a11y tests.

michaelchadwick:
I agree that this ideally would go upstream, but that seems like a heavier lift. What if we change the colors to use Ilios colors? :-D

jrjohnson:
The problem is I don't know why this broke our a11y tests in the first place, I suspect color, but I'm not sure. My guess is as soon as you can sort out and prove the fix it will be really easy to get it merged upstream. Before that I'm opposed to bringing anything back here that might get us in trouble again. At the very least I'd want to see a way to test our tests by expecting a failure and checking for it in some way.

michaelchadwick:
Since this is really just something for me, I'm closing this.

PR #8690, opened 2025-07-16
Assorted a11y Fixes

jrjohnson:
Last couple of stragglers on the frontend tests when the full a11y suite is enabled.

### DIFF
--- a/packages/ilios-common/addon/components/leadership-manager.gjs
+++ b/packages/ilios-common/addon/components/leadership-manager.gjs
@@ -56,68 +56,74 @@ import UserStatus from 'ilios-common/components/user-status';
         <tr>
           {{#if @showDirectors}}
             <td class="text-top" data-test-directors>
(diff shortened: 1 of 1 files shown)


packages/frontend/app/components/reports/curriculum/header.gjs
```diff
@@ -160,9 +160,9 @@ export default class ReportsCurriculumHeader extends Component {
         {{/if}}
         {{#if @showReportResults}}
           {{#if @loading}}
-            <button type="button" class="done text">
+            <div type="button" class="loading-results">
```
michaelchadwick, line 163: I'm not sure I've really seen `type=button` applied to a `<div>` before, and I don't think we have that anywhere else, so does it need that attribute for something containing a non-interactive spinner icon?

michaelchadwick (approved):
👍

PR #8694, opened 2025-07-17
Fix Update Notification Styling

jrjohnson:
Along the way this got a bit out of whack. Back to basics because it now just sits in the dom and doesn't need much.

Current:

current

New:

new


stopfstedt (approved):
nice! LGTM.

PR #8696, opened 2025-07-18
Global Search Box Style Adjustments

jrjohnson:
I went in to make sure the icon was off sufficient size to click, but I was able to simplify the styles overall and remove no longer needed autocomplete styles along the way. Slightly more accessible, simpler styles, very similar look and feel, but now the search icon is more centered in the box at all sizes.

Refs ilios/ilios#6322

I believe I didn't stomp on #8133, but linking to it for testing.


jrjohnson:
Something about the way I organized the styles didn't minify well, I reorganized them and things should look a lot better now.

jrjohnson:
I've re-aligned the header to work better at smaller sizes, the image wasn't constrained causing the overflow and the search input was getting a browser default size of 20. Fixing both required some adjustments, hopefully all good.

michaelchadwick (changes requested):
This is awesome now. One more minor quibble: there is no space between the search input and the toolbar once you go under 400px.

If you were to add the following to the `.global-search-box` section of https://github.com/ilios/frontend/blob/master/packages/frontend/app/styles/components/ilios-header.scss, then it would have that spacing, but not sure if this is _Ilios-style_ or not:

```
@include m.for-phone-only {
  margin-bottom: 4px;
  max-height: 27px;
}
```

jrjohnson:
As will happen I found a hundred other things while fixing this one. I've now inserted the space, but I also had to start from scratch with the logo and ended up replacing it entirely at smaller screens. Slapping a needs team discussion on this for review at our check in tomorrow before moving it forward.

michaelchadwick:
Dude! This is **sick** now. All the work you've put in is clearly shown in the result. Thanks for sticking with it.

michaelchadwick (changes requested):
Buzz harshing coming in, unfortunately: below 768px, when you focus on the search input, the element outline goes outside the input. Tested in all browsers.

Screenshot 2025-07-22 at 4 57 38 PM

Looks like using `margin-top` instead of `padding-top` fixes it in https://github.com/ilios/frontend/pull/8696/files#diff-66dd304f076c73c3b85e71d1293346dd2d9c09283ee7985383b7a64fcef0eac6R60-R68, but then the input needs a `max-height` at that resolution and below to get it back to the shape you had it before switching that attribute out.

jrjohnson:
@michaelchadwick thanks, ready again. I ended up using the grid gap to create the space.

packages/frontend/app/styles/components/global-search-box.scss
```diff
@@ -8,6 +8,7 @@
   align-items: center;
   max-width: 40rem;
   border-radius: 3px;
+  flex-direction: row-reverse;
```
michaelchadwick, line 11: I've never actually used any `-reverse` flex-ing before, so neat to see this being useful.

PR #8699, opened 2025-07-21
display courses associated with program years in new component

stopfstedt:
replaces the current component that lists associated courses with a new one that's modeled after the one used for learner group management.

refs https://github.com/ilios/frontend/pull/8644


michaelchadwick (dismissed):
Looks good, but the courses are default sorting by `id`, not `title`, which is unlike how they're sorted on your previous version for Learner Groups. They have to be clicked as the sort column to properly sort them by `title`.

PR #8702, opened 2025-07-22
Add performance acceptance tests

michaelchadwick:
Fixes ilios/ilios#6340


stopfstedt:
> For this to be truly useful, I should probably add more "loading" to things like Courses and Reports so they mimic actual sizes of data.

thanks for elaborating. i think i'm starting to see the light.

PR #8710, opened 2025-07-25
Drop the Color Change Utility

jrjohnson:
We can get the same effect through css alone.


packages/test-app/tests/integration/components/ilios-calendar-event-month-test.gjs
```diff
@@ -25,7 +25,7 @@ module('Integration | Component | ilios calendar event month', function (hooks)
     });
     assert.dom(s).hasStyle({
       'border-left-style': 'solid',
-      'border-left-color': 'rgb(0, 173, 86)',
+      'border-left-color': 'color(srgb 0 0.5 0.247549)',
```
stopfstedt, line 28: naturally.

stopfstedt (approved):
visually compared in the dashboard calendar/all schools, looks identical.
LGTM.

jrjohnson:
This isn't the time for this, we need to wait for full support for the relative color syntax when we drop safari 17.

PR #8711, opened 2025-07-25
Choose Calendar Color Based on Session Type

jrjohnson:
Session type colors are user selected across any range, we can't rely on them to be high contrasting with our text colors. Instead we need to determine the relative luminosity of the background color and then choose a light or dark alternative.

There is going to be a css way to handle this in the future, but for now, this is the best I could find.

Sample:
Screenshot 2025-07-25 at 8 55 26 AM


jrjohnson:
I know. I actually spent a while at https://www.npmjs.com/package/color and at the GitHub repo making sure it was a real library. Felt like a trap. Seems legit though even though the authors profile photo is obvious AI chicanery.

PR #8717, opened 2025-07-26
Course Print Preview Overhaul

michaelchadwick:
This PR is to give the `/courses/[course_id]/print` route a visual and a11y overhaul.

Thus far:
* Inline header elements that weren't actually inlined
* Improve consistency in spacing for `static-list` elements
* Remove `Actions` column as it's not valid in this context
* Improve Directors display in header
* Add missing testing element attributes
* Stop creating empty lists
* Stop creating empty `<div class="content">` elements
* Make Session Offering Instructors flush with header
* Generally tidy up spacing here and there

### DIFF
--- a/packages/ilios-common/app/styles/ilios-common/mixins/ilios-list.scss
+++ b/packages/ilios-common/app/styles/ilios-common/mixins/ilios-list.scss
@@ -25,17 +25,17 @@
   li {
     font-weight: bold;
 
+    ul,
+    ol {
+      margin-left: 1em;
+
+      li {
+        font-weight: normal;
+      }
+    }
+
     &:last-of-type {
       margin-bottom: 0;
-    }
-  }
-
-  ul {
-    margin-left: 1em;
-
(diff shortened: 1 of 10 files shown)


dartajax (approved):
the changes I have noticed are good ones - step in the right direction - fine with me

packages/ilios-common/app/styles/ilios-common/mixins/ilios-list.scss
```diff
@@ -19,16 +19,21 @@
...
+
+    &:last-of-type {
+      margin-bottom: 0;
+    }
   }
 
   ul {
     margin-left: 1em;
 
+    /* stylelint-disable-next-line no-descending-specificity */
```
jrjohnson, line 36: This points to an issue in the overall structure of the styles here and usually causes problems. Refactor to remove the need to this please!

jrjohnson (changes requested):
Awesome! One comment.

PR #8721, opened 2025-07-29
Use replaceState instead of pushState for querystring updates

michaelchadwick:
Fixes ilios/ilios#6350

### DIFF
--- a/packages/ilios-common/addon/routes/weeklyevents.js
+++ b/packages/ilios-common/addon/routes/weeklyevents.js
@@ -11,6 +11,9 @@ export default class WeeklyeventsRoute extends Route {
     week: {
       replace: true,
     },
+    year: {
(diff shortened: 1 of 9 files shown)


packages/ilios-common/addon/routes/weeklyevents.js
```diff
@@ -11,6 +11,9 @@ export default class WeeklyeventsRoute extends Route {
     week: {
       replace: true,
     },
+    year: {
+      replace: true,
```
jrjohnson, line 15: This is the only one I'm not sure about, feels like back should navigate you to the year before and not back to the dashboard here.

michaelchadwick: I can get behind this.

PR #8723, opened 2025-07-30
Fix All Weeks Year Calculation

jrjohnson:
DateTime.fromObject uses a default of january 1st of no day or month are specified. In a year with 53 weeks january 1st falls into the last week of the previous year. By using year in this calculation we were causing the year after a year with 53 weeks to also, but incorrectly, have too many weeks. Luxon provides a tool to do this correctly, we just needed to use it.

Fixes ilios/ilios#6366


michaelchadwick (changes requested):
Not sure if it's just me, but this change does not fix the issue. I can still trigger the error by going to 2027, 2033, etc.

michaelchadwick (approved):
EVERYTHING IS FINE

PR #8725, opened 2025-07-30
Add Fallback for User Menu

jrjohnson:
Sometimes, in some situations, the username doesn't fully load and we end up with a blank here. This may only happen with Siteimprove tests, but in case it's ever visible to users we now fall back to a generic User menu instead of a blank.


jrjohnson:
I couldn't find a way to test. For any logged in user we have their model and name without network traffic. I'm not sure what siteimprove is doing to get into this state, I think it's possibly a render bug. I could possibly mock a failure to get the user in at test? I'll try that.

michaelchadwick (approved):

PR #8729, opened 2025-07-31
rm unnecessary top border.

stopfstedt:
this one was causing doubling up of the border between components.


michaelchadwick (approved):
Single or nothing

PR #8731, opened 2025-07-31
Fix My Profile->Permissions header size and content

michaelchadwick:
Fixes https://github.com/ilios/ilios/issues/6375, https://github.com/ilios/ilios/issues/6376


michaelchadwick:
Agh, no on this one, too. Ignore!

PR #8732, opened 2025-07-31
My Profile->Permissions header size/content and a11y fixes

michaelchadwick:
Fixes ilios/ilios#6375, ilios/ilios#6376, ilios/ilios#6377

This fixes the following issues with My Profile->Permissions:
1. Headers were not consistent size both when there was content or not
2. Headers can not be followed by empty content so they were switched out for `<span>`
3. Header buttons had unneeded `aria-label` attributes and were causing visible/accessible mismatches when there was content

### DIFF
--- a/packages/frontend/app/components/user-profile-permissions.gjs
+++ b/packages/frontend/app/components/user-profile-permissions.gjs
@@ -525,11 +525,6 @@ export default class UserProfilePermissionsComponent extends Component {
           {{#if this.directedPrograms.length}}
             <h3 class="title" data-test-title>
               <button
-                aria-label={{if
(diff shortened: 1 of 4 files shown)


dartajax (changes requested):
The Permissions label got much smaller that is was before - not feeling it since the other labels are still at the original size.

before this PR ...
image

in the PR build ...
image

It is not quite as dramatic in these screen shots thanks to the zooming in or out on the screen captures but I think it is evident upon review.

dartajax (approved):
it looks fine now as far as I can tell - the issue I reported has been fixed - and the other one too - approval granted

PR #8733, opened 2025-07-31
show direct learner-group membership count instead of totals in offering form

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6363


michaelchadwick (approved):
WTG

PR #8736, opened 2025-08-01
Add headers to Global Search filter groups

michaelchadwick:
Fixes ilios/ilios#6385

Screenshot 2025-08-01 at 3 45 01 PM


jrjohnson (changes requested):
Has test errors

PR #8739, opened 2025-08-04
Replace Froala with Quill

michaelchadwick:
Fixes ilios/ilios#6353

The Quill is mightier than the Froala, or something.

Still have a hojillion tests to fix, but this is a Proof of Concept that works pretty well as a replacement. The only thing I still need to add, feature-wise, is a more dynamic, local CSS load (it's currently global on the `index.html` file), and the pre-chosen links for Insert Link.

Feel free to mess around with it and find where it breaks!


dartajax:
This is our current froala config with the inactive undo and redo buttons up higher on the same level as Bold, Italic and the other controls - also with the white background rather than blue.

<FROALA>
image
</FROALA>

<QUILL>
image
</QUILL>

In my testing, undo using Quill erases the entirety of the text entered, which could be disastrous if "saved".

michaelchadwick:
@dartajax the blue background and undo/redo issues should be fixed.

@jrjohnson I also threw in the "hide insert link popup if you click away from it" fix.

packages/ilios-common/addon-test-support/ilios-common/helpers/quill-editor.js
```diff
@@ -0,0 +1,51 @@
+import { findOne } from 'ember-cli-page-object/extend';
+import { loadQuillEditor } from 'ilios-common/utils/load-quill-editor';
+import { later } from '@ember/runloop';
+
+export async function fillInQuillEditor(element, html) {
+  const editor = await getEditorInstance(element);
+  editor.setContents(editor.clipboard.convert({ html }));
+  // editor.undo.saveStep();
```
jrjohnson, line 8: ??

packages/ilios-common/addon-test-support/ilios-common/helpers/quill-editor.js
```diff
@@ -0,0 +1,51 @@
...
+}
+export async function quillEditorValue(element) {
+  const editor = await getEditorInstance(element);
+  return editor.root.innerHTML.split('  ').join(' &nbsp;').replaceAll('<p><br></p>', '');
```
jrjohnson, line 12: This needs a comment or link. I'm guessing this is the only way to get at the value here, but it seems weird to reach into the HTML and that there isn't an API in quill to get the current value of the editor.

packages/ilios-common/addon/components/html-editor.gjs
```diff
@@ -1,111 +1,293 @@
...
-        this.html.set(component.args.content);
-        if (component.args.autoFocus) {
-          this.events.focus();
+      const { QuillEditor } = this.loadQuillData.value;
+      this.editor = new QuillEditor(element, options);
+      this.editor.setContents(this.editor.clipboard.convert({ html: this.args.content }));
+      if (this.args.autofocus) {
+        this.editor.focus();
+      }
+      this.loadFinished = true;
```
jrjohnson, line 34: Wondering if this is still needed now that the loading and setup isn't in a callback?

packages/ilios-common/addon/components/html-editor.gjs
```diff
@@ -1,111 +1,293 @@
...
+
+      this.editor.on('text-change', () => {
+        if (!this.isDestroyed && !this.isDestroying) {
+          this.editorHasNoRedo = !this.editor.history.stack.redo.length;
+          this.editorHasNoUndo = !this.editor.history.stack.undo.length;
+          // make sure to retain multiple spaces
```
jrjohnson, line 40: I agree this is what the code does, but maybe a deeper comment on why is needed? Why not just let quill collapse spaces?

packages/ilios-common/addon/utils/load-quill-editor.js
```diff
@@ -0,0 +1,22 @@
+export async function loadQuillEditor() {
+  const { default: QuillEditor } = await import('quill');
+
+  // Quill doesn't include redo/undo icons by default, so gotta hack 'em in
```
jrjohnson, line 4: Could we use what's in font awesome for this? I'm not worried about style yet, but I don't love carrying some random SVG stuff here.

jrjohnson (changes requested):
Some code changes, two behavior things I noticed:
1. Needs to expand to fill full space:
Froala:

Image

Quill:

Image

And when the add link popup is open it's hard to close. Needs a x button to click, but I think escape in the input should also close it. Clicking outside works, that's nice.

packages/ilios-common/addon-test-support/ilios-common/helpers/quill-editor.js
michaelchadwick: That comment needs to go. It was leftover from Froala, which used that method for some reason.

packages/ilios-common/addon/components/html-editor.gjs
michaelchadwick: The reason for this code is because Froala also retains multiple spaces, but Quill needs this extra bit to achieve parity.

packages/ilios-common/addon-test-support/ilios-common/helpers/quill-editor.js
michaelchadwick: There is a call to get the contents: https://quilljs.com/docs/api#getcontents, but it doesn't work as well as what I used.

There are three ways to get at what's in the editor (I had just added the exclamation point at the end):
Screenshot 2025-08-11 at 1 13 41 PM

packages/ilios-common/addon/components/html-editor.gjs
michaelchadwick: I removed it and nothing bad happened, so perhaps we do not.

packages/ilios-common/addon/utils/load-quill-editor.js
michaelchadwick: I tried Fontawesome but the icons don't match the rest of the toolbar buttons.

The `<svg>` markup is not random, as it comes from Quill itself: `node_modules/quill/assets/icons/[undo|redo].svg`. Why this isn't wired up already, I do not know (nor does anyone else using Quill that I could find). Perhaps an upstream PR is needed?

I put a comment in the code about this, and reproduced the link I followed here so you know where this method came from: https://github.com/slab/quill/issues/885#issuecomment-609384079

michaelchadwick:
@jrjohnson Editor instance now fills container, and link popup closes on escape key. Froala doesn't currently have an 'X'-type close button, so I didn't think to add it to Quill. However, it could be added.

michaelchadwick:
@jrjohnson The best thing I came up with so far for loading the redo/undo svg files is here https://github.com/ilios/frontend/pull/8739/commits/1f8059651051d84dcd79041de398482d1ab1c225

jrjohnson:
I'm good with the `fetch` process. Works for me. Waiting for tests to pass, but I'm all in favor. Think it's worth a second review from @stopfstedt as well.

stopfstedt (approved):

dartajax (commented):
A couple of thing regarding inserting links ... I will put them in separate comments though. Here's one thing I noticed.

image

michaelchadwick:
@dartajax I added in some logic to hopefully turn protocol-less links like that into protocol-rich links.

michaelchadwick:
Look into adding the link target toggle functionality.

michaelchadwick:
@dartajax I added the functionality to make a link open in a new tab or not, so that's taken care of. Please give it a whirl to make sure it works. Still need to add tests, though.

dartajax (commented):
I will leave this one as a comment rather than a change request although it may end up being that as well.

There may be a few of these also as I have noticed some issues.

This is in Froala ...

image

After clicking the edit link (as shown above), you get the following ...

image

With Quill what happens? ...

image

Click edit as shown above ...

The only thing that can be edited is the link itself and not whether it opens in a new tab or what the text display value is. Froala allows for editing of this information.

But it gets worse or weirder. When you click "Save" ...

image

After clickiing "Save" as shown above ...

Return to the link and see this ...

image

It is true if you cancel the operation and don't save any changes, the original link will still be there but there may be other text changes that needed to be made and that should not cause the link to be erased.

dartajax (commented):
not a change request but yes a review still having to do with that pesky link business - the URL stays better than it did before but if you enter without the `https://` business you get a URL like this for navigation purposes ...

https://deploy-preview-8739--ilios-frontend.netlify.app/courses/2318/sessions/www.nytimes.com

... which ends you up with a Zoinks since it can't be found - is there any sort of validation that can be done on the URL itself? And, there is no way to set a link to open in a new tab or not once it has been saved. I'd be find documenting that as a reality in the user guide - set it once and if incorrect, remove and reset / replace the link if need be.

michaelchadwick:
@dartajax I will look into linking again. Going to change how the URL field works (base it on our Virtual Learning Link model and use URL validation and a default value of `https://`)

stopfstedt (changes requested):
unordered lists input is saved as ordered list. 

the markup generated when inserting ordered and un-ordered lists is unnecessarily verbose.

dartajax:
if possible, correct behavior would be (as Froala does) one "undo" click undoes the typing recently entered and then sets the "undo" button to inactive and "redo" button to active. 

Quill currently - need to click "undo" twice to erase one word (not the end of the world bur not optimal either) and the "redo" button becomes enabled (correct) but the "undo" button remains active (not correct) - one more click and bam the text is completely gone but can be reinserted using the "redo" button so at least there's that.

michaelchadwick:
@dartajax Froala and Quill both use a similar _delay_ amount that it waits for prior to registering a change that can be undone/redone.

Froala [typingTimer](https://froala.com/wysiwyg-editor/docs/options/#typingtimer)
Quill [History::delay](https://quilljs.com/docs/modules/history#delay)

Froala and Quill probably use different values, so I'll have to tinker with it.

dartajax:
> @dartajax Froala and Quill both use a similar _delay_ amount that it waits for prior to registering a change that can be undone/redone.
> 
> Froala [typingTimer](https://froala.com/wysiwyg-editor/docs/options/#typingtimer) Quill [History::delay](https://quilljs.com/docs/modules/history#delay)
> 
> Froala and Quill probably use different values, so I'll have to tinker with it.

Fun, eh? Well thanks for checking it out.

michaelchadwick:
The `delay` doesn't seem to work the way I'd think.

**Expectation**: save an undo status after user stops typing AND after `delay` ms
**Reality**: save an undo status after `delay` ms, whether I stop typing or not

Hmmmm.

jrjohnson:
This may be unrelated, but I see siteimprove is picking up [Role with implied hidden content has keyboard focus](https://my2.siteimprove.com/Inspector/1231346/A11Y/Page?pageId=339648844707&impmd=D9BE5081090F02D9F675A75A8B9C5AF2&issue=cantTell+failed&wcag=twopointone&decision=dismissed+falsePositive+none#/sia-r90/failed/eFQ1TkVFUGg=;ZjA3MjljMTk=) on a demo quill edit link when it's within our expanding text widget. I'm wondering if something about the HTML output is slightly different and now triggering an error where we haven't had one before.

dartajax (approved):

PR #8741, opened 2025-08-05
Pin and Separate Days in Week at a Glance

jrjohnson:
It can be difficult in a busy week to see the split between days, and in a busy day as you scroll to remember where you are on the page. Adding a border to split the days makes it easier and as you scroll down the page the day of the week pins itself to the top. This allows us to remove the day from each even making them easier to read as well.

### DIFF
--- a/packages/frontend/tests/acceptance/dashboard/week-test.js
+++ b/packages/frontend/tests/acceptance/dashboard/week-test.js
@@ -164,7 +164,6 @@ module('Acceptance | Dashboard Week at a Glance', function (hooks) {
 
     await page.visit();
     const options = {
-      weekday: 'long',
       hour: 'numeric',
       minute: 'numeric',
     };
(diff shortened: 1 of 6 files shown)


dartajax:
image

jrjohnson:
Cleaned that up @dartajax and fixed the overrun when the screen is smaller by replacing the full day name with shorter versions. Opening it up for code review, but keep any feedback coming!

dartajax:
I would be glad to go with it the way it is but is it possible to line these ups all the way vertically?

dartajax:
> Sure @dartajax, but I'm not sure what way to line it up. Can you mess with a screenshot and a line and let me know what should line up with what? Top of Day top of event? Or maybe just remove the space over the first event so it's closer to the orange line?

That would work - I'd suggest removing the space over the first event so it ...

a. goes closer to the orange line and 
b. matches the top of the day label too

or c. settle on a compromise version of vertical spacing they can share 

my screen shot is not the hottest but here you go ...

image

PR #8747, opened 2025-08-07
Adjust Size of Year in All Weeks View

jrjohnson:
The year header was a big too large. I also found, and removed, a style sheet that was targeting the same component, but was in frontend where it doesn't belong.


dartajax (approved):
seems fine to me

PR #8756, opened 2025-08-08
CurrInv Report list removeConfirmation row now stretches across table on all resolutions

michaelchadwick:
Fixes ilios/ilios#6409

This fixes both Curriculum Inventory Report lists and Sequence Block lists.


stopfstedt (approved):
nice one. LGTM.

PR #8763, opened 2025-08-11
Enable WCAG 2.2 A and AA Tests

jrjohnson:
This is higher than our current target conformance level, but it's a starting place to target any changes we think are worth it before we make this our official supported standard.


jrjohnson:
closing this experiment. We'll wait for axe to make these defaults. It's too weird and annoying to enable them now.

PR #8764, opened 2025-08-13
Create TermsCalendarFilter component like CoursesCalendarFilter component

michaelchadwick:
Fixes ilios/ilios#6414


jrjohnson (approved):
This is an excellent change and I'm all for it as is. Once possible suggestion would be to put the `<SelectedVocabulary>` component in the same file as the new `< TermsCalendarFilter>`. I had trouble figuring out why there are two components and if one was reusable somewhere else.

PR #8768, opened 2025-08-14
focus button first and then click

michaelchadwick:
When running frontend tests in a browser, there is one test that always fails if the browser is not currently focused: `Integration | Component | school list | create new school`. If you focused on the browser and then re-ran it, it worked fine. This PR fixes that by focusing on the button, then clicking it, which fixes the issue.


stopfstedt (approved):
awesome. thanks!

stopfstedt (approved):

PR #8769, opened 2025-08-15
Convert Route Templates to Template Tag Format

jrjohnson:
Two steps to maintain git history, but I converted the remaining route templates to `.gjs`. I hope it reflects that way in the GitHub UI and history, I can't think of any other way to make this work.


stopfstedt (approved):
click-tested all routes and spot-checked code.
LGTM.

PR #8773, opened 2025-08-18
Improve Readability of Single Events

jrjohnson:
The spacing and mix of bold and italic type made the critical information here difficult to read. I've removed most of that and used the icons to call out the attributes. I moved the location into the same visual space as the time as I think that makes it easier to understand what it is.

I also got rid of the triple curlies and used a getter to process the session description, this does the same thing, but doesn't require us to ignore a linting error.

Old:
Screen Shot 2025-08-18 at 10 28 59

New:
Screen Shot 2025-08-18 at 10 28 45


packages/ilios-common/translations/es.yaml
```diff
@@ -347,10 +346,10 @@ general:
   showOnlyMyInstructorActivities: Mostrar solo las actividades de mi instructor
   showOnlyMyLearnerActivities: Mostrar solo mis actividades de aprendizaje
   sortObjectives: Clasificar los objetivos
-  specialAttireIs_Required_: "Se '<strong><em>'require'</em></strong>' vestimenta especial"
+  specialAttireIsRequired: "Se require vestimenta especial"
```
stopfstedt, line 349: :+1:

stopfstedt (approved):
LGTM

PR #8777, opened 2025-08-18
Lighter Bold

jrjohnson:
I've locked in a value for our boldest font at 600. Previously we had many at 600, some defaults at 700, and some places with a mix. Hoping this will help with https://github.com/ilios/frontend/pull/8773#pullrequestreview-3129522233 needs review and discussion.


jrjohnson:
Team Discussion - overall, good. However, in areas where user's in put data (such as session description) we should hold on to the darker bold. It is simply too light to differentiate in this content (such as in a single event view).

stopfstedt (approved):

PR #8783, opened 2025-08-28
Add user id to admin user search

michaelchadwick:
Fixes ilios/ilios#4693

Added `[id]` to search results for easier reference.

### DIFF
--- a/packages/frontend/app/components/manage-users-summary.gjs
+++ b/packages/frontend/app/components/manage-users-summary.gjs
@@ -330,7 +330,7 @@ export default class ManageUsersSummaryComponent extends Component {
                     data-userid={{result.user.id}}
                     {{on "click" (perform this.clickUser result.user)}}
                   >
-                    <span class="name">
+                    <span class="name-and-id">
                       {{result.user.fullName}}
                       [{{result.user.id}}]
                       <UserStatus @user={{result.user}} />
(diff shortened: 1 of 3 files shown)


dartajax (approved):
works for me - I'll get the tests going - merge tomorrow maybe

dartajax:
I am not going to merge this but will keep it open for discussion - assigned to me. This ticket was created with a purpose at hand but this is not the correct solution. I apologize for that. The need does arise for admin users to obtain more information about users during a user search. This occurs when multiple users have the same email address, which has happened. Maybe we need to provide a callout, mouse-over, or other report for these users but the inclusion of the Ilios ID on the same line in the interface has been decided not to be the solution. Keep this open please.

michaelchadwick:
@dartajax I think we decided as a team to go through with this one but only if it gets added as a tooltip instead of inline with the name, so I'll update this to that effect.

dartajax:
I slapped a `DO NOT MERGE` on here - wanted @saschaben to check it out first.

saschaben:
This is an incorrect solution: we are pulling the Ilios ID as the tooltip, when we should be pulling the Campus ID (the "02" number). Swap out the data attribute getting called and I think we are back on track.

dartajax:
see my recent comment - gonna close

PR #8797, opened 2025-09-08
removes broken sonarqube integration.

stopfstedt:
doesn't work, and - according to some - it never did. let's get rid of this.


stopfstedt:
as discussed today -  this is not gonna happen, i'm yanking this PR.

PR #8805, opened 2025-09-16
Improve Session Copy Data Loading

jrjohnson:
Not only were we getting some a11y issues from an empty year list in some places, but we were re-loading a lot of data that is already in the store when navigating here from the courses page.

I've optimized the loading of academic years and applied our existing data loaders to getting all the courses in a school loaded. This speeds up loading a lot on this page and improves the a11y as well.

Fixes ilios/ilios#6505


stopfstedt (approved):
LGTM

PR #8809, opened 2025-09-17
only show filters on dashboard school-calendar

stopfstedt:
fixes ilios/ilios#4299

### DIFF
--- a/packages/ilios-common/addon/controllers/dashboard/calendar.js
+++ b/packages/ilios-common/addon/controllers/dashboard/calendar.js
@@ -80,6 +80,9 @@ export default class DashboardCalendarController extends Controller {
       this.school = null;
     } else {
       this.mySchedule = true;
+      this.showFilters = false;
+      this.courseFilters = true;
+      this.clearFilters();
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/ilios-common/addon/controllers/dashboard/calendar.js
michaelchadwick, line 1: I understand this change, so that any filters that get set while in "All Events" mode will be cleared when switching to "My Schedule" mode.

However, the site does not currently do that, and is there a use case where one would toggle back and forth between the two modes, and not want to lose the filter information while doing so?

dartajax: Not that this is great news or makes anything easier, but I think that use case may indeed exist. You would want to be able to toggle and not have to re-apply the filters upon return - my $.02.

stopfstedt: let's not track values in the URL - like selected filters - for functionality that doesn't apply directly to what's on the screen here.

@dartajax back-button still works if you need to get back to a previous calendar state.

dartajax: I could see as a user appreciating it per @michaelchadwick's question but I can support it with the back button use instead.

michaelchadwick (approved):

PR #8811, opened 2025-09-18
Subject Reports->Competencies by Session now actually runs

michaelchadwick:
Fixes ilios/ilios#4539

This report, when doing a GraphQL query of `competencies` never returned any results (even after giving the web server 30 minutes to finish), so something was amiss. Turns out the number of joins this has to do in the backend is too complex and never actually worked. Thus, I took the session `id` to find the course `id` to find its `courseObjectives`, which have `programYearObjectives`, which have `competencies`. Filter by unique values, and you get results!


stopfstedt (approved):
LGTM

dartajax (approved):
click tested and screen review - seems fine to me

PR #8821, opened 2025-09-22
hide rollover button if given course is locked.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6510

### DIFF
--- a/packages/frontend/tests/acceptance/course/overview-test.js
+++ b/packages/frontend/tests/acceptance/course/overview-test.js
@@ -359,6 +359,18 @@ module('Acceptance | Course - Overview', function (hooks) {
     assert.notOk(page.details.overview.rollover.isVisible);
   });
 
+  test('rollover hidden from privileged users if course is locked', async function (assert) {
+    this.user.update({ administeredSchools: [this.school] });
+    const course = this.server.create('course', {
+      year: 2013,
+      school: this.school,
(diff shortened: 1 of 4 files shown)


michaelchadwick (changes requested):
You can still rollover a course by visiting the `/course/[course-id]/rollover` route directly.

I was also able to rollover a locked course by clicking the rollover icon before it's locked, going to a different tab to lock the course, coming back to the original tab (and even refreshing it), and then clicking "Done" like usual.

Probably need to put some permission check on https://github.com/ilios/frontend/blob/master/packages/ilios-common/addon/components/course/rollover.gjs itself.

stopfstedt:
> You can still rollover a course by visiting the `/course/[course-id]/rollover` route directly.
> 
> I was also able to rollover a locked course by clicking the rollover icon before it's locked, going to a different tab to lock the course, coming back to the original tab (and even refreshing it), and then clicking "Done" like usual.
> 
> Probably need to put some permission check on https://github.com/ilios/frontend/blob/master/packages/ilios-common/addon/components/course/rollover.gjs itself.

valid points but also out of scope. url tampering and multi-tab shenanigans _void warranty_.

the real issue seems to be that the backend allows for rollover of locked courses.

dartajax (approved):
works for me

PR #8833, opened 2025-09-23
Subject Reports->Instructors by Learning Material now actually runs

michaelchadwick:
Fixes ilios/ilios#4535

This report fix, like #8811, is an exception made for Instructor lookup when filtering through a Learning Material. Now should run right as rain.


jrjohnson (changes requested):
`owningUser` isn't what the original report does. Take a look at the query in https://github.com/ilios/ilios/blob/master/src/Repository/UserRepository.php#L829-L852
What you'll need to do is go from LM to courseMaterial/sessionMaterial -> session -> session and get all the instructors attached to offering and ILM by individual or group.

michaelchadwick:
@jrjohnson Ahhh, thank you for clarifying what the original report does. That seems like a more useful report.

jrjohnson (approved):
Nice!

PR #8834, opened 2025-09-23
adding API version to payload prevents API-mismatch notice from showing.

stopfstedt:
this was causing intermittent visual regression test failures.

### DIFF
--- a/packages/frontend/tests/acceptance/header-test.js
+++ b/packages/frontend/tests/acceptance/header-test.js
@@ -17,10 +17,12 @@ module('Acceptance | header', function (hooks) {
 
   test('privileged users can view search', async function (assert) {
     assert.expect(1);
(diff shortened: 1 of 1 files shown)


packages/frontend/tests/acceptance/header-test.js
```diff
@@ -46,11 +50,14 @@ module('Acceptance | header', function (hooks) {
...
 
   test('students can not view search', async function (assert) {
+    const { apiVersion } = this.owner.resolveRegistration('config:environment');
     this.server.get('application/config', function () {
       return {
         config: {
           searchEnabled: true,
+          apiVersion,
         },
+        apiVersion,
```
michaelchadwick, line 60: Why this second instance of `apiVersion` outside of the `config` key?

stopfstedt: thanks for catching that. i removed this line.

PR #8835, opened 2025-09-24
Fix remaining slow subject reports

michaelchadwick:
The following slow/DNF reports will be handled:

* All Courses by MeSH Term
* All Competencies by Session Type
* All Instructors by Session
* All Instructors by Session Type
* All MeSH Terms by Learning Material
* All MeSH Terms by Session
* All Terms by Competency

These reports finish after about 30s, so could be improved, but not sure how:
* All Learning Materials
* All Sessions


packages/frontend/app/components/reports/subject/competency.gjs
```diff
@@ -49,6 +49,41 @@ export default class ReportsSubjectCompetencyComponent extends Component {
...
   }
 
+  async getResultsForSessionType(sessionTypeId, filters) {
+    filters.push(`id: ${sessionTypeId}`);
+
+    const results = await this.graphql.find(
+      'sessionTypes',
+      filters,
+      'school { competencies { id, title, school { title } } }',
+    );
```
stopfstedt, line 59: this gives you all the competencies defined the school that the given session type is defined in.
which yields the same list of all competencies for each session type defined in the same given school.

this is not the data we're after. you'd have to climb through session-type > session > session objective > course-objective > program-year objective > competency instead.

i'm not sure if this is even viable.
but this implementation isn't yielding the correct results.

stopfstedt (changes requested):
all but "All Competencies by Session Type" work as intended.

suggesting you remove the one non-working report, and then we can merge this.

stopfstedt (approved):
LGTM.

PR #8840, opened 2025-09-24
Subject Reports for all schools now show school name in session type object dropdown

michaelchadwick:
Fixes ilios/ilios#6527

Screenshot 2025-09-24 at 4 51 39 PM

### DIFF
--- a/packages/frontend/app/components/reports/subject/new/session-type.gjs
+++ b/packages/frontend/app/components/reports/subject/new/session-type.gjs
@@ -31,6 +31,16 @@ export default class ReportsSubjectNewSessionTypeComponent extends Component {
   }
 
   get sortedSessionTypes() {
+    if (!this.args.school) {
+      return sortBy(
+        this.filteredSessionTypes.map((s) => ({
+          ...s,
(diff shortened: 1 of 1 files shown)


packages/frontend/app/components/reports/subject/new/session-type.gjs
```diff
@@ -31,6 +32,16 @@ export default class ReportsSubjectNewSessionTypeComponent extends Component {
   }
 
   get sortedSessionTypes() {
+    if (!this.args.school) {
+      return sortBy(
+        this.filteredSessionTypes.map((st) => ({
+          id: st.id,
+          title: `${st.school.get('title')}: ${st.title}${!st.active ? ` (${this.intl.t('general.inactive')})` : ''}`,
```
stopfstedt, line 39: `st.school` is a async relationship. i don't see a guarantee that is resolved when accessing the school title here. 
suggesting you wrap this in a async function + TrackedAsyncData in a companion `sortedSessionTypesData` getter and resolve it there.

needs `const school = await st.school;` somewhere in there.

stopfstedt (approved):
LGTM

PR #8858, opened 2025-10-01
Refactor EditableField and FadeText

jrjohnson:
The purpose was to fix ilios/ilios#6532, but I needed to take the entire thing apart so we could remove links only when needed. Took advantage of contextual component, some things we've learned about modifiers, and a new css feature `mask-image` to remove a lot of duplication.

Best reviewed as individual commits.

### DIFF
--- a/packages/test-app/tests/integration/components/fade-text-test.gjs
+++ b/packages/test-app/tests/integration/components/fade-text-test.gjs
@@ -1,6 +1,6 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'test-app/tests/helpers';
-import { render } from '@ember/test-helpers';
+import { render, waitFor } from '@ember/test-helpers';
 import { component } from 'ilios-common/page-objects/components/fade-text';
 import FadeText from 'ilios-common/components/fade-text';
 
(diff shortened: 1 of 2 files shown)


stopfstedt:
i was elated about this refactoring, until i got to the link replacement part of it. i believe there's a cleaner and simpler solution that builds on most of this.

please see https://github.com/jrjohnson/frontend/pull/896

stopfstedt (approved):
LGTM

stopfstedt (changes requested):
needs a minor fix in the test coverage, otherwise good.

packages/test-app/tests/integration/components/fade-text-test.gjs
```diff
@@ -1,10 +1,8 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'test-app/tests/helpers';
-import { render, waitFor } from '@ember/test-helpers';
+import { render } from '@ember/test-helpers';
```
stopfstedt: waitFor is still needed.


```suggestion
import { render, waitFor } from '@ember/test-helpers';
```

PR #8859, opened 2025-10-01
vertically align buttons at the bottom.

stopfstedt:
this aims to align buttons containing icons with non-button icons in the same "row" - such as the actions in the courses list.

refs https://github.com/ilios/ilios/issues/6512


dartajax (approved):
this is nice

PR #8861, opened 2025-10-06
Add a report results text filter

michaelchadwick:
Fixes ilios/ilios#6003

Example: Courses
Screenshot 2025-10-06 at 1 52 28 PM


dartajax:
Interesting - wondering if a download is performed, I am assuming all data gets downloaded. Also the message indicated below would not be accurate if applied to the filtered output.

image

dartajax:
image

michaelchadwick:
@dartajax I've only done this for subject reports that result in Courses thus far, as a PoC. Also, I'm not sure what to do for reports that have more than 200 results, but are also filtered using this new filter.

PR #8869, opened 2025-10-07
Reduce top/bottom padding from LG members

michaelchadwick:
Fixes ilios/ilios#4156

Top Before/After:
members-top-before
members-top-after

Bottom Before/After:
members-bot-before
members-bot-after


dartajax (approved):
looks good to me

PR #8893, opened 2025-10-17
adds aria-labels that handle duplicate session titles in the same course.

stopfstedt:
refs https://github.com/ilios/ilios/issues/6567


michaelchadwick (changes requested):
Sorting sessions by status fails and results in an error

Image

jrjohnson:
I'd like to talk through this implementation a bit.

PR #8940, opened 2025-11-07
display end-of-day, ILM-based events in the calendar

stopfstedt:
fixes ilios/ilios#6631

couple of notes:

- `calendarStartDate` and `calendarEndDate` are derived values and should really be getters. as a follow-up, make event objects based on an actual class, rather than using a plain JS object. then, we can make these attrs actual getters.
- in our test coverage, it is currently  not possible to retrieve the start/end date and -time of events on the calendar via cli page objects. i think this can be rectified by adding `data-test-` attrs to the event notes in the calendar grid that would carry this information, which would allow us to the read them back out during testing. another follow-up item.

### DIFF
--- a/packages/ilios-common/addon/classes/events-base.js
+++ b/packages/ilios-common/addon/classes/events-base.js
@@ -115,6 +115,27 @@ export default class EventsBase extends Service {
     obj.postrequisites = sortBy(obj.postrequisites, 'startDate');
     obj.postrequisites = sortBy(obj.postrequisites, 'name');
     obj.isUserEvent = isUserEvent;
+
+    // The start and end date of the event, for display purposes on the calendar. See comment block below.
+    obj.calendarStartDate = obj.startDate;
+    obj.calendarEndDate = obj.endDate;
+
+    // ACHTUNG!
+    // ILMs don't really have a duration, they have due-date.
+    // But in order to display them in a calendar, we have to make up a duration for them, otherwise they won't show up.
+    // So we're filling in a start-date that's the equivalent of the actual due-date,
+    // and give it an end-date that's fifteen minutes out.
+    // However, if the given start-date is 11:45p or later in the day, that event would be considered to continue
+    // into the next day, effectively making it a "multi-day" event.
+    // Multi-days are currently not shown in the calendar, instead they are displayed in a table below the calendar.
+    // To prevent this from happening, we pin the calendar display start-date to 11:45p and the end-date to 11:59p.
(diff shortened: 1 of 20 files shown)


packages/ilios-common/addon/classes/events-base.js
```diff
@@ -115,6 +115,27 @@ export default class EventsBase extends Service {
...
+    // ILMs don't really have a duration, they have due-date.
+    // But in order to display them in a calendar, we have to make up a duration for them, otherwise they won't show up.
+    // So we're filling in a start-date that's the equivalent of the actual due-date,
+    // and give it an end-date that's fifteen minutes out.
+    // However, if the given start-date is 11:45p or later in the day, that event would be considered to continue
+    // into the next day, effectively making it a "multi-day" event.
+    // Multi-days are currently not shown in the calendar, instead they are displayed in a table below the calendar.
+    // To prevent this from happening, we pin the calendar display start-date to 11:45p and the end-date to 11:59p.
+    // [ST 2025/11/07]
+    const startDate = DateTime.fromISO(obj.startDate);
```
jrjohnson: Readability tweak: could this be inside the `if` block and then, since it is ilm only in there, be `dueDate`. Because it's confusing on the next couple of lines to be doing `startDate.set`, even though luxon treats `set` as returning a new instance of the value, it's hard to read it that way. Looks like you're just setting a value on `startDate` twice.

packages/ilios-common/addon/classes/events-base.js
```diff
@@ -115,6 +115,27 @@ export default class EventsBase extends Service {
     obj.postrequisites = sortBy(obj.postrequisites, 'startDate');
     obj.postrequisites = sortBy(obj.postrequisites, 'name');
     obj.isUserEvent = isUserEvent;
+
+    // The start and end date of the event, for display purposes on the calendar. See comment block below.
+    obj.calendarStartDate = obj.startDate;
```
jrjohnson: I like this rename, in the followup to this when we build events as classes we should also drop outside access to startDate and endDate if possible.

jrjohnson (approved):
I'm good with this as an iterative change and it solves the problem nicely. I'm in full agreement that another refactor is needed for the `Event`. There are too many places where we're faking the shape of the object in both app and test code. We're also having to create luxon objects with the same data many times in many places and sometimes reading the original data as well for sorting and display. All this could be eased significantly with a base class that was more complete.

dartajax (commented):
works for me

PR #8946, opened 2025-11-12
New Curriculum Report: Session Offerings

michaelchadwick:
Fixes ilios/ilios#6225

This adds another option to the Curriculum Reports choices: Session Offerings. It is very similar to the Session Objectives report, but swaps out objectives for offerings, and adds Learner Groups (count to summary, and names to download).

### DIFF
--- a/packages/frontend/app/components/reports/curriculum/session-offerings.gjs
+++ b/packages/frontend/app/components/reports/curriculum/session-offerings.gjs
@@ -190,7 +190,7 @@ export default class ReportsCurriculumSessionOfferingsComponent extends Componen
       rhett[this.intl.t('general.course')] = o.courseTitle;
       rhett[this.intl.t('general.year')] = o.courseYear;
       rhett[this.intl.t('general.session')] = o.sessionTitle;
-      rhett[this.intl.t('general.date')] = o.offeringDate;
+      rhett[this.intl.t('general.offeringDate')] = o.offeringDate;
       if (o.learnerGroups) {
         rhett[this.intl.t('general.learnerGroups')] = o.learnerGroups.join(', ');
       }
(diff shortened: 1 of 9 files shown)


packages/frontend/translations/en-us.yaml
```diff
@@ -397,9 +398,13 @@ general:
   sequenceBlockTitlePlaceholder: Enter a title for this sequence block
   sequenceNumber: "Sequence #"
   sessionAttributes: Session Attributes
+  sessionLink: Session URL
```
jrjohnson: Should either be `Web Link` like the other reports or those reports should get updated to use this translation. I'd vote for `Web Link` personally.

jrjohnson (commented):
Blazing fast and seems pretty feature complete with what I remember of the requirements. Couple of tiny changes in the code needed.

packages/frontend/translations/en-us.yaml
michaelchadwick: I changed it to `general.link` to match the others.

packages/frontend/app/components/reports/curriculum/session-offerings.gjs
```diff
@@ -127,7 +127,11 @@ export default class ReportsCurriculumSessionOfferingsComponent extends Componen
       return a.courseTitle.localeCompare(b.courseTitle);
     }
 
-    return a.sessionTitle.localeCompare(b.sessionTitle);
+    if (a.sessionTitle !== b.sessionTitle) {
+      return a.sessionTitle.localeCompare(b.sessionTitle);
+    }
+
+    return a.offeringDate.localeCompare(b.offeringDate);
```
jrjohnson: This is comparing the two strings `2025-11-18, 8:10 AM`. That's probably mostly going to work. I'm not sure. I think it's wroth converting them into `Date` or `DateTime` and doing a date comparison as well. Might be better to save a timestamp in the `this.results` and discard it for display. 

Maybe I'm just being pedantic about sort, push back welcome!

michaelchadwick: I think you're right. I'm going to initially pull in the raw `startDate` for `this.results`, but then in the compare function convert it to use `.toUnixInteger()`.

jrjohnson (approved):
nailed it!

dartajax (approved):
I downloaded the data and spot-checked it - from a first pass, it looks good to me.

PR #8956, opened 2025-11-18
apply some sanity to creating events from data.

stopfstedt:
__this is work in progress.__

pushed it up here so we can talk about it some more.

i'll pull this apart into meaningful changesets once we have a solid solution to this problem.

### DIFF
--- a/packages/frontend/app/components/learner-group/calendar.gjs
+++ b/packages/frontend/app/components/learner-group/calendar.gjs
@@ -18,6 +18,7 @@ import LoadingSpinner from 'ilios-common/components/loading-spinner';
 
 export default class LearnerGroupCalendarComponent extends Component {
   @service localeDays;
+  @service schoolEvents;
   @tracked selectedDate = DateTime.now();
   @tracked showSubgroupEvents = false;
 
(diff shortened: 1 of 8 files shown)


packages/ilios-common/addon/classes/events-base.js
```diff
@@ -97,87 +152,7 @@ export default class EventsBase extends Service {
...
-    slug += schoolId;
-    slug += DateTime.fromISO(event.startDate).toFormat('yyyyMMdd');
-    if (event.offering) {
-      slug += 'O' + event.offering;
-    }
-    if (event.ilmSession) {
-      slug += 'I' + event.ilmSession;
-    }
-    return slug;
+    return new Proxy(obj, this.handler);
```
stopfstedt: using a Proxy here makes the most sense to me. i started by hand-rolling my own proxy wrapper, but realized half-way through that there's a ton of properties and that i didn't want to endlessly write getters for all of these.

packages/frontend/app/components/learner-group/calendar.gjs
```diff
@@ -61,22 +62,24 @@ export default class LearnerGroupCalendarComponent extends Component {
...
-        isPublished: session.isPublished && course.isPublished,
-        isBlanked: false,
-      };
+      const school = await course.school;
+      return this.schoolEvents.createEventFromData(
+        {
+          startDate: offering.startDate.toISOString(),
+          endDate: offering.endDate.toISOString(),
+          courseTitle: course.title,
+          school: school.id,
```
stopfstedt: for learner-group and offering calendars, i decided to create these pseudo-events as school events, seems the less-wrong context hack here. adding `school` for consistency, this attr is currently not processed (we don't create event slugs in these calendars).

packages/ilios-common/addon/classes/events-base.js
```diff
@@ -1,8 +1,63 @@
 import Service from '@ember/service';
 import { DateTime } from 'luxon';
 import { mapBy, sortBy, uniqueValues } from 'ilios-common/utils/array-helpers';
+import { getSlug } from 'ilios-common/utils/event-helpers';
 
 export default class EventsBase extends Service {
+  /**
+   * Proxy handler for events.
+   */
+  handler = {
```
stopfstedt: the proxy handler. 

i wonder what this does for performance, as the "intercepted attrs" are recalculated on each read. 

a cached getter would probably be better here, but cannot be implemented.

as another aside - the native Proxy class cannot be sub-classed :clown_face: .

packages/frontend/app/components/learner-group/calendar.gjs
```diff
@@ -61,22 +62,24 @@ export default class LearnerGroupCalendarComponent extends Component {
...
-        location: offering.location,
-        color: '#84c444',
-        prerequisites: [],
-        postrequisites: [],
-        isScheduled: session.isScheduled || course.isScheduled,
-        isPublished: session.isPublished && course.isPublished,
-        isBlanked: false,
-      };
+      const school = await course.school;
+      return this.schoolEvents.createEventFromData(
```
stopfstedt: using the same `createEventFromData()` method across the board for creating event object from data is probably the most immediately beneficial change in this PR.
if anything here should land then it's this (alongside the same change in the offering calendar component).

stopfstedt:
withdrawn.

PR #8959, opened 2025-11-19
Use factory to create events on offering and learner-group calendars

stopfstedt:
extracted from #8956 

this standardizes how events are created for calendars across the entire application. 

offerings for the learner-group calendar and offering-calendar are now created as school events. this close a minor context gap - prior to this change these were neither user- nor school-events, the calendar components expect either one of these event types as input.


jrjohnson (approved):

dartajax (approved):
I checked both of these calendars and found an issue that has absolutely nothing to do with this PR. I approve of this - no related issues found.

PR #8960, opened 2025-11-19
remove pointless data transformation.

stopfstedt:
this doesn't do anything meaningful and gets in the way of using a more robust data structure, like a class with getters or a Proxy object. begone!

extracted from #8956 

**side-note**

further down, we're modifying the events by setting a `isMulti` flag if applicable. i believe that the removed code is related to that, although i can't back that claim up based on commit messages.

ideally, we'd not be modifying these events at all, once they're generated. conceptually, they are "read only".

will explore a solution that gets rid of a `event.isMulti` as a separate issue.

### DIFF
--- a/packages/ilios-common/addon/components/ilios-calendar.gjs
+++ b/packages/ilios-common/addon/components/ilios-calendar.gjs
@@ -36,8 +36,7 @@ export default class IliosCalendarComponent extends Component {
         if (!(hash in hashedEvents)) {
           hashedEvents[hash] = [];
         }
-        //clone our event, so we don't trample on the original when we change location
-        hashedEvents[hash].push(Object.assign({}, event));
+        hashedEvents[hash].push(event);
       });
       const compiledEvents = [];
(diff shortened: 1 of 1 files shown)


stopfstedt:
> My only concern is the `//clone our event, so we don't trample on the original when we change location` seems to point to this doing something very specific. I don't know what though. I'm guessing it's the `isMulti` where we change the location to `tbd`? I don't have any memory for how these fit together.

this has to do with displaying multiple events as one in the weekly and monthly calendar. if it's a multi-event, clicking on it will take you to the calendar's day-view with all similar events expanded. 
non-multi event will let you click through to the event's detail screen directly.

jrjohnson (approved):

dartajax (approved):
click-tested per instructions from @stopfstedt - seems fine - no changes noticed on Dashboard calendar - clicking into multiples from Week or Month view - behaves as before so ... approval granted

PR #8972, opened 2025-11-25
Prereq linkage fix

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6684


dartajax (approved):
works for me - great stuff - activity check boxes are here too - as far as I can tell, all set with this one

jrjohnson (approved):

PR #8984, opened 2025-12-03
New Curriculum Report: Tagged Terms (Listed)

michaelchadwick:
Fixes ilios/ilios#6126

This adds another new Curriculum Report that returns course, sessions, and terms tagged on them both. All terms are listed one per table row per course/session.

A follow-up is to add the potential for "rolled up" rows for term: https://github.com/ilios/frontend/pull/8984.

### DIFF
--- a/packages/frontend/app/components/reports/curriculum/tagged-terms.gjs
+++ b/packages/frontend/app/components/reports/curriculum/tagged-terms.gjs
@@ -25,7 +25,7 @@ export default class ReportsCurriculumTaggedTermsComponent extends Component {
   @cached
   get queryPromises() {
     const chunks = chunk(this.args.courses, 5);
-    const sessionData = ['id', 'title', 'terms { id, title }'].join(', ');
+    const sessionData = ['id', 'title', 'terms { id, title, vocabulary { title } }'].join(', ');
 
     const data = [
       'id',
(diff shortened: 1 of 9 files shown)


packages/frontend/app/components/reports/curriculum/tagged-terms.gjs
```diff
@@ -0,0 +1,252 @@
...
+      return new TrackedAsyncData(this.graphql.find('courses', filters, data.join(', ')));
+    });
+  }
+
+  get completedPromises() {
+    return this.queryPromises.filter((tad) => tad.isResolved);
+  }
+
+  get reportRunning() {
+    return this.queryPromises.length > this.completedPromises.length;
```
stopfstedt: i think this is more to the point than comparing array lengths.

```suggestion
    return !this.queryPromises.every((tad) => tad.isResolved);
```

stopfstedt (changes requested):
i have one issue, the rest looks good.

stopfstedt (approved):
LGTM

dartajax (changes requested):
image

dartajax:
More like this actually - the Vocabulary is still applicable and the URL leaves off the session. This is another reason we need one per row - could be from a different vocabulary at course level too.

image

michaelchadwick:
@dartajax Thanks for the feedback! Back to the workshop...

michaelchadwick:
@dartajax OK, the report should now look like your latest screenshots.

dartajax (approved):
looks good to me

stopfstedt (approved):
re-reviewed. LGTM.

PR #8995, opened 2025-12-04
replaces graph-with-data-table mixin with class equivalent.

stopfstedt:
fixes ilios/ilios#6668
refs ilios/ilios#6608


michaelchadwick (approved):
👍

dartajax (approved):
I clicked around quite a bit - didn't notice anything awry - focused on Visualizations around the application - seems fine.

PR #8999, opened 2025-12-08
replaces SCSS main-list-box-header-title mixin with class equivalent

stopfstedt:
fixes ilios/ilios#6667
refs https://github.com/ilios/ilios/issues/6608


dartajax (approved):
seems fine to me - click tested areas mentioned and more - no visual diffs here or when using our new "percy"

michaelchadwick (approved):
This all looks good. Moving that one pre-existing mixin to the new class works for now (it already had others). We can handle removing it later.

PR #9031, opened 2026-01-07
New ProgramYear selection now disables unavailable years instead of removing them

michaelchadwick:
Fixes ilios/ilios#6730

All possible years are now listed in dropdown, but only non-existing ones are selectable, and the first selectable year is now the default.


stopfstedt (approved):
LGTM

dartajax (approved):

PR #9037, opened 2026-01-08
convert calendarEventTooltip utility fn into service.

stopfstedt:
this allows us to inject the intl service directly into the tooltip service, rather than passing it as input to the util function. which will help with tracking and linting translations used within.

fixes ilios/ilios#6784


michaelchadwick (approved):
The creation of something awesome: completed

dartajax (approved):
Visual Diff approved - click tested a bit - good to go.

PR #9038, opened 2026-01-09
pass translated strings to flash-messages service instead of lang key.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/6787


michaelchadwick (approved):
Awesome

dartajax (approved):
click tested - checked language strings - seems fine to me

PR #9041, opened 2026-01-09
use translations in locale chooser in lintable manner.

stopfstedt:
this change has the beneficial side-effect that we can now fail more gracefully if the current locale is not on the supported list. don't think that's a realistic scenario, but why not.

refs ilios/ilios#6792


michaelchadwick (approved):
Tubular

dartajax (approved):
makes sense to me

PR #9042, opened 2026-01-09
move translations into right spot, and remove duplicate entry.

stopfstedt:


dartajax (approved):
click tested - seems fine

dartajax (approved):
click tested - seems fine - maybe if the screen had properly refreshed the first time, I would not have posted the same comment ... twice

jrjohnson (approved):

PR #9043, opened 2026-01-10
make lang string for calendar view types lintable.

stopfstedt:
this has to be a getter to be receptive to global language changes.

refs ilios/ilios#6792


michaelchadwick (approved):
Swoot

dartajax (approved):
click tested - seems fine to me

PR #9044, opened 2026-01-10
make usage of day-of-the-week labels in offering form lintable.

stopfstedt:
i was hoping to do away with these translations altogether, but i can't figure out how to conjure a list of localized weekday names out of ember-intl, not Luxon, without creating a week's worth of dates first and then extracting that info from those. which is way to complicated, so lang strings it remains.

refs ilios/ilios#6792


michaelchadwick (approved):
Gnarly

dartajax (approved):
click tested a bit in offering form - seems fine - day of week and all

PR #9045, opened 2026-01-10
Adds LocalStorage service and saves user locale preference

michaelchadwick:
Fixes ilios/ilios#6702

This does two things: it creates a `LocalStorage` service that reads/writes to an `ilios` key in `localStorage`, and it starts saving the user locale preference in a `locale` item.


packages/frontend/app/components/locale-chooser.gjs
```diff
@@ -41,6 +42,7 @@ export default class LocaleChooserComponent extends Component {
   changeLocale(id, event) {
     this.isOpen = false;
     this.intl.setLocale(id);
+    this.ls.set('locale', id);
```
stopfstedt:
```suggestion
    this.localStorage.set('locale', id);
```

stopfstedt (changes requested):
couple of minor syntax changes. otherwise, this looks great.

stopfstedt (approved):
LGTM

dartajax (approved):
very cool

PR #9046, opened 2026-01-10
make file-type specific translatable labels lintable.

stopfstedt:
this comes at the expense of making the code slightly more verbose. getting a handle on linting is worth it though, imho.

refs ilios/ilios#6792


michaelchadwick (approved):
Noice

PR #9076, opened 2026-01-21
scroll deletion confirmation dialog into view.

stopfstedt:
fixes ilios/ilios#6821


michaelchadwick (approved):

dartajax (approved):
Ilios is fine and so is this. Works for me.

PR #9101, opened 2026-01-28
Only get/set localStorage for locale if not testing

michaelchadwick:
Fixes ilios/ilios#6841

Due to the change in [ilios/ilios#9045](https://github.com/ilios/frontend/pull/9045), the locale is now being saved in localStorage. However, this means tests in the browser can fail if the locale has been changed from the default. This PR makes it so the app only interacts with localStorage if it's not testing.


jrjohnson (commented):
This solves the problem, but I think maybe consolidate both the get/set in the localStorage service and returning `null` whenever testing as if nothin was set would be a bit better than spreading `isTesting` into more than one place.

michaelchadwick:
@jrjohnson OK, I moved the logic into the service itself. If it's doing a `get()`, it just returns `null`. If it's doing a `set()`, then it just returns early.

jrjohnson (approved):
🔥

dartajax:
Visual Diff - nice - nothing out of whack but worked fine.

dartajax (approved):

PR #9112, opened 2026-02-03
Program overview label/value elements aligned and shortTitle revert mechanism fixed

michaelchadwick:
Fixes ilios/ilios#6874
Fixes ilios/ilios#6889

While updating the proper alignment of the Program overview label/value elements, I noticed that changing the shortTitle and then canceling did not do what it was supposed to do. So, that is rolled into this fix, as well.


jrjohnson (approved):
🪨 🪨

dartajax (approved):
click tested - approved

PR #9115, opened 2026-02-04
Breadcrumb links hover improvement

michaelchadwick:
Fixes ilios/ilios#6857

breadcrumb-hover

Breadcrumb links now do a color/bgcolor flip when hovered over, and the text does not underline. The last bit, where you currently are, has had its border removed and changed to neutral black color as it is not a link. Some re-organizing of rule order was necessary to pass linting, too.

### DIFF
--- a/packages/ilios-common/app/styles/ilios-common/components/breadcrumbs.scss
+++ b/packages/ilios-common/app/styles/ilios-common/components/breadcrumbs.scss
@@ -1,13 +1,8 @@
 @use "sass:math";
 
 .breadcrumbs {
-  $breadcrumb-border-color: var(--blue);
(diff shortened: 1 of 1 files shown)


jrjohnson (changes requested):
The current location (last crumb) needs something, probably a border, possibly a background.

I think it's also worth breaking up the variables at the top since they all point to css variables now and make it a little difficult to tell what's up. We probably stole the original style and the variables came along for customization.

jrjohnson (approved):

dartajax (approved):
fine with me

dartajax:
Visual Diff did a better job for me at least than Percy on this PR - It probably needed a rebase but I went in and scoped it out anyway. Gonna merge this one.

PR #9122, opened 2026-02-04
Visualization "no data" translation message update

michaelchadwick:
Fixes ilios/ilios#6893

This fixes one visualization route translation message (`SessionTypes`) to be more consistent with the `Objectives` message, and more accurate. It also reorganizes some translation labels to be better grouped, contextually.

### DIFF
--- a/packages/frontend/tests/acceptance/course/visualizations-test.js
+++ b/packages/frontend/tests/acceptance/course/visualizations-test.js
@@ -1,8 +1,9 @@
 import { module, test } from 'qunit';
 import { currentURL } from '@ember/test-helpers';
-import { setupApplicationTest } from 'frontend/tests/helpers';
+import { setupApplicationTest, takeScreenshot } from 'frontend/tests/helpers';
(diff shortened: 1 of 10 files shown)


jrjohnson (approved):

jrjohnson:
Approved, but if you have time this probably needs better visual tests since it came up clean.

michaelchadwick:
@jrjohnson Besides adding an acceptance test screenshot for the "list" route of visualizations, I'm not sure what else would need one.

jrjohnson (approved):

dartajax (approved):
I approve this PR because the messaging is better and more consistent indeed. There are other issues in my opinion - will put them in separate PR's.

dartajax (changes requested):
Can we line then up while at it and add the verb "are" to the messages to make complete sentences such as for the second one "No sessions are currently linked to instructional time."

Image

dartajax (approved):

PR #9123, opened 2026-02-04
Vocabular(ies/y) by Session Type now has proper unique `<title>`

michaelchadwick:
Fixes ilios/ilios#6896

### DIFF
--- a/packages/frontend/app/components/school/session-type-visualize-vocabularies.gjs
+++ b/packages/frontend/app/components/school/session-type-visualize-vocabularies.gjs
@@ -1,8 +1,21 @@
 import { LinkTo } from '@ember/routing';
 import { hash } from '@ember/helper';
 import t from 'ember-intl/helpers/t';
+import { pageTitle } from 'ember-page-title';
 import VisualizeSessionTypeVocabulariesGraph from 'frontend/components/school/visualize-session-type-vocabularies-graph';
 <template>
+  {{pageTitle
+    "Schools | "
+    @model.school.title
+    " | "
+    (t "general.visualizations")
+    " | "
+    (t "general.sessionTypes")
+    " | "
+    @model.title
+    " | "
+    (t "general.vocabularies")
+  }}
   <section
     class="school-session-type-visualize-vocabularies data-visualization"
     data-test-school-session-type-visualize-vocabularies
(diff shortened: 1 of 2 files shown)


packages/frontend/app/components/school/session-type-visualize-vocabularies.gjs
```diff
@@ -1,8 +1,21 @@
 import { LinkTo } from '@ember/routing';
 import { hash } from '@ember/helper';
 import t from 'ember-intl/helpers/t';
+import { pageTitle } from 'ember-page-title';
 import VisualizeSessionTypeVocabulariesGraph from 'frontend/components/school/visualize-session-type-vocabularies-graph';
 <template>
+  {{pageTitle
+    "Schools | "
```
jrjohnson: I'm surprised `Schools` is needed here, usually this would come from a higher level route. Also should be a translation, weird the linter didn't flag that. My preference would be that it's not here and on a higher level instead. If that would require our routes to be reorganized, then it can stay, but we should address that separately.

jrjohnson (changes requested):

packages/frontend/app/components/school/session-type-visualize-vocabularies.gjs
michaelchadwick: Good call on the translation. Fixed.

As for the source of such things, I feel like you had something similar to say about this when I made my first pass at unique `<title>` values (this one got missed because it was on a route that I missed): https://github.com/ilios/frontend/pull/8665

Did not know how to fix it then, nor now, but if it can a followup makes sense.

packages/frontend/app/components/school/session-type-visualize-vocabulary.gjs
```diff
@@ -1,8 +1,23 @@
 import { LinkTo } from '@ember/routing';
 import { hash } from '@ember/helper';
 import t from 'ember-intl/helpers/t';
+import { pageTitle } from 'ember-page-title';
 import VisualizeSessionTypeVocabularyGraph from 'frontend/components/school/visualize-session-type-vocabulary-graph';
 <template>
+  {{pageTitle
+    "Schools | "
```
jrjohnson: Needs translation here as well.

jrjohnson (changes requested):

jrjohnson (approved):

dartajax (approved):
looks great - tested - approved

dartajax:
Reviewed Visual Diff again - seems fine to me - Percy? not sure - probably no changes reported

PR #9126, opened 2026-02-05
Add Cohort Members toggle component to Program Years

michaelchadwick:
Fixes ilios/ilios#6872

This adds a `ProgramYearCohortMembers` component and places it under the `AssociatedCourses` component on a `ProgramYear` detail route. It works much like that component, except it shows members, and they link to their user detail route.

### DIFF
--- a/packages/frontend/app/components/program-year/cohort-members.gjs
+++ b/packages/frontend/app/components/program-year/cohort-members.gjs
@@ -1,0 +1,133 @@
+import Component from '@glimmer/component';
+import { cached, tracked } from '@glimmer/tracking';
+import { fn, uniqueId } from '@ember/helper';
+import { on } from '@ember/modifier';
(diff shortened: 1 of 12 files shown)


jrjohnson (dismissed):
I love this. Super complete implementation too. Could you add some visual tests? The acceptance test should get percy and visual diff, the integration test can get a bunch of visual diff tests.

stopfstedt (approved):

dartajax (approved):
I like this.

dartajax:
I reviewed Visual Diff as well - Percy was more accurate in the sense that Percy caught all of the places where the cohort members would be listed below the Associated Courses area on Program Year detail. I am not sure if Visual Diff highlights changes that occur way down at the bottom of the page - any given page - not sure but either way this looks good - will merge in any case.

PR #9129, opened 2026-02-06
Remove text underline on button hover

michaelchadwick:
Fixes ilios/ilios#6878

I did a few sweeps of the site, and the only two I could find were on the Subject/Curriculum switcher and the Unassigned Students Summary manage buttons.

### DIFF
--- a/packages/frontend/app/styles/components/reports/switcher.scss
+++ b/packages/frontend/app/styles/components/reports/switcher.scss
@@ -4,6 +4,7 @@
   display: flex;
   justify-content: center;
   margin: 1em;
+
   a {
     background-color: var(--white);
     color: var(--blue);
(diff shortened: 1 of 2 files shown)


jrjohnson:
No conflicts, but this needs a rebase (and possibly changes) after #9127 was merged and `Ilios-link-button` was removed.

jrjohnson (changes requested):
Needs a focus state outline now that underline is removed. I think the switcher also needs a hover state background change, but that's possibly more personal preference.

jrjohnson (approved):

dartajax (approved):
while this PR does what it's supposed to do - and that is of course great and also why I'm approving the request - I agree that there needs to be some indication - callout/background color - something to indicate what's going on here - and a pattern that fits with the rest of the application naturally

PR #9131, opened 2026-02-07
unhinkify details cohort component

stopfstedt:
fixes ilios/ilios#6720

now with test coverage and code comments, i hope this helps.

as an aside - i tried to filter out unmodified course objectives by checking their [`hasDirtyAttributes`](https://api.emberjs.com/ember-data/release/classes/Model#hasDirtyAttributes) value but couldn't get it to work. maybe this doesn't work on collections...  anyway - hence the two-step process. if someone  has a more efficient/better solution please let me know.

### DIFF
--- a/packages/frontend/tests/acceptance/course/cohorts-test.js
+++ b/packages/frontend/tests/acceptance/course/cohorts-test.js
@@ -82,49 +82,58 @@ module('Acceptance | Course - Cohorts', function (hooks) {
 
   test('list cohorts', async function (assert) {
     await page.visit({ courseId: this.course.id, details: true });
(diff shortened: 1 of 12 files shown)


jrjohnson (approved):
Hinkey free! LGTM, but can you check on the visual diff results? Looks like our normal font based whitespace issue, but the change I'm seeing the most is in the program year cohorts section, it wasn't changed here, but since it says cohorts just want to validate it's ok.

dartajax (approved):
I clicked tested looking for discrepancies pointed out by VIsual Diff and just testing in general - it seems fine to me. Approved.

PR #9141, opened 2026-02-11
remove on-focus styling from header menu item.

stopfstedt:
applying styles on focus turns out to be too much - the first header list item is always in focus when the menu is expanded, permanently coloring it blue.
that's not what we want.
removing the :focus pseudo-class seems to be the only viable option here.

fixes ilios/ilios#6912

### DIFF
--- a/packages/frontend/app/styles/shared/header-menu.scss
+++ b/packages/frontend/app/styles/shared/header-menu.scss
@@ -55,8 +55,7 @@
         color: var(--white);
       }
 
-      &:hover,
-      &:focus {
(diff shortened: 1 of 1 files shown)


michaelchadwick (approved):
Simple, yet effective. Ap-proved.

jrjohnson (changes requested):
This makes it impossible to keyboard navigate these menus as there is no indication of focus anymore. The first element is immediately in focus when the menu opens for that reason as well. Will probably need something more complicated, or we might have to live with this issue.

PR #9147, opened 2026-02-13
better differentiate active, focus, and hover states in header menus

stopfstedt:
fourth attempt at fixing this. this time around, i'm de-emphasizing the hover state.

a grey, solid border now indicates that the given menu item is in focus. green background still indicates the active menu item. a blue background now highlights the menu item that the user is hovering over, with the exception of the currently active item (which remains green). focused elements in a hover and active state have light-green and light-blue solid borders around.

image

image

image

image

image

let me know what you think, thanks.

fixes ilios/ilios#6912

### DIFF
--- a/packages/frontend/app/components/locale-chooser.gjs
+++ b/packages/frontend/app/components/locale-chooser.gjs
@@ -95,13 +95,7 @@ export default class LocaleChooserComponent extends Component {
         break;
     }
   }
-  @action
-  clearFocus(event) {
-    const buttons = event.target.parentElement.children;
-    for (let i = 0; i < buttons.length; i++) {
-      buttons[i].blur();
-    }
-  }
+
   @action
   toggleMenu(event) {
     const { key } = event;
@@ -151,7 +145,7 @@ export default class LocaleChooserComponent extends Component {
               data-test-item
               {{on "click" (fn this.changeLocale loc.id)}}
               {{on "keydown" this.moveFocus}}
-              {{on "mouseenter" this.clearFocus}}
+              {{on "keyup" this.moveFocus}}
               {{focus (eq index 0)}}
             >
               {{loc.text}}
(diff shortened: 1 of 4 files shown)


dartajax (approved):
I like this but it warrants a team discussion to be certain.

stopfstedt:
> I think we should flip the state. Outline is usually for focus and color for hover and that's the default browser styling we have everywhere right now. As we work on [ilios/ilios#6914](https://github.com/ilios/ilios/issues/6914) we'll probably make the outline explicit on more stuff so it would make sense here as well.

i'm taking it back to the workbench.

stopfstedt:
> I think we should flip the state. Outline is usually for focus and color for hover and that's the default browser styling we have everywhere right now. As we work on [ilios/ilios#6914](https://github.com/ilios/ilios/issues/6914) we'll probably make the outline explicit on more stuff so it would make sense here as well.

reversed the styles - dashed border on focus, color change on hover.

however - this unveiled a discrepancy between how the language picker and the user menu work - the items in the language picker lose focus when you hover out, whereas the user menu items remain in focus :melting_face: .

working on realigning these.

stopfstedt:
@jrjohnson @dartajax @saschaben -  i switched the styles for hover and focus. dashed border now indicates that the given menu item is in focus. green background still indicates the active menu item. a blue background now highlights the menu item that the user is hovering over, with the exception of the currently active item (which remains green).

i've updated the PR description accordingly, but kept the screenshots in place since they are still good enough for demonstrative purposes.

please re-review, thanks.

dartajax (approved):
if this meets everybody else's expectations, and satisfies accessibility requirements, fine with me

dartajax:
Visual Diff - no outstanding weirdness - can be merged.

michaelchadwick (approved):

PR #9153, opened 2026-02-18
drop obsolete enlint-env config comments.

stopfstedt:
eslint v10 is ignoring these now and emitting a warning. stripping these out completely does not raise any new linter warnings, so these can be dropped without replacement.

please see https://eslint.org/docs/latest/use/configure/migration-guide#eslint-env-configuration-comments for reference.

fixes ilios/ilios#6923

### DIFF
--- a/packages/frontend/config/coverage.js
+++ b/packages/frontend/config/coverage.js
@@ -1,5 +1,3 @@
-/* eslint-env node */
-
 module.exports = {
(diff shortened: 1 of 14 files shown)


jrjohnson (approved):
Whew, I'm really happy these warnings turned out to be fixable this cleanly. I was worried!

PR #9154, opened 2026-02-18
correct delete permissions check on learner groups.

stopfstedt:
we must wait for two tracked props resolve before we can determine if a given learner group can be deleted or not.
doing this fixes a visual glitch where the delete buttons would appear activated during load and then become de-activated.

ditching the isLinked() getter as a consequence of this fix, it has become obsolete.

fixes ilios/ilios#6922


jrjohnson (approved):

dartajax (approved):
click-tested - this looks great

dartajax:
Visual Diff reviewed - seems fine to me - typical stuff - updated output though - nice - ready to merge

PR #9157, opened 2026-02-19
return the user to the list of courses in the same year as the currently given course's year.

stopfstedt:
fixes ilios/ilios#6894

### DIFF
--- a/packages/ilios-common/addon/components/course/back-to-courses.gjs
+++ b/packages/ilios-common/addon/components/course/back-to-courses.gjs
@@ -1,4 +1,5 @@
 import Component from '@glimmer/component';
+import { hash } from '@ember/helper';
 import { service } from '@ember/service';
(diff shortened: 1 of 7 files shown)


packages/ilios-common/addon/components/course/back-to-courses.gjs
```diff
@@ -13,10 +14,15 @@ export default class CourseBackToCoursesComponent extends Component {
       return false;
     }
   }
+
+  get year() {
+    return this.args.course?.year;
```
jrjohnson: Why the protection of `course?`? I can't think of any reason this would be null, but maybe its not fully resolved at some point?

jrjohnson (approved):

dartajax (approved):

dartajax:
I would consider merging this one but alas I cannot - why the Safari Tests are required and failing. Visual Diff is fine in any case.

PR #9210, opened 2026-03-04
adds side-loading for learnergroup offerings to session data loader.

stopfstedt:
this is a workaround to prevent ember-data from re-fetching offerings that should already be in the store when loading the offering calendar. without it, the offering in view will reload, forcing a re-render that is effectively closing the opened calendar prematurely.

refs https://github.com/ilios/ilios/issues/6931


jrjohnson (approved):

dartajax (approved):
click-tested - fine with me

PR #9215, opened 2026-03-06
Prevent Negative Margins

jrjohnson:
Adding a linting rule which will prevent negative margins. These are a problem in Ilios because they allow a component to exist outside of it's intended layout which makes working with multiple components in a grouping difficult.

### DIFF
--- a/packages/frontend/.stylelintrc.js
+++ b/packages/frontend/.stylelintrc.js
@@ -4,6 +4,9 @@ module.exports = {
   plugins: ['stylelint-scales'],
   extends: ['stylelint-config-recommended-scss'],
   rules: {
+    'declaration-property-value-disallowed-list': {
+      '/^margin/': ['/-[\\d.]/'],
+    },
     'property-disallowed-list': ['font-size', 'line-height'],
     'font-weight-notation': 'numeric',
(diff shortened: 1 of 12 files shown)


michaelchadwick (changes requested):
Sorry this took so long to get to, but I forgot about it and then it slipped through the cracks.

This all seems well and fair! Thankfully Ilios is not using negative margins much alredy. Tested some now-invalid margins and the linter catches them. The only issue is the last commit about screenshots doesn't seem to relate to the rest of the PR, so should it be in this?

Also, this is old enough to need some minor merge conflict resolution now ;P

jrjohnson:
Rebased. You're right the last commit isn't necessarily in the same vein, but it's required for the tests added in this to work because some of the components under test have a `{{let}}}` at the top which acts as an element and messes up the screenshots. 

I can pull it out into its own PR, but then it's kind of like... why? So I'm inclined to leave it here where it's clear why it was needed if we ever want to mess with it.

michaelchadwick:
@jrjohnson Fair enough! Approved.

michaelchadwick (approved):
🏆

dartajax (approved):

PR #9224, opened 2026-03-14
Fix diverging File Structure->Class names

michaelchadwick:
## SO!

This is an _extremely_ nit-picky PR, because the class name you give to a Javascript file _literally_ doesn't matter (I tested this with @stopfstedt once and realized you can just write `export default class extends Component` and it works fine).

## HOWEVER!

When you're programming, seeing a class name that doesn't match the filename or doesn't follow the same naming structure as the established convention, then it can be distracting.

### Established Convention

1. `if (componentInRoot) then file-name.gjs -> FileNameComponent`
2. `if (componentInSubdir) then subdir/file-name.gjs -> SubdirFileNameComponent`)

## THUS!

I went through both `/packages/frontend/` and `/packages/ilios-common` and normalized all the class names.

## END!


stopfstedt (approved):
you're right, no contest. LGTM.

PR #9243, opened 2026-03-25
restrict search results container height.

stopfstedt:
fixes ilios/ilios#6981

this is a global change and affects all areas where user search is in place.

examples:

#### offering form

image

image

#### user admin

image

etc.

### DIFF
--- a/packages/ilios-common/app/styles/ilios-common/components/user-search.scss
+++ b/packages/ilios-common/app/styles/ilios-common/components/user-search.scss
@@ -13,6 +13,7 @@
     box-shadow: 0 2px 2px var(--very-transparent-black);
     color: var(--black);
     left: 0;
+    max-height: 23rem;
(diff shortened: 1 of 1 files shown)


jrjohnson (commented):
Up for discussion - but making this smaller on smaller devices and larger on larger ones seems better.

dartajax (commented):
I can approve this as it is better than before and fixes the issue. I am kind of wondering what it might look like with ...

`max-height: 40dvh;` ... 

I hate to be a bother but can you try the 40 instead of 33 just to see? If it's just that one line that needs to be changed. I am definitely going to approve this either way. 40 means 40%?

dartajax (approved):
1/3 is fine and probably how it was in the past - so I approve this PR

jrjohnson (approved):

PR #9250, opened 2026-03-30
Update Ember to v6.11.2

jrjohnson:
In all apps and with all blueprints.

This did one weird thing, it removed the `no-implicit-route-model` feature flag. As far as I can tell from https://deprecations.emberjs.com/id/using-amd-bundles/ this flag is no longer relevant.


jrjohnson:
Going to replace this with a move to vite.

PR #9256, opened 2026-04-01
use a yellow instead of red as background for "publish as-is" option.

stopfstedt:
image (4)
image

this is to convey a warning/alert, rather than danger/error.

fixes ilios/ilios#6989


dartajax (dismissed):
I think this is great but it also needs to work at the course level, which still has the red background - can we do both with this PR?

stopfstedt:
> image

good catch, handled. :+1:

dartajax (approved):
great stuff - approved

michaelchadwick (approved):
I like how 1% of this change was fixing the problem, and 99% was making the test suite way better 👍

PR #9257, opened 2026-04-01
rm obsolete polyfill.

stopfstedt:
looks like all the functions we polyfilled in are now supported by modern browsers across the board, including Safari on mobile.

https://developer.mozilla.org/en-US/docs/Web/API/Crypto#browser_compatibility

in the spirit of spring cleaning - let's get rid of this.

shoutout to @erykmynn for having me look for obsolete polyfills and at https://43081j.com/2026/03/three-pillars-of-javascript-bloat.


jrjohnson (approved):
Nice!

dartajax (approved):
click tested - no issues found - seems good to me

PR #9258, opened 2026-04-02
fixes typo that was  causing wrong output in bulk session publisher.

stopfstedt:
fixes ilios/ilios#7031


dartajax (approved):
That's funny. I just noticed the image on a DartAjax release "Zone of Respectability (Vol. 2)" contained the caption "Zone of Respetability (Vol. 2)" - I never noticed it until the other day. This is like that.

PR #9260, opened 2026-04-03
Don't show Offering actions if user currently confirming deletion

michaelchadwick:
Fixes ilios/ilios#7040


dartajax (approved):
looks fine to me

stopfstedt (approved):

PR #9261, opened 2026-04-03
Fix main viewport scrolling when using keyboard to navigate Publication menus

michaelchadwick:
Refs ilios/frontend#9166
Fixes ilios/ilios#7044


stopfstedt (approved):
LGTM

PR #9263, opened 2026-04-04
replace absolute with relative paths for imports.

stopfstedt:
fixes ilios/ilios#7045

note that we can't effectively lint against absolute paths just yet, since some relative includes - for example from `frontend/config/environment` or any test subjects from within our test cases - are not working yet. 

i got as far as this in `frontend/eslint.config.mjs` on this, but that's evidently not enough to make it work.

```js
export default [
  // ...
  rules: {
     // ...
      'no-restricted-imports': [
        'error',
        {
          patterns: [
            {
              regex: 'frontend/(?!config)',
              message: 'Use relative imports within the frontend package.',
            },
          ],
        },
      ],
    },
];
```

for reference: https://eslint.org/docs/latest/rules/no-restricted-imports


jrjohnson (approved):
🚧 💯

dartajax (approved):
clicked around for a while - seems fine to me

PR #9270, opened 2026-04-07
override smooth scrolling if user prefers reduced motion.

stopfstedt:
todo: 
- [x] add test coverage.

### DIFF
--- a/packages/ilios-common/addon-test-support/ilios-common/helpers/set-prefer-reduced-motion.js
+++ b/packages/ilios-common/addon-test-support/ilios-common/helpers/set-prefer-reduced-motion.js
@@ -1,0 +1,48 @@
+/**
+ * This overrides the `window.matchMedia()` method to enable testing of reduction motion preference.
+ *
+ * ACHTUNG!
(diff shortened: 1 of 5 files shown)


michaelchadwick:
This all seems to work (tested on macOS for Chrome/Firefox/Safari), but it did cause me to notice that the `smooth` scrolling doesn't seem to work on macOS Firefox, regardless of any setting. In other words, it just always does `instant`. Can you confirm?

michaelchadwick (approved):
Regardless of Firefox for macOS ignoring the `smooth` behavior in any case (maybe just my machine), this works.

jrjohnson (approved):

jrjohnson:
A possible future enhancement, utterly unneeded, but for thinking about - 
Create a service `browserConfig` or `motionPreference` with a getter for `prefersReducedMotion`.

Then the API for checking this becomes `this.browserConfig.prefersReducedMotion` and if we want to mess with it in tests we can either mock the service or we can add a test-specific override to it. Which would match the other places we do this kind of thing.

dartajax:
I'm not going to auto-merge this one until I test manually a bit more - approved Visual Diff in the meantime.

PR #9278, opened 2026-04-08
Update Bulk Session Publish workflow terminology

michaelchadwick:
Fixes ilios/ilios#6852

image

image

### DIFF
--- a/packages/ilios-common/addon-test-support/ilios-common/page-objects/components/publish-all-sessions.js
+++ b/packages/ilios-common/addon-test-support/ilios-common/page-objects/components/publish-all-sessions.js
@@ -10,6 +10,10 @@ import {
 
 const definition = {
   scope: '[data-test-publish-all-sessions]',
+  header: {
+    scope: '[data-test-header]',
+    title: text('[data-test-title]'),
+  },
   unpublishableSessions: {
     scope: '[data-test-unpublishable]',
     title: text('> div > [data-test-title]'),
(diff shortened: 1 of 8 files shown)


stopfstedt (approved):
LGTM

dartajax:
The `Anger Inducer TM` is back ...

This is from my notes and could be applied in a new ticket if out of scope for this one.

Three sections total (New proposal) 

- Incomplete Sessions: cannot publish (x) – no change at all except for word order 
- Published Sessions (x) – show them all starting with the ones that have zero “missing” items – and then maybe have ILM’s which are usually “lacking” offerings be second in line, and then the remaining ones.
- Not Published Sessions: for review (x) – show all sessions with offerings or ILM’s with due dates or post-req – like above start with the ones in the list with have zero “missing” items – I miss “Unpublished” myself verbiage-wise – same sort as above with “Published Sessions (x)”

The final section requiring review is the area where the action check boxes would apply since a decision should (or can) be made regarding the session's publication status. The first two are basically for reference only since ... 

- a. the session cannot be published - it's incomplete (first section)
- b. the session has already been published - the action buttons could be available here I guess but am not too sure about that (second section)

dartajax:
Sort order could be specified by the user actually based on column headings.

dartajax:
If my comments are out of scope for this PR, let me know and I will be glad to do a follow-up ticket after approving and merging this one.

michaelchadwick:
@dartajax Great thoughts on how to further make this section better. Trying to do this piecemeal for now, so currently out of scope, though. Please make a separate ticket.

dartajax (approved):

dartajax:
I did the Visual Diff stuff manually by clicking and testing around. Seems fine - tagged it for merge.

PR #9279, opened 2026-04-08
LearnerGroup can now be deleted without confirmation if there are no members or subgroups

michaelchadwick:
Fixes ilios/ilios#6870

This PR does what was requested, but I'm not sure I like it ;P Maybe if there's no confirmation, some kind of flash message needs to pop up upon deletion? Also, the inconsistency between "just does it" vs. "needs extra step to do it" is odd.


michaelchadwick:
Discussed with team and decided not something we want to add.

PR #9280, opened 2026-04-09
School->SessionType-Visualization->Vocab by SessionType breadcrumb replacement

michaelchadwick:
Fixes ilios/ilios#6895

The breadcrumbs on the two routes `school/session-type-visualize-vocabularies` and `school/session-type-visualize-vocabulary` are confusing because they contain links that go nowhere and so I replaced them with simpler `Back to School` on the former and `Back to Vocabularies` on the latter.

To test: go to https://demo.iliosproject.org/schools/1?schoolSessionTypeDetails=true, choose a Visualization icon link, and then click into one of the vocabularies. You can then click `Back to Vocabularies` and then `Back to School` to get all the way back to the School.

image

### DIFF
--- a/packages/frontend/app/components/school/session-type-visualize-vocabularies.gjs
+++ b/packages/frontend/app/components/school/session-type-visualize-vocabularies.gjs
@@ -1,5 +1,6 @@
 import { LinkTo } from '@ember/routing';
 import t from 'ember-intl/helpers/t';
+import { hash } from '@ember/helper';
 import { pageTitle } from 'ember-page-title';
 import VisualizeSessionTypeVocabulariesGraph from './visualize-session-type-vocabularies-graph';
 <template>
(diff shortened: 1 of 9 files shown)


stopfstedt (approved):
LGTM

dartajax:
I'm sure you can imagine what I might request being the pain that I am. 

I take it there is no way to return the user to exactly where they were before. User gets routed to the top of the screen and from there can scroll down to find the session type with which they were just tinkering. Or maybe just back to the session types area of the screen? Let me know that this is out of scope and annoying (which I assume it is) and I will approve this PR and merge it.

michaelchadwick:
@dartajax This was mainly to fix the odd breadcrumbs with supposed links that were not, and simplify the navigation back to the School of which you were managing. As far as where on the page it takes you I think that is out of scope, but should be ticketed separately.

dartajax (approved):

dartajax:
I manually tested this and approved the Visual Diff stuff as a result of my own testing. There is too much noise over there right now.

PR #9283, opened 2026-04-13
rm redundant dependency.

stopfstedt:
we're getting this via ilios-common already, no need to re-declare this in frontend explicitly.


michaelchadwick (changes requested):
Removing this breaks Ilios for me (WSOD, essentially). And our tests also fail.

```
loader.js:247 Uncaught (in promise) Error: Could not find module `@fortawesome/free-solid-svg-icons` imported from `(require)
```

stopfstedt:
@michaelchadwick  thanks for checking. looks like we need to keep this dependency since `ilios-common` evidently can't make its dependency on this package available to `frontend`. bummer.

PR #9281, opened 2026-04-10
Add tooltips to disabled trashcan icons, explaining disablility

michaelchadwick:
Fixes ilios/ilios#6868

PR #9293, opened 2026-04-17
Normalize disabling actions when confirming deletion

michaelchadwick:
Fixes ilios/ilios#7072
Fixes ilios/ilios#7073

Relies on https://github.com/ilios/frontend/pull/9281 going through first

Sections to check
- [x] Course
- [x] Course Objective
- [x] Session
- [x] Session Objective
- [x] Session Offering
- [x] Course/Session Learning Material
- [x] Learner Group
- [x] Instructor Group
- [x] School Vocabulary
- [x] School Session Type
- [x] Program
- [x] Program Year
- [x] Program Year Objective
- [x] Curricular Inventory Report
- [x] Sequence Block

Additional things
- [x] Add/update tests

### DIFF
--- a/packages/frontend/app/components/courses/list.gjs
+++ b/packages/frontend/app/components/courses/list.gjs
@@ -191,6 +191,7 @@ export default class CoursesListComponent extends Component {
               <ListItem
                 @course={{course}}
                 @coursesForRemovalConfirmation={{this.coursesForRemovalConfirmation}}
+                @showRemoveConfirmation={{includes course.id this.coursesForRemovalConfirmation}}
                 @savingCourseIds={{this.savingCourseIds}}
                 @lockCourse={{fn (perform this.lockCourse)}}
                 @unlockCourse={{fn (perform this.unlockCourse)}}
--- a/packages/frontend/app/components/curriculum-inventory/report-list-item.gjs
+++ b/packages/frontend/app/components/curriculum-inventory/report-list-item.gjs
@@ -120,11 +120,6 @@ export default class CurriculumInventoryReportListItemComponent extends Componen
             class="link-button{{if this.showRemoveConfirmation ' disabled'}}"
             type="button"
             {{on "click" (set this "showRemoveConfirmation" true)}}
-            aria-label={{if
-              this.showRemoveConfirmation
-              (t "general.disabledByConfirmation")
-              (t "general.remove")
-            }}
             title={{if
               this.showRemoveConfirmation
               (t "general.disabledByConfirmation")
--- a/packages/frontend/app/components/curriculum-inventory/report-list-item.gjs
+++ b/packages/frontend/app/components/curriculum-inventory/report-list-item.gjs
@@ -97,7 +97,6 @@ export default class CurriculumInventoryReportListItemComponent extends Componen
             href={{@report.absoluteFileUri}}
             rel="noopener noreferrer"
             target="_blank"
-            disabled={{this.showRemoveConfirmation}}
             class={{if this.showRemoveConfirmation "disabled"}}
             data-test-download
           ><FaIcon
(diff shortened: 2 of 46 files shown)


michaelchadwick:
NTS: <button> only need `title`, and `<FaIcon>` does not need a `title`. Also finish tests.

packages/frontend/app/components/courses/list.gjs
```diff
@@ -191,6 +191,7 @@ export default class CoursesListComponent extends Component {
               <ListItem
                 @course={{course}}
                 @coursesForRemovalConfirmation={{this.coursesForRemovalConfirmation}}
+                @showRemoveConfirmation={{includes course.id this.coursesForRemovalConfirmation}}
```
jrjohnson, line 194: Moving this up one level is really nice cleanup, love it!

packages/ilios-common/addon/components/offering-manager.gjs
```diff
@@ -245,26 +245,42 @@ export default class OfferingManagerComponent extends Component {
...
+                  (t "general.edit")
+                }}
+                disabled={{this.showRemoveConfirmation}}
                 {{on "click" this.toggleIsEditing}}
                 data-test-edit
               >
-                <FaIcon @icon={{faPenToSquare}} class="enabled" />
+                <FaIcon
+                  @icon={{faPenToSquare}}
+                  class="{{if this.showRemoveConfirmation ' disabled' ' enabled'}}"
```
jrjohnson: Extra white space here. See this a few other places here where the space is added before the class. I don't think it's. big deal and it shouldn't cause any issues, just pointing it out. Wonder if we could lint this?

packages/frontend/app/components/curriculum-inventory/report-list-item.gjs
```diff
@@ -98,30 +97,52 @@ export default class CurriculumInventoryReportListItemComponent extends Componen
             href={{@report.absoluteFileUri}}
             rel="noopener noreferrer"
             target="_blank"
+            disabled={{this.showRemoveConfirmation}}
```
jrjohnson: As far as I can tell the `disabled` attribute doesn't apply to a link. I'm ok with a css only solution here, we probably don't need to do anything fancier like removing the href.

packages/ilios-common/app/styles/ilios-common/components/body.scss
```diff
@@ -22,6 +22,10 @@ body {
 a,
 .link {
   @include m.ilios-link;
+
+  &.disabled {
+    cursor: default;
```
jrjohnson: Would prefer [pointer-events: none](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/pointer-events) here which prevents the click entirely. Probably useful for the buttons too, but definitely the links.

jrjohnson (changes requested):
Spot checked and clicked a few, this is extremely thorough and very very nice. Found one thing to change on the way links are handled.

Not needed for this PR, but for followup:
1. Consolidate disabled button and disabled icon style. Icons in disabled buttons shouldn't also need their own class.
2. We don't need title & aria label. Title become the label. Only needed if they're different. This is something we should fix everywhere, including a lot of places that don't show up in this PR.

packages/ilios-common/app/styles/ilios-common/components/body.scss
michaelchadwick: I was not aware of this attribute, but thanks for pointing it out! I will add it to `&.disabled:active` because otherwise the tooltip does not appear.

packages/ilios-common/addon/components/offering-manager.gjs
michaelchadwick: I see this and a couple others in my PR that need fixing. There are even a few outside of the PuRview but I'll leave those alone.

jrjohnson (approved):

dartajax (approved):
click tested - looks good to me

dartajax:
#9281 already merged so let's get this one in too

PR #9295, opened 2026-04-20
Session Publication Review Re-Styling

michaelchadwick:
Fixes ilios/ilios#6907

This is a little styling update for the Session Publication Review section to make it more consistent with the page, and, most importantly, imbue it with a heading.


jrjohnson (approved):
Will accept the scroll delay as a quick fix, but probably needs a deeper look to see why it is necessary. Once that element is in the dom it should be scrollable, if it's not we probably have an issue with our modifier.

michaelchadwick:
@jrjohnson there are now a couple places where the scroll needs a slight delay, and they may be related

dartajax (approved):

dartajax:
manually tested - and @jrjohnson did too - ok to get this one merged - an improvement indeed

PR #9297, opened 2026-04-21
Improve School Vocab by Session Type Visualization loading

michaelchadwick:
Fixes ilios/ilios#6871

Not sure of the correct way to do this, so experiment thus far.


michaelchadwick:
Never made it very far with this, and will need to re-do it after re-orging the app packages, so closing for now.

PR #9300, opened 2026-04-22
Make all manual instances of `faSpinner` use `LoadingSpinner` component

michaelchadwick:
Fixes ilios/ilios#7085

There should be no visual change or difference in rendered templates. This PR merely replaces all manual instances of `<FaIcon @icon={{faSpinner}} @spin={{true}} />` with `<LoadingSpinner />`, which does the same thing.

The `LoadingSpinner` component, however, now takes `@class` and `@size` arguments, since Ilios has a couple non-standard uses of the `faSpinner` icon, and now they can be replaced, too.

### DIFF
--- a/packages/ilios-common/addon/components/course/materials.gjs
+++ b/packages/ilios-common/addon/components/course/materials.gjs
@@ -17,6 +17,7 @@ import includes from 'ilios-common/helpers/includes';
 import capitalize from 'ilios-common/helpers/capitalize';
 import formatDate from 'ember-intl/helpers/format-date';
 import { pageTitle } from 'ember-page-title';
+import LoadingSpinner from 'ilios-common/components/loading-spinner';
 import { faSpinner } from '@fortawesome/free-solid-svg-icons';
 
 const DEBOUNCE_DELAY = 250;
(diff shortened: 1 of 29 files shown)


packages/ilios-common/addon/components/course/materials.gjs
```diff
@@ -352,7 +353,8 @@ export default class CourseMaterialsComponent extends Component {
             {{#if this.isLoading}}
               <tr>
                 <td colspan="9" align="center">
-                  <FaIcon @icon={{faSpinner}} class="orange" @size="2x" @spin={{true}} />
```
jrjohnson, line 355: I think I would rather leave this one case than support passing things through to the component. But I can be talked around.

jrjohnson (dismissed):

packages/ilios-common/addon/components/course/materials.gjs
michaelchadwick: There are two other cases where custom options are given, which is why I augmented `LoadingSpinner` to support them. I even wrote a test for it! :P One of the benefits of using a custom component is the ability to pass arguments to it.

Thus, I think keeping them consistent is more important.

jrjohnson: I'm worried we're passing arguments through the compliment. They mean nothing to us, we're just forwarding on to FAIcon. But I'm not going to defend the hill. It's fine.

stopfstedt (approved):
looks good to me.

dartajax (approved):

PR #9301, opened 2026-04-23
Make ProgramYear list sortable by startYear and default to descending

michaelchadwick:
Fixes ilios/ilios#6873

### DIFF
--- a/packages/frontend/app/components/program-year/list.gjs
+++ b/packages/frontend/app/components/program-year/list.gjs
@@ -1,5 +1,6 @@
 import Component from '@glimmer/component';
 import { cached, tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
 import { TrackedAsyncData } from 'ember-async-data';
 import { sortBy } from 'ilios-common/utils/array-helpers';
 import { task } from 'ember-concurrency';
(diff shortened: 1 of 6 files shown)


dartajax:
Screenshot 2026-04-23 at 4 17 56 PM

Screenshot 2026-04-23 at 4 19 00 PM

michaelchadwick:
@dartajax the first image is meant to be ascending (A->Z), and the second image, Program Years, is meant to be descending (9->1). It's using numeric sorting since they are years.

packages/frontend/app/components/program-year/list.gjs
```diff
@@ -38,6 +42,9 @@ export default class ProgramYearListComponent extends Component {
   }
 
   get sortedProgramYears() {
+    if (this.sortBy.includes('startYear')) {
+      return this.programYears.slice().sort((a, b) => a.startYear - b.startYear);
```
stopfstedt: please use [`toSorted()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) here instead.

```suggestion
      return this.programYears.toSorted((a, b) => a.startYear - b.startYear);
```

stopfstedt (changes requested):
one minor wrinkle, otherwise this looks good to me.

stopfstedt (approved):
LGTM

dartajax (approved):

dartajax:
Wow - it's hard to see the difference between 9-1 vs 0-1 - oops. I approved this and the Visual Diff - the first column got a bit wider and would have resulted in many screen shot capture diffs in all likelihood - either way this seems fine to me at this point.

PR #9302, opened 2026-04-24
Fix incorrect spacing between Session Term lists

michaelchadwick:
Fixes ilios/ilios#7091

The fix was to _only_ remove margin from the last session terms list, not all of them.


stopfstedt (approved):
LGTM

dartajax (approved):

PR #9306, opened 2026-04-29
Hide 'skip to main content' link from visual users in Safari

michaelchadwick:
Fixes ilios/ilios#7095

Safari needs the `ember-a11y-skip-link` element to be hidden by either `clip` (deprecated) or `clip-path` to be truly hidden. Also, there may be some kind of issue with the focus-ring applied to the element when clicking the page, which causes it to "drift" down the page.

### DIFF
--- a/packages/frontend/app/components/ilios-header.gjs
+++ b/packages/frontend/app/components/ilios-header.gjs
@@ -74,8 +74,12 @@ export default class IliosHeaderComponent extends Component {
       <NavigationNarrator
         @navigationText={{t "general.navigationCompleteText"}}
         @skipText={{t "general.skipToMainContent"}}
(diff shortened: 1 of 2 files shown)


michaelchadwick:
NTS: change `skipLink` to true, and add our own with our own styles

jrjohnson (dismissed):

jrjohnson:
Test failure is legit, need to fix the tab order of this new skip link.

jrjohnson (approved):
grumble grumble, would rather use our own ID and not theirs for this, but not so much that I need to stop the work.

dartajax (approved):
nice fix - sketchy bug - tested on Safari

PR #9309, opened 2026-04-29
If Learner Group has no members, managing it will no longer show empty table with headers

michaelchadwick:
Fixes ilios/ilios#7075


dartajax:
not sure we even need this one anymore since the "Manage Members" is disabled in an empty group anyway - will test more though

stopfstedt (approved):
nice.

dartajax (approved):

PR #9311, opened 2026-04-29
Add fixed table header and column sorting to Students Requiring Cohort Assignment

michaelchadwick:
Fixes ilios/ilios#6876

This will also move the list filter into the manager itself, as suggested in the issue.


stopfstedt (approved):
LGTM

dartajax (approved):
nice

PR #9314, opened 2026-05-01
Improve Screenshot Capture 

jrjohnson:
When an individual test fails we don't really want screenshots capture and comparison to fail without reaching the comparison step.

Also reverts #9310 as, hopefully, no longer needed now that test failure won't cause screenshot failure.

### DIFF
--- a/.github/workflows/visual_diff.yml
+++ b/.github/workflows/visual_diff.yml
@@ -53,6 +53,7 @@ jobs:
         if: steps.screenshot-cache.outputs.cache-hit != 'true'
         run: pnpm install
       - name: Capture Screenshots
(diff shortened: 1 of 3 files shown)


michaelchadwick:
Just to clarify: when you say "When a test fails we don't really want screenshots to fail" did you mean to say "When a _specific screenshot test fails_ we don't really want _all screenshot tests_ to fail?"

jrjohnson:
Edited the PR description. Hopefully that's clearer. Also you can see what happens in this PR now. The mesh test fails, so we get a visual diff failure because 4 screenshots didn't get captured.

stopfstedt (approved):
this tracks.

PR #9320, opened 2026-05-08
Remember expanded years in Curriculum Reports

michaelchadwick:
Fixes ilios/ilios#7121

This PR moves the `expandedYears` variable out of the component and into the route so it can be kept track of in the URL and properly expand/collapse when refreshing or sharing a link.

~~TODO: fix tests~~
TODO: Add more tests

### DIFF
--- a/packages/frontend/app/components/reports/curriculum.gjs
+++ b/packages/frontend/app/components/reports/curriculum.gjs
@@ -134,6 +134,8 @@ export default class ReportsCurriculumComponent extends Component {
         />
         <ChooseCourse
           @selectedCourseIds={{@selectedCourseIds}}
+          @expandedYears={{@expandedYears}}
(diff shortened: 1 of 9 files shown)


jrjohnson (dismissed):
Looks great. needs some more tests.

`ReportsCurriculumChooseCourseComponent` needs tests for clicking a year firing the correct data to `this.args.setExpandedYears` and for updating the UI when that passed value changes.

Now that the router is involved also need acceptance tests for choosing years that checks propagation up to `currentUrl` and checks that setting the URL when the page loads checks the right boxed.

stopfstedt (approved):
LGTM

dartajax (approved):
works for me

dartajax:
I was totally committed to manually merging this one but if "auto-merge" can take care of it, that's fine too.

PR #9322, opened 2026-05-11
Breadcrumb sections above and below link text now clickable

michaelchadwick:
Fixes ilios/ilios#7129

This should make the text, and the area above and below the text, clickable. The area to the left and right (i.e. the > parts) are not currently clickable because they are created by `span::before` and `span::after` CSS attributes.

Will follow this up with a re-write of breadcrumbs as a proper component so future accommodations are much easier.

### DIFF
--- a/packages/ilios-common/app/styles/ilios-common/components/breadcrumbs.scss
+++ b/packages/ilios-common/app/styles/ilios-common/components/breadcrumbs.scss
@@ -9,6 +9,8 @@
   text-align: left;
 
   a {
+    display: block;
+
     &:focus,
     &:hover {
       color: var(--white);


dartajax:
I hate to be that guy but it still doesn't work in the "triangle" areas but is that expected at this stage?

dartajax:
> I hate to be that guy but it still doesn't work in the "triangle" areas but is that expected at this stage?

No problem if that is the case - just wanted to point it out.

michaelchadwick:
@dartajax The triangle areas are *not* expected to work at this stage, no. That would require a complete restructuring of how breadcrumbs work, which I plan to do in a followup PR once this goes in. I updated the description of this PR to reflect that, too.

dartajax:
> @dartajax The triangle areas are _not_ expected to work at this stage, no. That would require a complete restructuring of how breadcrumbs work, which I plan to do in a followup PR once this goes in. I updated the description of this PR to reflect that, too.

coolio - I will approve this

dartajax (approved):

jrjohnson (approved):

PR #9325, opened 2026-05-12
added a period to admin console full sentence captions

dartajax:
Fixes https://github.com/ilios/ilios/issues/5591

This is what I normally do in the user guide. I added periods "." to the admin console messages regarding user assignment. 

If there is something wrong with the way I have done this ...

- I apologize.
- Please terminate my effort however it is necessary to do. 
- Let me know how I can fix this - if that is even possible. 

I only updated the translation files with the period where I thought it was applicable. That's why they call them pull "requests" I guess. I am asking if this is cool ... or not.


michaelchadwick (approved):
👍

michaelchadwick (changes requested):
Legitimate failed tests that have string checking that needs to be updated

michaelchadwick (approved):

PR #9341, opened 2026-05-21
Lock Older Snapdom

jrjohnson:
Now that we have stable fonts with MSW I'm trying an older version of snapdom to see if it can help with some other issues we have with things moving around on the page.


jrjohnson:
Ok, this helps a lot. We lose the checkboxes, but get a lot more stability in the images themselves. All the differences in the candidate images are on the correct side.

michaelchadwick (approved):

PR #9343, opened 2026-05-21
MSW Improvements

jrjohnson:
Couple of things that were on my mind. Probably more to do, but this seems like a good set.

Probably best reviewed as individual commits.


packages/ilios-common/addon-test-support/msw/utils/identity-manager.js
```diff
@@ -1,78 +1,27 @@
...
-
-function isNumber(n) {
-  return (+n).toString() === n.toString();
-}
-
-/**
- * @param {boolean} returnIdAsString
- */
 class IdentityManager {
+  #usedIds = [];
```
michaelchadwick, line 2: Why variable name use octothorpe?

stopfstedt (approved):

packages/ilios-common/addon-test-support/msw/utils/identity-manager.js
michaelchadwick: Ah private member

PR #9344, opened 2026-05-22
Wait for `authTypeConfig` to load before searching directory

michaelchadwick:
Fixes ilios/ilios#7153


jrjohnson (approved):
It is marvelous, it is not a hack 😁

PR #9348, opened 2026-05-26
Whizbang Learner Group Scroll Improvements

michaelchadwick:
Fixes ilios/ilios#5634

* Current LG, if empty, shows a bottom border to separate it from "All other members" table
* Both "Current" and "All Other Members" tables now with slight side margins for better contextual display
* Headers of each table are now "sticky" (just like the table header columns)
* "All other members" now hides table column headers if no content


stopfstedt (approved):
leaving out the ongoing convo about some orange divider line in the associated ticket, this looks good to me.

dartajax:
I would like to continue testing this one before approving - hope that's cool. Comments were submitted and discuss in Slack.

dartajax (approved):
couldn't reproduce what I saw yesterday and I know from @michaelchadwick's commit message that he submitted a fix for it anyway - looks good to me

PR #9355, opened 2026-05-28
replace sass:math functions with calc() or fixed value.

stopfstedt:
saw this in passing. refs ilios/ilios#6608

### DIFF
--- a/packages/ilios-common/app/styles/ilios-common/components/progress-bar.scss
+++ b/packages/ilios-common/app/styles/ilios-common/components/progress-bar.scss
@@ -1,5 +1,3 @@
-@use "sass:math";
-
 .progress-bar {
   background-color: var(--super-light-grey);
   border: 1px solid var(--lightest-grey);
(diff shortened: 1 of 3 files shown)


packages/ilios-common/app/styles/ilios-common/components/progress-bar.scss
```diff
@@ -14,7 +13,7 @@
     background-repeat: repeat-x;
     background-size: 40px 40px;
     border: 1px solid var(--dark-orange);
-    border-radius: math.div(3px, 1.5);
+    border-radius: calc(3px / 1.5);
```
michaelchadwick: Is there any reason to not just use `border-radius: 1.5px` here?

stopfstedt: that would be `2px`, but yes.

stopfstedt: handled.

michaelchadwick (approved):

PR #9356, opened 2026-05-28
rm all imports of the SASS color module.

stopfstedt:
refs https://github.com/ilios/ilios/issues/6608


michaelchadwick:
There are really ~30 completely useless imports? I can't find an actual legit usage of that import anywhere, so Ilios must've used it a while ago.

stopfstedt:
> There are really ~30 completely useless imports? I can't find an actual legit usage of that import anywhere, so Ilios must've used it a while ago.

precisely.

michaelchadwick (approved):

PR #9357, opened 2026-05-29
Publication Review Buttons and Checkboxes now using Radio Button Groups

michaelchadwick:
Fixes ilios/ilios#6854

This changes the "Publish All As-Is" and "Mark All As Scheduled" bulk action buttons for Publication Review into a semantically clearer radio button group. It also changes individual session row checkboxes into radio button groups.

### DIFF
--- a/packages/ilios-common/addon/components/publish-all-sessions.gjs
+++ b/packages/ilios-common/addon/components/publish-all-sessions.gjs
@@ -27,6 +27,7 @@ export default class PublishAllSessionsComponent extends Component {
 
   @tracked totalSessionsToSave;
   @tracked currentSessionsSaved;
+  @tracked userSelectedAction = 'scheduleAll';
 
   @tracked userSelectedSessionsToPublish = [];
   @tracked userSelectedSessionsToSchedule = [];
@@ -245,12 +246,14 @@ export default class PublishAllSessionsComponent extends Component {
   publishAllAsIs() {
     this.userSelectedSessionsToSchedule = [];
     this.userSelectedSessionsToPublish = [...this.overridableSessions];
+    this.userSelectedAction = 'publishAllAsIs';
   }
 
   @action
   scheduleAll() {
     this.userSelectedSessionsToPublish = [];
     this.userSelectedSessionsToSchedule = [...this.overridableSessions];
+    this.userSelectedAction = 'scheduleAll';
   }
 
   @action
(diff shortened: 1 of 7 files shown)


packages/ilios-common/addon/components/publish-all-sessions.gjs
```diff
@@ -74,6 +75,22 @@ export default class PublishAllSessionsComponent extends Component {
     );
   }
 
+  get sessionsToAllBePublished() {
+    if (this.sessionsToPublish?.length) {
+      return !this.sessionsToSchedule?.length ? true : false;
```
stopfstedt:
```suggestion
      return !this.sessionsToSchedule?.length;
```

packages/ilios-common/addon/components/publish-all-sessions.gjs
```diff
@@ -74,6 +75,22 @@ export default class PublishAllSessionsComponent extends Component {
...
+    }
+
+    return false;
+  }
+
+  get sessionsToAllBeScheduled() {
+    if (this.sessionsToSchedule?.length) {
+      return !this.sessionsToPublish?.length ? true : false;
```
stopfstedt:
```suggestion
      return !this.sessionsToSchedule?.length;
```

stopfstedt (changes requested):

packages/ilios-common/addon/components/publish-all-sessions.gjs
michaelchadwick: In fact, this could be simplified even further?

```
get sessionsToAllBePublished() {
  return !this.sessionsToSchedule?.length;
}
```

packages/ilios-common/addon/components/publish-all-sessions.gjs
michaelchadwick: In fact, this could be simplified even further?

```
get sessionsToAllBeScheduled() {
  return !this.sessionsToPublish?.length;
}
```

stopfstedt (approved):

dartajax (approved):
looks good to me - I like this - much improved

packages/ilios-common/addon/components/publish-all-sessions.gjs
stopfstedt: yes!

PR #9361, opened 2026-06-02
Lock Older Snapdom

jrjohnson:
Now that we have stable fonts with MSW I'm trying an older version of snapdom to see if it can help with some other issues we have with things moving around on the page.

This repeats #9341 after #9352 was accidentally merged and undid the lock.


stopfstedt (approved):
my bad, sorry.

PR #9367, opened 2026-06-03
Disable New Program Year save button if no available academic years

michaelchadwick:
Fixes ilios/ilios#7168

### DIFF
--- a/packages/frontend/app/components/program-year/new.gjs
+++ b/packages/frontend/app/components/program-year/new.gjs
@@ -56,6 +56,10 @@ export default class ProgramYearNewComponent extends Component {
     return findBy(this.availableAcademicYears, 'value', this.year);
   }
 
+  get noAvailableYears() {
+    return !this.selectedYear;
+  }
+
   saveNewYear = task({ drop: true }, async () => {
     await this.args.save(this.selectedYear.value);
   });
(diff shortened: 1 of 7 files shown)


dartajax:
What an excellent catch @michaelchadwick

dartajax (approved):
click tested - looks fine to me

packages/frontend/app/components/program-year/new.gjs
```diff
@@ -91,7 +95,13 @@ export default class ProgramYearNewComponent extends Component {
         <div class="buttons">
           <button
             type="button"
-            class="done text"
+            class="done text{{if this.noAvailableYears ' disabled'}}"
```
stopfstedt: the added class is not necessary if you use the `:disabled` pseudo class in  the stylesheet instead.

packages/frontend/app/styles/components/program-year/new.scss
```diff
@@ -21,5 +21,12 @@
 
   .buttons {
     @include m.ilios-form-buttons;
+
+    button.disabled {
```
stopfstedt: use the `:disabled` pseudo class instead.

```suggestion
    button:disabled {
```

stopfstedt (changes requested):
the fix works as intended. code looks good. the styles can be streamlined, please see line comments.

stopfstedt (approved):

PR #9374, opened 2026-06-05
New Curriculum Report: Course Competencies

michaelchadwick:
Fixes ilios/ilios#7167

### DIFF
--- a/packages/frontend/app/components/reports/curriculum/header.gjs
+++ b/packages/frontend/app/components/reports/curriculum/header.gjs
@@ -5,6 +5,7 @@ import { guidFor } from '@ember/object/internals';
 import t from 'ember-intl/helpers/t';
 import { on } from '@ember/modifier';
 import { eq, not } from 'ember-truth-helpers';
+import sortBy from 'ilios-common/helpers/sort-by';
 import CopyButton from 'ilios-common/components/copy-button';
 import perform from 'ember-concurrency/helpers/perform';
 import mouseHoverToggle from 'ilios-common/modifiers/mouse-hover-toggle';
@@ -182,7 +183,7 @@ export default class ReportsCurriculumHeaderComponent extends Component {
               <label data-test-report-selector>
                 {{t "general.run"}}
                 <select {{on "change" this.changeSelectedReport}}>
-                  {{#each this.reportList as |report|}}
+                  {{#each (sortBy "value" this.reportList) as |report|}}
                     <option
                       value={{report.value}}
                       selected={{eq report.value this.selectedReport.value}}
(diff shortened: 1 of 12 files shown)


dartajax (approved):
seems fine to me

packages/frontend/tests/integration/components/reports/curriculum/header-test.gjs
```diff
@@ -28,30 +28,30 @@ module('Integration | Component | reports/curriculum/header', function (hooks) {
...
       component.reportSelector.options[0].text,
-      'Session Objectives',
+      'Course Competencies',
       'report types selector has correct first option text',
     );
-    assert.ok(
+    assert.notOk(
       component.reportSelector.options[0].isSelected,
-      'report types selector FIRST option IS chosen',
+      'report types selector first option is not chosen',
```
jrjohnson, line 221: This doesn't make sense to me. The first option should always be picked. We should either clean up the list to be alphabetical (in a new PR) or put this new report not first in the list. Having the second report by the default just feels weird.

jrjohnson (changes requested):

packages/frontend/tests/integration/components/reports/curriculum/header-test.gjs
michaelchadwick: That's a fair point. I wasn't sure where to put this test in the list, but it being about Course information was both 1) alphabetical and 2) organizationally *first* to me, so I put it there. I will gladly make a followup PR to alphabetize the list and update tests accordingly.

jrjohnson (approved):

jrjohnson (changes requested):
Hold on... still needs tests. Both integration for the new test as well as tests added to `packages/frontend/tests/acceptance/reports/curriculum-test.js`.

packages/frontend/app/components/reports/curriculum/header.gjs
jrjohnson: It's worth it to alpha sort here (or in the getter) and not just rely on the order being correct.

packages/frontend/tests/acceptance/reports/curriculum-test.js
```diff
@@ -75,6 +75,15 @@ module('Acceptance | Reports - Curriculum Reports', function (hooks) {
         return { data: { courses } };
       };
     };
+
+    this.getCourseCompetenciesResponse = (assert) => {
+      return () => {
+        assert.step('API called');
```
jrjohnson: This should move down to just above the return. That was if the server call fails it will be much easier to see where.

jrjohnson (dismissed):
Looking great, two small tweaks.

dartajax (approved):
it looks good to me - click-tested and downloaded a sample

stopfstedt (approved):
LGTM

PR #9377, opened 2026-06-08
Adds input for JWT attributes in setupAuthentication helper.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/7174

### DIFF
--- a/packages/ilios-common/addon-test-support/ilios-common/helpers/setup-authentication.js
+++ b/packages/ilios-common/addon-test-support/ilios-common/helpers/setup-authentication.js
@@ -9,19 +9,15 @@ const defaultUserId = 100;
  * That new mock user will be "logged in" and returned.
  *
  * @param {object} userObject The user attributes, such as id and relationships, for the user to be created/authenticated.
+ * @param {object} jwtOptions Attributes to encode in the user session's JWT, for example audience ("aud") values.
  * @returns {Promise<object>} A promise resolving to the mock user model that was created and authenticated in the process.
  */
-export default async function (userObject = { id: defaultUserId }) {
+export default async function (userObject = { id: defaultUserId }, jwtOptions = {}) {
   // Establish user identity, throw an error if that fails.
   const userId = userObject && 'id' in userObject ? userObject.id : defaultUserId;
   if (!userId) {
     throw new Error('Invalid or missing user id');
   }
-
-  // Create a JWT object for the given user. It will be encoded later on.
-  const jwtObject = {
-    user_id: userId,
-  };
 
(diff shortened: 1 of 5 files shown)


packages/test-app/tests/integration/services/current-user-test.js
```diff
@@ -12,7 +13,7 @@ module('Integration | Service | Current User', function (hooks) {
   hooks.beforeEach(async function () {
     await authenticateSession({
       // this token de-serializes to object with "user_id:100" property/value
```
jrjohnson: This comment is a bit confusing with the refactor. I think it can go away now when it's clear what is being passed in.

packages/test-app/tests/unit/utils/jwt-decode-test.js
```diff
@@ -1,10 +1,10 @@
 import jwtDecode from 'test-app/utils/jwt-decode';
 import { module, test } from 'qunit';
+import { jwtEncode } from 'ilios-common';
 
 module('Unit | Utility | jwt decode', function () {
-  const token =
-    'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWV9.TJVA95OrM7E2cBab30RMHrHDcEfxjoYZgeFONFh7HgQ';
   test('it decodes a token', function (assert) {
+    const token = jwtEncode({ sub: '1234567890', name: 'John Doe', admin: true });
```
jrjohnson, line 7: Way better!

packages/ilios-common/addon-test-support/ilios-common/helpers/setup-authentication.js
jrjohnson: My preference would be for this to use the same spread operator as you're doing to make the `jwtObject`, but either way they should be the same syntax.

jrjohnson (changes requested):

packages/ilios-common/addon-test-support/ilios-common/helpers/setup-authentication.js
stopfstedt: handled :heavy_check_mark: . using the spread operator in place of object.assign. way more readable to me.

packages/test-app/tests/integration/services/current-user-test.js
stopfstedt: removed. :heavy_check_mark:

jrjohnson (approved):

PR #9378, opened 2026-06-08
rm unnecessary root user creation in the mock current session.

stopfstedt:
This either removes the root flag altogether without replacement if possible,
or scopes this down to the perms granted by providing user relationships instead.

I elaborated on the remaining root flags in code comments.

In some cases, additional data massaging was necessary due to changes in
fixture creation order.

fixes ilios/ilios#7179


jrjohnson (approved):

jrjohnson:
Visual diffs show some possibly unintended changes. It looks like in some places permissions have grown (and the diffs have more admin content). Needs more investigation.

stopfstedt:
> Visual diffs show some possibly unintended changes. It looks like in some places permissions have grown (and the diffs have more admin content). Needs more investigation.

i reviewed the viz diffs. all of these are on point.

PR #9380, opened 2026-06-09
adds LTI login route and adds layout restrictions based on application-scope relayed in auth token

stopfstedt:
fixes ilios/ilios#7150

### DIFF
--- a/packages/frontend/app/routes/lti-login.js
+++ b/packages/frontend/app/routes/lti-login.js
@@ -1,6 +1,6 @@
 import Route from '@ember/routing/route';
 import { service } from '@ember/service';
-import { decodedLtiTokenHasLtiAudienceClaims, jwtDecode } from 'ilios-common/utils/jwt-utils';
+import { decodedJwtHasLtiAudienceClaims, jwtDecode } from 'ilios-common/utils/jwt-utils';
 
 export default class LtiLoginRoute extends Route {
   @service serverVariables;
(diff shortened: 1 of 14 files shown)


packages/frontend/tests/acceptance/application-layout-test.js
```diff
@@ -0,0 +1,42 @@
...
+      assert.dom('header.ilios-header').doesNotExist();
+      assert.dom('.ilios-logo').doesNotExist();
+      assert.dom('[data-test-ilios-navigation]').doesNotExist();
+      assert.dom('#main').exists();
+      assert.dom('footer.ilios-footer').doesNotExist();
+      await takeScreenshot(assert);
+    },
+  );
+
+  test.each(
```
jrjohnson, line 26: rad

jrjohnson (approved):

packages/frontend/app/routes/lti-login.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+    const jwt = await this.getNewToken(token);
+    await this.session.authenticate('authenticator:ilios-jwt', { jwt });
+    this.router.transitionTo('index');
+  }
+
+  async getNewToken(ltiToken) {
+    const apiHost = this.iliosConfig.apiHost;
+    const url = `${apiHost}/auth/token`;
+    const response = await fetch(url, {
```
jrjohnson, line 24: I'd like to see this in the `fetch` service, just for consolidation. This can be followup.

jrjohnson (approved):

jrjohnson:
Note: Testing with the LTI server in an LMS shows an issue with the session storage. For frontend we store the session data (mainly the JWT) in a cookie. But for the LTI it's 'ephemeral' and is not persisted anywhere. The cookie based storage doesn't work in the LTI context as storing the cookie would cause the user to be logged in to Ilios, but in an LTI context. If they visit the application directly they will have a broken experience.  For this solution to work we'll need to choose the store dynamically. Probably in the `ilios-jwt` authenticator based on the LTI claims.

PR #9387, opened 2026-06-15
Show inactive terms in taxonomy manager.

stopfstedt:
refs https://github.com/ilios/ilios/issues/7183

since we're putting inactive terms on the screen now, we have to consider a few new rules:

- inactive terms will be suffixed with a ` (inactive)` label
- children of inactive terms will now also be displayed, but won't be selectable
- children of inactive terms that are active  _won't_ be suffixed with the ` (inactive)` label
- inactive terms that are not already selected cannot be selected
- inactive terms and children of inactive terms that are already selected can be de-selected, but will not be re-selectable.
- any non-selectable or non-de-selectable terms will not be rendered as buttons, they won't have a hover state and their mouse pointer won't be the finger.

image

### DIFF
--- a/packages/ilios-common/addon/components/taxonomy-manager-terms-list-item.gjs
+++ b/packages/ilios-common/addon/components/taxonomy-manager-terms-list-item.gjs
@@ -3,21 +3,23 @@ import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { guidFor } from '@ember/object/internals';
 import { on } from '@ember/modifier';
+import t from 'ember-intl/helpers/t';
 import mouseHoverToggle from 'ilios-common/modifiers/mouse-hover-toggle';
 import set from 'ember-set-helper/helpers/set';
 import IliosTooltip from 'ilios-common/components/ilios-tooltip';
 import FaIcon from '@fortawesome/ember-fontawesome/components/fa-icon';
 import { faPlus, faXmark } from '@fortawesome/free-solid-svg-icons';
+import { and, or } from 'ember-truth-helpers';
 
 export default class SelectableTermsListItemComponent extends Component {
   @tracked isHovering;
 
-  get termsListItemButtonId() {
+  get termsListItemId() {
     return `taxonomy-manager-terms-list-item-button-${guidFor(this)}`;
   }
 
-  get termsListItemButtonElement() {
-    return document.getElementById(this.termsListItemButtonId);
+  get termsListItemElement() {
+    return document.getElementById(this.termsListItemId);
   }
 
   get isSelected() {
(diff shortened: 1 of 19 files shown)


dartajax (dismissed):
functional review - looks good

michaelchadwick (dismissed):
Click-tested and suggested making the inactive terms more subdued, which was handled. Seems good to me.

dartajax:
image

There is no way to know anything is awry until entering maintenance mode to see ...

image

I would posit that it would be helpful to know from a higher viewpoint that the term added to this course is in some way shape or form "inactive" - maybe just keep the "inactive" at the sub-term level if the parent is "inactive" - sorry about my idea here - I will wait on the merge for now.

dartajax:
or `"Soundscape >> Treble (inactive)"` in the sense that the "inactive" state has been passed from the parent down

dartajax (dismissed):
I left a couple of tedious comments about how the "inactive" label might be helpful if it appeared at the higher level as well so users will know that an "inactive" term has been applied to the course, session, or other curricular object. Two suggestions for the label appear in my comments. This is regarding how sub terms for inactive parents have been applied. So either ...

1. `[Vocabulary] >> [Term] (inactive)`
or ...
2. `[Vocabulary] (inactive) >> [Term]`

We can dismiss this and I will approve either way - I just wanted to point out from my perspective what I was thinking. I'd say team discussion might be good on this one - outlying situation anyway I guess.

stopfstedt:
> I left a couple of tedious comments about how the "inactive" label might be helpful if it appeared at the higher level as well so users will know that an "inactive" term has been applied to the course, session, or other curricular object. Two suggestions for the label appear in my comments. This is regarding how sub terms for inactive parents have been applied. So either ...
> 
>     1. `[Vocabulary] >> [Term] (inactive)`
>        or ...
> 
>     2. `[Vocabulary] (inactive) >> [Term]`
> 
> 
> We can dismiss this and I will approve either way - I just wanted to point out from my perspective what I was thinking. I'd say team discussion might be good on this one - outlying situation anyway I guess.

not a bad idea, but out of scope for this one. please file a follow-up ticket. i prefer the second form, as it applies the inactive label to the parent term that's actually inactive.

dartajax (commented):
leaving this open until tomorrow to discuss whether my comments about the "inactive" label are applied in this or a possible follow-up PR / issue - not the same as a PR issue

packages/ilios-common/addon/components/taxonomy-manager-terms-list-item.gjs
```diff
@@ -0,0 +1,106 @@
...
+  get level() {
+    return this.args.level ?? 0;
+  }
+
+  get showTooltip() {
+    return this.args?.term.description?.length && this.isHovering;
+  }
+
+  get labelAsInactive() {
+    return this.args.hasActiveParent && !this.args?.term.active;
```
michaelchadwick: Is there a reason why the second check adds an optional to `args` but not the first?

stopfstedt: yes. we don't need it on `hasActiveParent` input since we're checking it for its truthiness directly. whereas on `term.active`, we need to short-circuit if `term` input hasn't resolved to a model object yet.

dartajax:
I truly believe there should be an indicator that either child or parent (or both) has been set to (inactive) before needing to click "Manage Terms" to reveal that information.

image

image

The higher level (before clicking Manage Terms) should read in my opinion `Soundscape (inactive) >> Treble` and then even if "Treble" is also set to inactive, only one (inactive) label (the first time an inactive attribute is discovered) should be displayed. 

Personally as I mentioned, I'd put them both in - less code needed too - if inactive, show it wherever it applies. I don't think I made a good enough pitch on that one. In any case, I do believe at the top level, before clicking Manage Terms, there should be an indicator that the parent or the child or both have effectively rendered this term to be "inactive". Is this possible? I didn't put in a change request in case this is vetoed but I did want to point this out from a user's perspective - I believe it is important to know that a course has been tagged with a term that has been set as inactive (either on its own, or because of its parent, or both).

packages/ilios-common/addon/components/taxonomy-manager-terms-list-item.gjs
stopfstedt: as @michaelchadwick pointed out to me out of band - the optional chaining operator is being applied  at the wrong level.
turns out it's not even necessary to guard against an undefined term here, that's an unrealistic scenario.

i removed the operator and all similar checks from the component since.

dartajax (approved):
this works for me

packages/test-app/tests/integration/components/taxonomy-manager-terms-list-item-test.gjs
michaelchadwick, line 1: All the `assert.step()`s (line 73, 106, and 139) should be moved to the bottom of their blocks like the other instances in the other files in this PR.

michaelchadwick (dismissed):

packages/test-app/tests/integration/components/taxonomy-manager-terms-list-item-test.gjs
stopfstedt: on  it.

stopfstedt: done.

michaelchadwick (approved):
Rad City, Population: you

PR #9389, opened 2026-06-16
remove nav link to curriculum inventory manager from side-menu.

stopfstedt:
the curr inv manager is deprecated, we're phasing it out. removing the nav link is a first step towards this.


jrjohnson (approved):

dartajax (approved):
pretty jarring to see it gone - and many Visual Diff changes of course as a result - but as long as we can reuse the code for reporting like Verification Tables and all that, I approve

PR #9390, opened 2026-06-16
Session Offerings header now stuck to top of viewport when scrolled past

michaelchadwick:
Fixes ilios/ilios#7176


jrjohnson (approved):

jrjohnson:
Possible followup - in a long list of offerings on the same day the day now disappears as you scroll. Maybe it needs to be visible as well?

michaelchadwick:
@jrjohnson That _would_ be cool, but some tinkering on the idea makes it clear that that would be a non-trivial change. Also, the "day now disappearing as you scroll" issue is present in the app pre-PR, as well. I'd like to keep this PR just for changing the header.

stopfstedt (approved):
LGTM

dartajax (approved):
looks good to me

PR #9391, opened 2026-06-16
Offering shows year next to Day of Week and Month if offerings span multiple years

michaelchadwick:
Fixes ilios/ilios#7175

If a session's offerings span multiple years, the year will be shown. Proper `this.intl.formatDate` being used now so it should look good in all translations.

### DIFF
--- a/packages/frontend/tests/acceptance/course/session/offerings-test.js
+++ b/packages/frontend/tests/acceptance/course/session/offerings-test.js
@@ -2,6 +2,7 @@ import { DateTime } from 'luxon';
 import { module, test } from 'qunit';
 import { setupAuthentication, freezeDateAt, unfreezeDate } from 'ilios-common';
 import { setupApplicationTest, takeScreenshot } from 'frontend/tests/helpers';
+import currentAcademicYear from 'ilios-common/utils/current-academic-year';
 import page from 'ilios-common/page-objects/session';
 
 module('Acceptance | Session - Offerings', function (hooks) {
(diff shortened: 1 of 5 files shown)


jrjohnson:
I'm not sure I buy that this should be on every offering, maybe only when the calendar year is crossed by a set of offerings?

dartajax:
> I'm not sure I buy that this should be on every offering, maybe only when the calendar year is crossed by a set of offerings?

This ticket was more of a result of me being confused when a course spans several years itself. I was having a hard time figuring out the dates of offerings since the year was not included so I would say rather when the calendar year is crossed by a set of offerings, maybe if a course spans more than one calendar year start to finish, we could show the year - or of course just show the year with a comma to celebrate its arrival such as "July 1, 2026" rather than "July 1 2026"

dartajax:
I am not seeing the downside of providing the year, myself - but of course, I entered the ticket.

dartajax (changes requested):
`July 1, 2026`
rather than 
`July 1 2026` - put the comma in - my $.02

michaelchadwick:
Change so it only shows the year when there are actual offerings that span more than a year, not every offering.

dartajax (approved):
While it is actually a very rare situation where the two offerings in the same session span into different years, I tested this PR and it accomplishes exactly that - I was looking at this course https://deploy-preview-9391--ilios-frontend.netlify.app/courses/2346 which spans three years and I was getting confused which year was which when I was reviewing the sessions - it is true you can see the year in the listing and / or sort by year but I still found it confusing so I entered this ticket. I am approving this since it does what it has been set out to do, but it's not really what I was proposing from the get-go - I just wanted to see the year along with the date basically always to provide more information to the user not finding that it would make the page overly busy - I guess that was just me. And then I was thinking oh yeah if a COURSE spans more than one year, the year could appear - that would mean that in a course the nature of our friend 2346, all offerings would have the year appended since the COURSE spanned more than one year. However I guess that would affect a lot of courses that run November 2025 to February 2026 for example and that might have been confusing too. As far as this PR is actually designed for situations where offerings in a session span more than one year, this PR does that so - works for me.

stopfstedt (approved):

dartajax (approved):

PR #9397, opened 2026-06-22
indicate inactive ancestors on selected taxonomy terms.

stopfstedt:
currently, only the selected term itself is indicated as such if it is inactive. this adds the indicator to any ancestors of the selected term if applicable.

like so:
image

refs https://github.com/ilios/frontend/pull/9387#issuecomment-4772038331


stopfstedt:
blended into #9387

PR #9400, opened 2026-06-24
Course Publication workflow improvement

michaelchadwick:
Fixes ilios/ilios#6848

This simplifies the Course Publication Menu by removing the 'Review X Items' option and having that feature as part of the "Publish As-is". You now publish the course at the `/publicationcheck` route.

- [x] Remove "Review # items" option in Course Publication button
- [x] Publish remains the same (no missing items)
- [x] Publish As-Is now incorporates the old "Review # items" if there are missing items and requires a "Yes, publish, despite these missing items"
- [x] Expand the Course, remove collapse button

Potential followup items:
- [ ] Missing items should be links to edit them instead of just text (in-page anchors? double-up function of on-page buttons?)
- [ ] Move the Publish As-Is route to above the Course overview section (only one outlet currently and it's either above or below everything)
- [ ] Loosening menu option visibility logic so that "Publish As-Is" shows even if required element (cohort) is not present

### DIFF
--- a/packages/frontend/app/components/courses/list-item.gjs
+++ b/packages/frontend/app/components/courses/list-item.gjs
@@ -9,7 +9,7 @@ import LoadingSpinner from 'ilios-common/components/loading-spinner';
 import PublicationStatus from 'ilios-common/components/publication-status';
 import t from 'ember-intl/helpers/t';
 import { on } from '@ember/modifier';
-import { fn } from '@ember/helper';
+import { fn, hash } from '@ember/helper';
 import FaIcon from '@fortawesome/ember-fontawesome/components/fa-icon';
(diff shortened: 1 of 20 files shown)


michaelchadwick:
@dartajax I'm not sure what "Missing items should be links to edit them instead of just text" means functionally, as all the missing items (terms, cohorts, and objectives) are already on the page with their own Add/Manage buttons, so at best these links would just be in-page anchors, and I'm not sure how helpful that is.

dartajax:
image

michaelchadwick:
@dartajax That does seem like a bug, but it's out-of-band for this PR, as it deals with Session publication, not Course publication. Please ticket it separately.

stopfstedt (approved):
LGTM

dartajax (approved):
After some initial confusion, I approve this.

PR #9404, opened 2026-06-26
Better vertically align Calendar active filters

michaelchadwick:
Fixes ilios/ilios#7223

### DIFF
--- a/packages/ilios-common/app/styles/ilios-common/components/dashboard/courses-calendar-filter.scss
+++ b/packages/ilios-common/app/styles/ilios-common/components/dashboard/courses-calendar-filter.scss
@@ -9,7 +9,12 @@
       @include m.ilios-heading-h6;
     }
 
-    button {
+    /*
+     * kludge-o-rama: fixes bug where
+     * 'Clear Filters' button lost styling
+     * [MC 2026/06/29]
+    */
+    button:not(.filters-clear-filters) {
       @include m.ilios-button-reset;
     }
 
--- a/packages/ilios-common/app/styles/ilios-common/components/dashboard/calendar.scss
+++ b/packages/ilios-common/app/styles/ilios-common/components/dashboard/calendar.scss
@@ -107,6 +107,8 @@
 
     .filters-list {
       clear: both;
+      display: flex;
+      flex-direction: column;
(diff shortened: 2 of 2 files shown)


dartajax:
While it appear to be lined up a bit better the font is no longer blue or light blue and has increased in size slightly. Is this what we wanted? I will put my review in next week but wanted to point that out.

michaelchadwick:
@dartajax Uh, the preview is not how it looks on my local. That is strange. It should look like this:
Screenshot 2026-06-29 at 7 57 43 AM

After checking demo, it also looks like you describe. Hmmm...
Screenshot 2026-06-29 at 7 58 01 AM

packages/ilios-common/app/styles/ilios-common/components/dashboard/courses-calendar-filter.scss
```diff
@@ -9,7 +9,12 @@
       @include m.ilios-heading-h6;
     }
 
-    button {
+    /*
+     * kludge-o-rama: fixes bug where
```
jrjohnson, line 13: This is what I was getting at at code jam about rushing a bit. It's worth it to take the time to figure out what's going on here and style everything as expected. Instead of committing the kludge permanently.

jrjohnson (changes requested):

packages/ilios-common/app/styles/ilios-common/components/dashboard/courses-calendar-filter.scss
michaelchadwick: Fair enough, but for the record: this would permanently commit a `kludge-o-rama`, not a `kludge`. ;-)

michaelchadwick: After team discussion, as ugh as this is, it may be the only way right now, due to SASS build order bugginess.

jrjohnson (approved):

dartajax (approved):

dartajax:
I just checked in with the "Overview" hotel - Visual Diff - looks fine. I tested it as well. We can push this one through.

PR #9405, opened 2026-06-26
Make Curriculum Reports dominant report type and improve course selection/report run flow

michaelchadwick:
Fixes ilios/ilios#7220

- [x] Make Curriculum default reports type route
- [x] Force Curriculum report type selection first

### DIFF
--- a/packages/frontend/app/components/reports/curriculum/header.gjs
+++ b/packages/frontend/app/components/reports/curriculum/header.gjs
@@ -4,7 +4,7 @@ import { service } from '@ember/service';
 import { guidFor } from '@ember/object/internals';
 import t from 'ember-intl/helpers/t';
 import { on } from '@ember/modifier';
-import { eq, not, and, or } from 'ember-truth-helpers';
+import { eq, not, and } from 'ember-truth-helpers';
 import isEmpty from 'ember-truth-helpers/helpers/is-empty';
 import sortBy from 'ilios-common/helpers/sort-by';
 import CopyButton from 'ilios-common/components/copy-button';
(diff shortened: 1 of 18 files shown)


packages/frontend/app/components/reports/curriculum/header.gjs
```diff
@@ -249,7 +223,7 @@ export default class ReportsCurriculumHeaderComponent extends Component {
             type="button"
             class="done text"
             {{on "click" @runReport}}
-            disabled={{not @countSelectedCourses}}
+            disabled={{or (not @countSelectedCourses) (not this.selectedReport)}}
```
stopfstedt: let's flip  that around to make it readable to humans:

```suggestion
            disabled={{not (and @countSelectedCourses this.selectedReport)}}
```

you will have to drop the import of `or` from `ember-truth-helpers` in that component afterwards.

stopfstedt (changes requested):
LGTM.

requesting one minor change to conditional in order to make it more readable.

stopfstedt (approved):
LGTM

dartajax (dismissed):

dartajax:
I approved this but I have a question before merging this one. When you then change the report type after selecting one, the summary data does not refresh with the update and there is no "run" button to update the data. This makes it impossible to know which one was selected previously and what the high level data represents.

michaelchadwick:
@dartajax I think I know what you mean. Once the report is run, the user should not be able to change the report type, and that's a regression bug. Will fix.

dartajax (approved):

dartajax:
There may be a few formatting and caption tweaks I may suggest down the road but we can go ahead with this in my opinion.

PR #9407, opened 2026-06-29
Normalize lock/unlock icon width in Course/Program Year lists

michaelchadwick:
Fixes ilios/ilios#7231

### DIFF
--- a/packages/frontend/app/components/courses/list-item.gjs
+++ b/packages/frontend/app/components/courses/list-item.gjs
@@ -69,7 +69,7 @@ export default class CoursesListItemComponent extends Component {
       <td class="text-center hide-from-small-screen" colspan="2" data-test-end-date>
         {{formatDate @course.endDate day="2-digit" month="2-digit" year="numeric"}}
       </td>
-      <td class="text-right" colspan="3" data-test-status>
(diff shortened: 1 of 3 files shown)


dartajax (approved):
works for me

packages/frontend/app/components/courses/list-item.gjs
jrjohnson: Instead add fixedWith to all the icons here. This will ensure they take up the same space.
```
<FaIcon @icon={{faLock}} @fixedWidth=true />
```

This aligns us better for fontawesome upgrades in the future where the icon widths may change and cause issues with the `min-width`. In FA v7 this option becomes the default, so it makes sense to apply it here now.

jrjohnson (changes requested):

packages/frontend/app/components/courses/list-item.gjs
michaelchadwick: I couldn't find that option in the docs, which is why I did the `min-width` thing. Will fix.

michaelchadwick:
Somehow missed the obvious `@fixedWidth` option on an `<FaIcon>` instance, which is clearly better, so switched that in.

dartajax (approved):

jrjohnson (approved):

PR #9410, opened 2026-07-01
Remove all Overview headers

michaelchadwick:
As it was discussed recently that having the word "Overview" at the top of certain routes was unnecessary, this PR removes all instances of "Overview" across the board where it was found. The translation was removed, as well. Some minor spacing reductions were also applied to make up for the missing text.


stopfstedt (approved):
code changes LGTM. i defer to Dave for final say on the visual/layout changes.

dartajax:
I will check this momentarily.

dartajax (approved):
I may have a few follow-up "space saver" requests on this but then again ... Course and Session both feel a bit empty where "Overview" used to be but Program / Program Year are straight up space saving goodness. All in all, I approve.

dartajax:
Visual Diff approved also - hundreds of differences to be expected with a removal of this type.

PR #9413, opened 2026-07-07
Make Loading Less Random in Tests

jrjohnson:
We'd already locked down the number of rows, but sometimes the random characters wrap onto a new line making these screenshots unstable.


stopfstedt (approved):
"ilios rocks,ilios rocks,ilios rocks,ilios rocks"

PR #9415, opened 2026-07-13
Use new devEngines Property

jrjohnson:
Declare our pnpm version using [devEngines](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#devengines) instead of the now-deprecated corepack packageManager property.


stopfstedt (approved):
nifty. LGTM.

PR #9417, opened 2026-07-13
Session Publication workflow improvement

michaelchadwick:
Fixes https://github.com/ilios/ilios/issues/6850
Follows [ilios/ilios#9400](https://github.com/ilios/frontend/pull/9400)

This simplifies the Session Publication Menu by removing the 'Review X Items' option and having that feature as part of the "Publish As-is". You now publish the session at the `/publicationcheck` route.

- [x] Remove "Review # items" option in Session Publication button
- [x] Publish remains the same (no missing items)
- [x] Publish As-Is now incorporates the old "Review # items" if there are missing items and requires a "Yes, publish, despite these missing items"

Potential followup items:

- [ ] Missing items should be links to edit them instead of just text (in-page anchors? double-up function of on-page buttons?)
- [ ] Loosening menu option visibility logic so that "Publish As-Is" shows even if required element (offering) is not present

### DIFF
--- a/packages/ilios-common/addon/components/session-publicationcheck.gjs
+++ b/packages/ilios-common/addon/components/session-publicationcheck.gjs
@@ -1,9 +1,12 @@
 import Component from '@glimmer/component';
+import { service } from '@ember/service';
 import { TrackedAsyncData } from 'ember-async-data';
 import { cached } from '@glimmer/tracking';
+import { action } from '@ember/object';
 import Overview from 'ilios-common/components/session/overview';
 import { LinkTo } from '@ember/routing';
 import { array, hash } from '@ember/helper';
+import { on } from '@ember/modifier';
 import FaIcon from '@fortawesome/ember-fontawesome/components/fa-icon';
 import t from 'ember-intl/helpers/t';
 import scrollIntoView from 'ilios-common/modifiers/scroll-into-view';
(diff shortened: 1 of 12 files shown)


dartajax:
Caption should be "Publish Session" on the drop-down rather than "Publish As-is" but functionally this seems fine.

dartajax (approved):
seems fine to me but a follow-up ticket related to the same captioning as at the course level will be forthcoming in all likelihood

packages/ilios-common/addon/components/session-publicationcheck.gjs
```diff
@@ -55,6 +62,14 @@ export default class SessionPublicationCheckComponent extends Component {
     return objectivesWithoutParents.length > 0;
   }
 
+  @action
+  async publish() {
+    this.args.session.set('publishedAsTbd', false);
+    this.args.session.set('published', true);
+    await this.args.session.save();
+    this.router.transitionTo('session', this.args.session);
+    this.flashMessages.success(this.intl.t('general.publishedSuccessfully'));
```
jrjohnson, line 71: This should go before the transition. There is some chance, if the transition is fast, that the component will get torn down and this service won't exist anymore.

jrjohnson (changes requested):

jrjohnson (approved):

PR #9418, opened 2026-07-14
Log user out if api request returns 401

michaelchadwick:
Fixes ilios/ilios#7113

This should fix the issue of someone's JWT being invalid.

To test:
1) Log into Ilios normally
2) Open up Developer Console and go to the `ilios-session` cookie.
3) Edit it by changing the `jwt` value to something invalid (e.g. add a `LOL` right in the middle of it)
4) Refresh page

Any API requests (appConfig, schools) will be sent, return a 401, which will invalidate their session and shuttle the user to the `login` route.

### DIFF
--- a/packages/frontend/app/routes/application.js
+++ b/packages/frontend/app/routes/application.js
@@ -12,6 +12,7 @@ export default class AuthenticatedRoute extends Route {
   @service store;
   @service router;
   @service session;
+  @service iliosConfig;
   @service localStorage;
 
   @tracked event;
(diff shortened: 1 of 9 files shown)


jrjohnson (changes requested):
Works perfect!
1. Probably needs a flash message "You've been logged out because XX"
2. Should catch and do the same thing in the fetch and graphQL services

jrjohnson (approved):

dartajax (approved):

PR #9421, opened 2026-07-14
Use pnpm to Update Itself

jrjohnson:
In pnpm v11 the [self-update](https://pnpm.io/cli/self-update) command was added and we can now use it instead of fancy tricks to update our version. This will also update the lockfile appropriately.


stopfstedt (approved):
rad.

PR #9422, opened 2026-07-16
Don't show any message if no global search query

michaelchadwick:
Fixes ilios/ilios#6601

### DIFF
--- a/packages/frontend/app/components/global-search.gjs
+++ b/packages/frontend/app/components/global-search.gjs
@@ -144,13 +144,15 @@ export default class GlobalSearchComponent extends Component {
             {{t "general.currentlySearchingPrompt"}}
           </li>
         {{else}}
-          {{#each this.results as |course|}}
-            <CourseSearchResult @course={{course}} />
-          {{else}}
-            <li class="no-results">
-              {{t "general.noSearchResultsPrompt"}}
-            </li>
-          {{/each}}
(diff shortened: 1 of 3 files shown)


packages/frontend/app/components/global-search.gjs
```diff
@@ -144,13 +144,17 @@ export default class GlobalSearchComponent extends Component {
...
+              <li class="no-results">
+                {{t "general.noSearchResultsPrompt"}}
+              </li>
+            {{/each}}
           {{else}}
-            <li class="no-results">
-              {{t "general.noSearchResultsPrompt"}}
-            </li>
-          {{/each}}
+            <li class="no-query"></li>
```
jrjohnson: Seems to only be used in tests so should be `data-test-no-query` instead of adding a css class. Some of the others should probably be data-test as well, but that's for probably never followup.

jrjohnson (changes requested):
I'm a little worried bout the accessibility of an empty list. Wondering if it would work to wrap the entire `<ul>` in this check instead?

jrjohnson (approved):
LGTM, but I do think it would benefit from the addition of a component visual test in this state, but won't require it.

michaelchadwick:
Added a visual screenshot test!

jrjohnson (approved):

PR #9429, opened 2026-07-21
Normalize event session descriptions

michaelchadwick:
Fixes ilios/ilios#6656

After much work trying to get `TruncateText`'s "character count" algo to work more like `FadeText`'s "vertical space" algo, I gave up as I was essentially turning one component into the other for no reason.

In the end, just putting `TruncateText` on session descriptions when shown on the WaaG and/or Single Event routes feels like enough. If it's overly tall, it still won't truncate unless the character count is long enough, which feels like an edge cases most of the time.

This also squares up the presentation of session descriptions across WaaG, Single Event, Session, and editing Session, as they were not normalized (mainly in how they displayed lists).

Here's an example of various description "lengths" on the dashboard and how they'd now be handled.
Screenshot 2026-08-06 at 1 15 48 PM

The preview, logged in as dev, also currently has a course with a session and a wacky description to check all this again.

### DIFF
--- a/packages/ilios-common/addon/components/single-event.gjs
+++ b/packages/ilios-common/addon/components/single-event.gjs
@@ -8,6 +8,7 @@ import { findById } from 'ilios-common/utils/array-helpers';
 import createTypedLearningMaterialProxy from 'ilios-common/utils/create-typed-learning-material-proxy';
 import { TrackedAsyncData } from 'ember-async-data';
 import FaIcon from '@fortawesome/ember-fontawesome/components/fa-icon';
+import TruncateText from 'ilios-common/components/truncate-text';
 import t from 'ember-intl/helpers/t';
 import { and, eq, gt, not } from 'ember-truth-helpers';
 import { LinkTo } from '@ember/routing';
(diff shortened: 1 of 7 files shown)


michaelchadwick:
NTS: Try combining TruncateText but with the FadeText vertical checking.

dartajax:
image

image

stopfstedt (approved):

dartajax:
image

dartajax:
I know it's on me to merge this PR but I'm not quite feeling it ... yet.

dartajax:
from PR build ...

image

from Netlify ...

image

michaelchadwick:
@jrjohnson mentioned a magic number issue in one of the stylesheets, so I changed it to a value that is more consistent with other values we use. WaaG feels like an outlier on this one, which is why it gets the extra attribute. I tried other places, but this was the easiest way to get everything to line up. I'll let him be the judge.

michaelchadwick:
@dartajax The screenshot differences are because of the style adjustments I made. I _think_ they are all right, but I'd rather you be the judge :D

dartajax:
I'm not sure what to do on this one. Extra space (vertical space) is added to the WaaG view when it switches days - That would be okay if that is what's needed. 

But, Event Detail doesn't look as good either - is this all needed because of standardization? I will merge and approve if so. 

From the PR ...

image

From Netlify ...

image

Please let me know if this is all intended and if so, we can get this through.

michaelchadwick:
Quill, the plugin behind Ilios's editable textareas, uses a formula to figure out how much indentation for lists within lists with lists, etc. Should the standard be how its editor handles this, and the rest of the rules build off of it, or the other way around? Regardless, what should that standard be?

stopfstedt (approved):
code looks good, was happy to see one less usage of a SCSS mixin while reviewing.

dartajax (approved):
looks fine to me

PR #9431, opened 2026-07-22
Consolidate Fetching 

jrjohnson:
We have several calls to fetch around and about. Most of which also call on the session for auth data and the config to get the path correct. I've refactored them all to run through the fetch service. Which will make changes to the session easier to control.

### DIFF
--- a/packages/frontend/app/components/program-year/objective-list.gjs
+++ b/packages/frontend/app/components/program-year/objective-list.gjs
@@ -26,7 +26,7 @@ import {
 
 export default class ProgramYearObjectiveListComponent extends Component {
   @service iliosConfig;
-  @service session;
+  @service fetch;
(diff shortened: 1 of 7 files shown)


michaelchadwick:
ax72ne

michaelchadwick (changes requested):
This is all great! However, it seems like the `IliosAdapter` should get the same treatment: https://github.com/ilios/frontend/blob/master/packages/ilios-common/addon/adapters/ilios.js#L15-L25

jrjohnson:
Since it didn't use fetch directly I skipped it here, but my next PR with a tokenStorage service will replace all the places we create authHeaders with a single place on the token store.

michaelchadwick (approved):
👍

PR #9433, opened 2026-07-23
Refactor Auth to Login Form

jrjohnson:
Removing this from the authenticator makes everything there simpler and now the login is next to the form.


michaelchadwick (approved):
Looks all right to myself

PR #9434, opened 2026-07-23
Choose our Session Store Dynamically

jrjohnson:
We keep the cookie based session store for our regular frontend, but also allow LTI auth to use a memory only store that doesn't persist anywhere. This allows users to use the LTI as we have been, but not cause their authentication to carry over into the frontend if they visit it directly.


stopfstedt (approved):
looks good.

PR #9435, opened 2026-07-24
Global Search Page query clearing no longer affects results

michaelchadwick:
Fixes ilios/ilios#7268

Instead of doing a "blank" search if the user clears their query on the search page, use `window.history.replaceState()` instead.

### DIFF
--- a/packages/frontend/app/components/global-search-box.gjs
+++ b/packages/frontend/app/components/global-search-box.gjs
@@ -49,7 +49,9 @@ export default class GlobalSearchBoxComponent extends Component {
     this.internalQuery = q;
     if (this.router.currentRouteName === 'search') {
       if (q === '') {
-        this.args.search('');
+        const url = new URL(window.location.href);
+        url.searchParams.set('q', '');
+        window.history.replaceState(null, '', url.pathname + '?' + url.searchParams.toString());
       }
     }
   });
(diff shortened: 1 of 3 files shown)


stopfstedt (commented):
looks good to me, but i'm not loving the solution - this is the first instance i can find where we're manipulating the history object ourselves.
but can't really think of a better one myself atm.
so i'll solicit a second opinion from @jrjohnson on this.

jrjohnson (changes requested):
I think the original clearing of the search was needed when we had autocomplete and a "keep typing" prompt that are both gone now. Instead of messing with the history I think this entire block can just get deleted. I can't see that it's doing anything today.

packages/frontend/app/components/global-search-box.gjs
```diff
@@ -47,11 +47,6 @@ export default class GlobalSearchBoxComponent extends Component {
 
   setInternalQuery = task({ restartable: true }, async (q) => {
```
jrjohnson: Sorry to pile on, but this `task` is now too much. It's not doing anything tasky anymore. Could either be a `set` in the template block or else an action here. Either will remove the need to import perform or task.

jrjohnson (changes requested):

michaelchadwick:
@jrjohnson Removed your old code comment...for reasons ;P https://github.com/ilios/frontend/commit/2426e7d4ea0fb091c97df91b18f4cee026db3534#diff-5dcc15f896aa18580cd661284da505325dfc916cbf71cbe49cf4617bdf196a9eR54-R64

jrjohnson (approved):

jrjohnson:
This component is so much simpler now, love it.

dartajax (approved):

PR #9436, opened 2026-07-24
Include Official Email in Results

jrjohnson:
When searching for a user we check this field on the API so we should display it in the results so it is clear you found the person you were looking for.

Refs, but doesn't require ilios/ilios#7280 as a missing value is ok here.


stopfstedt (approved):
LGTM

dartajax (approved):

PR #9438, opened 2026-07-28
Add Hover State to All Buttons

michaelchadwick:
Fixes ilios/ilios#6915

- [x] `<button>`
- [x] `<a>` acting as buttons
- [x] ToggleButtons
- [x] YesNo Buttons
- [x] BigAdd/BigCancel Buttons
- [x] `<button>` around `<svg>`
- [x] `data-test-actions` with icons
- [x] work with light and dark mode


stopfstedt (commented):
click tested for a bit, the only "button" without a noticeable hover state that i found was this one on the courses list. (it's not really a button but it looks like one).
i'd be ok with dealing with this one in a follow-up.

Image

the choice of colors tracks for me.

stopfstedt (approved):

michaelchadwick:
@stopfstedt That's a ToggleButton component, right? They have a hover state now, but only on the inactive button. It seemed unnecessary to add one for the active button since clicking it doesn't do anything, since it's...already active. The hover state indicates something that can be clicked to change a state.

stopfstedt:
> @stopfstedt That's a ToggleButton component, right? They have a hover state now, but only on the inactive button. It seemed unnecessary to add one for the active button since clicking it doesn't do anything, since it's...already active. The hover state indicates something that can be clicked to change a state.

i looked into the[ web accessibility](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#hover-states) aspect of this, and a viz change on hover is not required here. 

> The pointer itself, via its location, is the indicator of whether the user is hovering on a component. Therefore, additional author-supplied visual treatments for hover are not "required to identify" the hover state. Those treatments can be considered supplemental and do not themselves need to contrast 3:1 against the background.

carry on.

michaelchadwick:
Please have a look and comment on the color choices, thank you.

jrjohnson (changes requested):
Lots of visual diff that isn't related to hover state. Probably needs a rebase. But it's possible color changes have bled into unrelated areas and need adjustment.

jrjohnson (approved):
I did my best to look through the changes. Visually awesome, huge improvement.

dartajax (approved):
heavy hitters have approved this as well - let's get this over the finish line

PR #9439, opened 2026-07-29
Fix `class` attribute conditional values that can leave trailing spaces (or worse!)

michaelchadwick:
Noticing several places where an element will have a conditional CSS class applied, but if the condition isn't met then the attribute will have a trailing space, which, while not breaking, looks wrong in the rendered template.

For example, `class="foo bar {{if this.hasBoop 'has-boop'}}`. If we got `hasBoop`, all fine, but if not, then we get the incorrect HTML:

```
<element class="foo bar ">Stuff Stuff</element>
```

Worse yet, if there are multiple conditional classes, and one or all are not applied, you might get something like the following:

For example, `class="{{ if this.hasGoop 'has-goop' }} {{ if this.hasFloomp 'has-floomp' }}"`. If there is neither goop nor floomp, then you get something like the following glop:

```
<element class="
     ">Cool Text</a>
```

I've fixed some in past PRs, but this PR is a concerted effort to fix all the remaining instances (generally with a simple space swap-a-roo, but sometimes more was needed), making rendered HTML look slick and intentional. Devs will appreciate it and curious View Source-rs, if they still exist, will appreciate it, too.

### DIFF
--- a/packages/frontend/app/components/curriculum-inventory/sequence-block-overview.gjs
+++ b/packages/frontend/app/components/curriculum-inventory/sequence-block-overview.gjs
@@ -1099,7 +1099,7 @@ export default class CurriculumInventorySequenceBlockOverviewComponent extends C
               {{/if}}
             </div>
             <div
-              class="block is-selective {{unless this.isSelective 'hidden'}}"
+              class="block is-selective{{unless this.isSelective ' hidden'}}"
               data-test-is-selective
             >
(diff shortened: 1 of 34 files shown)


stopfstedt (approved):
legendary.

jrjohnson (approved):
Love it. Next step: add a rule for this to https://github.com/ember-cli/eslint-plugin-ember

dartajax (approved):

PR #9440, opened 2026-07-30
Curriculum Reports Header revamp

michaelchadwick:
Fixes ilios/ilios#7117

Attempt at reducing curriculum report header vertical whitespace, and simplifying template a bit. Switched from action buttons being on their own line to a second element in a flexbox with the report type selector and run summary text. It's not perfect (the summary text is boxed in by the flexbox, so it doesn't flow under the buttons), but trying either a `float` or `position: fixed` situation did not really work well. Open to comments!

Note: this only affects tablet and up. Mobile still just vertically stacks everything like before.

Old
oldcur1
oldcur2
oldcur3

New
newcur1
newcur2
newcur3

### DIFF
--- a/packages/frontend/app/components/reports/curriculum/header.gjs
+++ b/packages/frontend/app/components/reports/curriculum/header.gjs
@@ -173,8 +173,11 @@ export default class ReportsCurriculumHeaderComponent extends Component {
     return window.location.href;
   };
   <template>
-    <div class="reports-curriculum-header" data-test-reports-curriculum-header>
-      <div class="type-and-summary">
+    <div
+      class="reports-curriculum-header{{if @showReportResults ' results'}}"
+      data-test-reports-curriculum-header
+    >
+      <div class="selection">
         {{#unless @showReportResults}}
           <label data-test-report-selector>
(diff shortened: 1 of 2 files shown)


michaelchadwick:
Screenshots don't totally match up final version, which now uses space a little more intelligently. Also had to use `float` for the buttons at a certain viewport on the "reports are being displayed" route to get the text to wrap around them. Regardless, I think this is an improvement.

jrjohnson:
@michaelchadwick Not sure if your comment indicates screenshot indicate more work is needed here? Or this is ready, but screenshots need approval?  Go ahead and mark them as approved if this is ready.

michaelchadwick:
@jrjohnson I can see how my comment was not clear. I meant that the original screenshots I posted in the main PR do not quite match up with what I ended up with. I just checked the actual visdiff screenshots and all the differences make sense, so I approved them.

jrjohnson:
lol, secret option number three!

jrjohnson (approved):
Looks way better!

dartajax (approved):
looks like an improvement to me

PR #9441, opened 2026-07-31
fix broken lock file.

stopfstedt:


stopfstedt:
operator error. i was still using pnpm v10. works fine with v11, which is the required version anyway.

i'm retracting the PR.

PR #9442, opened 2026-07-31
Preserve scroll position on LG->Cohort Members

michaelchadwick:
Fixes ilios/ilios#7305

Lots of time toying with complicated modifiers and scroll event listeners, but finally got a simple one to do the trick that uses the `scrollend` event so that it only triggers when scrolling stops, not on every `scroll` event. It only works with _modern_ ([Dec 2025 or newer](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollend_event)) versions of browsers, but will just be ignored otherwise.

Note: there already exists a `PreserveScroll` service, but that works when transitioning between routes, and this PR adds an element modifier that can remember scroll position within a route.

### DIFF
--- a/packages/frontend/app/components/learner-group/cohort-user-manager.gjs
+++ b/packages/frontend/app/components/learner-group/cohort-user-manager.gjs
@@ -18,6 +18,7 @@ import UserStatus from 'ilios-common/components/user-status';
 import LoadingSpinner from 'ilios-common/components/loading-spinner';
 import perform from 'ember-concurrency/helpers/perform';
 import { faPlus } from '@fortawesome/free-solid-svg-icons';
+import preserveScroll from 'ilios-common/modifiers/preserve-scroll';
 
 export default class LearnerGroupCohortUserManagerComponent extends Component {
   @service currentUser;
@@ -25,6 +26,8 @@ export default class LearnerGroupCohortUserManagerComponent extends Component {
   @tracked filter = '';
   @tracked selectedUsers = [];
   @tracked usersBeingMoved = [];
+
+  cohortScroll = 0;
 
   get sortedAscending() {
     return this.args.sortBy.search(/desc/) === -1;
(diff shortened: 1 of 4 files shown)


michaelchadwick:
NTS: Might be able to fix with different way of tracking the `<tbody>` vs capturing scroll position.

jrjohnson:
I think the answer is in `packages/frontend/app/components/learner-group/root.gjs`

Somewhere in...
```
  @cached
  get usersForCohortManagerData() {
    // Learnergroup members are only referenced here to trigger a re-computation on membership changes.
    return new TrackedAsyncData(
      this.getUsersToPassToCohortManager(this.args.learnerGroup, this.args.learnerGroup.users),
    );
  }

  get usersForCohortManager() {
    return this.usersForCohortManagerData.isResolved ? this.usersForCohortManagerData.value : [];
  }

  async getUsersToPassToCohortManager(learnerGroup) {
    const cohort = await learnerGroup.cohort;
    const topLevelGroup = await learnerGroup.getTopLevelGroup();
    const currentUsers = await topLevelGroup.getAllDescendantUsers();
    const users = await cohort.users;
    return users.filter((user) => !currentUsers.includes(user));
  }
```
...there is a place we could cache this data so we didn't return the `[]` empty during the rebuild. 

My suspicion is that it's in:
```
const topLevelGroup = await learnerGroup.getTopLevelGroup();
    const currentUsers = await topLevelGroup.getAllDescendantUsers();
```
...but might be somewhere else.

What should happen when a user is moved is that the array decreases by one person and, that will trigger a re-render in the shadow dom, and that one row will get removed.

What is actually happening is that `@users.length` in `cohort-user-manager.gjs` is getting set to `[]` for just a moment. This causes the entire list to get destroyed and recreated.

This is worth tracking down, the scroll position tracking feels like a bandaid to me.

michaelchadwick:
@jrjohnson I tried creating `get`ters for each of the things that use `await` in that function (because each one sets the list to (0) temporarily, like you said), but I can't figure out `const currentUsers = await topLevelGroup.getAllDescendantUsers();` as all my attempts have an additional `await` inside them and don't get rendered on load somehow. I'll have to give it another go later, unless you have a suggestion.

jrjohnson:
It needs work, but https://github.com/ilios/frontend/pull/9540 is what I'm looking at.

michaelchadwick:
#9540 has superseded this PR, so closing.

PR #9446, opened 2026-08-04
lint Glimmer components and templates againts class-spacing formatting mistakes.

stopfstedt:
refs https://github.com/ilios/ilios/issues/7304

this works as intended, those lists of class names on components get now linted for redundant white-space.

unfortunately, it also clashes with the ember plugin for eslint. ~~i have yet to figure out what the actual problem is, and how to resolve/evade it.~~ see comment below.

here's an example of the warnings emitted. 

```
packages/frontend lint:js:   0:0  warning 
packages/frontend lint:js: To lint Gjs/Gts files please follow the setup guide at https://github.com/ember-cli/eslint-plugin-ember#gtsgjs
packages/frontend lint:js: Note that this error can also happen if you have multiple versions of eslint-plugin-ember in your node_modules
packages/frontend lint:js: /home/stefan/dev/projects/frontend/packages/frontend/app/components/learner-group/user-manager.gjs
```

### DIFF
--- a/packages/frontend/eslint.config.mjs
+++ b/packages/frontend/eslint.config.mjs
@@ -19,7 +19,7 @@ import ember from 'eslint-plugin-ember/recommended';
 import eslintConfigPrettier from 'eslint-config-prettier';
 import qunit from 'eslint-plugin-qunit';
 import n from 'eslint-plugin-n';
-
+import html from '@html-eslint/eslint-plugin';
 import babelParser from '@babel/eslint-parser';
 
 const esmParserOptions = {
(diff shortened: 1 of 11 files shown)


stopfstedt:
based on the [docs](https://eslint.org/blog/2022/08/new-config-system-part-2/#goodbye-extends%2C-hello-flat-cascade), i don't think it's possible to do it this way.

> This config has two config objects with overlapping files patterns. The first config object applies to all .js and .cjs files while the second applies only to .js files. When linting a file ending with .js, ESLint combines both config objects to create the final config for the file. Because the second config sets semi to a severity of "warn", that takes precedence over the "error" that was set in the first config. **The last matching config always wins when there is a conflict.** 

in other words, we'll always be clobbering the `parser`/`parserOptions` if there's a conflict.

stopfstedt:
won't fix.

PR #9448, opened 2026-08-04
Add test helpers to augment Publication Menu integration tests when GUI window not active

michaelchadwick:
Fixes ilios/ilios#7043

Needed Claude to finally figure out how to fix this annoying issue (at least to me) when running Course/Session Publication Menu integration GUI tests. If the window/tab is not active, then navigating the dropdown menu and getting the correctly focused text does not work. These new test helpers are smarter about it and work even when the window is not active.


michaelchadwick:
NTS: Fix newly-broken tests, and also make sure the tests don't always pass (when code is written that should break them).

jrjohnson (approved):

PR #9452, opened 2026-08-05
remove redundant double-await.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/7324

```bash
grep -rl 'await await' packages --exclude-dir=node_modules --exclude-dir=dist | xargs sed -i 's/await await/await/g'
```


jrjohnson (approved):
What a day!

dartajax (approved):

PR #9455, opened 2026-08-05
Save Groups in Order

jrjohnson:
To match our new validation [requirements on the API](https://github.com/ilios/ilios/pull/7290) we need to add and remove groups in specific order instead of making the changes and saving everything at the same time. This will ensure that the group hierarchy tree is respected at each point.


jrjohnson:
This is ready for review, but it should be [tested against the API it is linked to](https://github.com/ilios/ilios/pull/7290) before merging. I put a Do Not Merge on it for that reason. Once it's been validated by someone else against that API PR that can be removed.

stopfstedt (approved):
code reviewed and click-tested against a backend on the companion PR.
lgtm.

PR #9460, opened 2026-08-10
replaces calls to RSVP.map() with native equivalents.

stopfstedt:
fixes ilios/ilios#7325

### DIFF
--- a/packages/ilios-common/addon/routes/course-visualize-term.js
+++ b/packages/ilios-common/addon/routes/course-visualize-term.js
@@ -1,6 +1,5 @@
 import { service } from '@ember/service';
 import Route from '@ember/routing/route';
-import { map } from 'rsvp';
 
 export default class CourseVisualizeTermRoute extends Route {
   @service store;
@@ -17,8 +16,8 @@ export default class CourseVisualizeTermRoute extends Route {
     const sessions = await course.sessions;
     return await Promise.all([
       term.vocabulary,
-      map(sessions, (s) => s.sessionType),
-      map(sessions, (s) => s.totalSumDuration),
+      ...sessions.map((s) => s.sessionType),
+      ...sessions.map((s) => s.totalSumDuration),
     ]);
   }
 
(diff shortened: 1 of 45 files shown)


packages/ilios-common/addon/routes/course-visualize-term.js
```diff
@@ -15,11 +14,9 @@ export default class CourseVisualizeTermRoute extends Route {
...
   async afterModel({ course, term }) {
     const sessions = await course.sessions;
-    return await Promise.all([
-      term.vocabulary,
-      map(sessions, (s) => s.sessionType),
-      map(sessions, (s) => s.totalSumDuration),
-    ]);
+    await term.vocabulary;
+    await Promise.all(sessions.map((s) => s.sessionType));
+    return await Promise.all(sessions.map((s) => s.totalSumDuration));
```
jrjohnson: This has slightly different loading behavior in that it would now send each network request individually, waiting for the response. This is repeated in a few of the routes here, a better patten would be something more like:

```suggestion
  //this is freehanded, not comitable, just an example
    return Promise.all([
      term.vocabulary,
      ...sessions.map((s) => s.sessionType),
      ...sessions.map((s) => s.totalSumDuration),
    ]);
```

that will fire all the requests at the same time and only move on with rendering when they've all resolved.

jrjohnson (dismissed):
Would like a slightly different pattern in the routes where we're doing pre-loading to fire all network requests at the same time (as well as resolution in different threads instead of waiting for each individually).

packages/ilios-common/addon/routes/course-visualize-term.js
stopfstedt: yea makes sense. i'll give the destructuring approach a try. thanks.

stopfstedt: handled. :white_check_mark:

jrjohnson (approved):

dartajax (approved):

PR #9463, opened 2026-08-10
Course->Publish missing details button bugfix

michaelchadwick:
Fixes ilios/ilios#7340

Fixes bug where after publishing a course the "Hide/Show Details" button was hidden.

### DIFF
--- a/packages/frontend/app/components/courses/list-item.gjs
+++ b/packages/frontend/app/components/courses/list-item.gjs
@@ -9,7 +9,7 @@ import LoadingSpinner from 'ilios-common/components/loading-spinner';
 import PublicationStatus from 'ilios-common/components/publication-status';
 import t from 'ember-intl/helpers/t';
 import { on } from '@ember/modifier';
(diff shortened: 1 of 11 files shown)


jrjohnson (changes requested):
Does `detailsCollapseControl` need to be a query parameter at all? Wouldn't it be safer to make this a `get showCollapseControl` on the controller or root component and then decide based on the current route?

michaelchadwick:
> Does `detailsCollapseControl` need to be a query parameter at all? Wouldn't it be safer to make this a `get showCollapseControl` on the controller or root component and then decide based on the current route?

On second thought, it does not! Refactored to make it a controller method instead, and it's much cleaner. Thank you!

jrjohnson (approved):

dartajax (approved):
As far as the hiding of the "Hide Details" button is concerned, this fix seems good to me. "Show Details" is not actually applicable in this particular situation since the screen is already set to show details even if hide details was the state the screen was in before doing the publication. That would be a different issue anyway.

PR #9464, opened 2026-08-10
Don't display "invalid auth" flash message on login page

michaelchadwick:
Fixes ilios/ilios#7344

### DIFF
--- a/packages/frontend/app/components/login-form.gjs
+++ b/packages/frontend/app/components/login-form.gjs
@@ -10,13 +10,14 @@ import t from 'ember-intl/helpers/t';
 import { on } from '@ember/modifier';
 import pick from 'ilios-common/helpers/pick';
 import set from 'ember-set-helper/helpers/set';
+import { waitForFetch } from '@ember/test-waiters';
 import YupValidationMessage from 'ilios-common/components/yup-validation-message';
 import perform from 'ember-concurrency/helpers/perform';
 import LoadingSpinner from 'ilios-common/components/loading-spinner';
 
 export default class LoginFormComponent extends Component {
   @service session;
-  @service fetch;
+  @service iliosConfig;
   @tracked error;
   @tracked username;
   @tracked password;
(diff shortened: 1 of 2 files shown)


jrjohnson (changes requested):
It's more verbose, but I'd be happier not using the `fetch` service on the form auth part of the login route than doing this one off string comparison in the fetch service. I don't like the design of the service needing to know about the route, feels backwards.

stopfstedt (dismissed):

michaelchadwick:
@jrjohnson I'm not sure exactly how to remove the `fetch` service from login in the way you're suggesting.

jrjohnson:
Instead of calling `this.fetch.getJsonFromApiHost(loginUrl);` in `packages/frontend/app/routes/login.js` call `fetch` directly with something like 

```
const trimmedPath = relativePath.replace(/^\//, '');
    const path = `${this.host}/${trimmedPath}`;
    const response = await waitForFetch(fetch(path, options));

    // if invalid auth, invalidate session
    if (response.status == 401) {
      this.flashMessages.alert(this.intl.t('errors.invalidAuthentication'));
      this.session.invalidate();
    }
```

and then handle the response. It's a little more verbose in that one place (because you don't get ll the nice path stuff), but it's the only place this will come up.

packages/frontend/app/components/login-form.gjs
```diff
@@ -60,16 +67,25 @@ export default class LoginFormComponent extends Component {
   }
 
   async loginWithCredentials(data) {
-    const response = await this.fetch.fetchFromApiHost('/auth/login', {
+    const loginUrl = `/auth/login`;
+    const trimmedPath = loginUrl.replace(/^\//, '');
```
jrjohnson: This code from the fetch service isn't needed here as `loginUrl` is a const that doesn't have a trailing slash. This is needed in the service to protect against bad input. Probably even clearer to just to `const path = ${this.host}/auth/login/`

jrjohnson: Or without the slash. Whichever way actually works. Not sure what the value of `this.host` is and if it has a trailing slash.

jrjohnson (changes requested):

jrjohnson (approved):

dartajax (approved):

PR #9467, opened 2026-08-12
Welcome to the Dark(er) Side

jrjohnson:
To support a better reading experience and allow users to match Ilios to their overall theme we're adding a dark mode ilios experience.

Adding the `color-scheme` option indicates that we support both a light
and dark color scheme.

I've tended towards a slightly softer dark mode than absolute white on
absolute black. We were able to re-use most of our existing colors so we
still get excellent balance and relationships between the colors in use.

Fixes ilios/ilios#7300

Wip:
- [x] I've applied a global --PINK to unfinished areas to make them easier to see, remove all instances
- [x] Fix any a11y issues
- [x] Check generate screenshots
- [x] Lots and lots of feedback

### DIFF
--- a/packages/ilios-common/app/styles/ilios-common/components/html-editor.scss
+++ b/packages/ilios-common/app/styles/ilios-common/components/html-editor.scss
@@ -54,7 +54,7 @@
 
   button {
     background-color: transparent;
-    color: var(--blue);
+    color: light-dark(var(--blue), var(--bright-blue));
     display: flex;
     justify-content: flex-end;
     margin-left: auto;
@@ -62,7 +62,7 @@
     width: auto;
 
     &:hover {
-      background-color: var(--lightest-grey);
+      background-color: light-dark(var(--lightest-grey), var(--dark-grey));
     }
   }
 }
@@ -71,9 +71,17 @@
 .ql-snow .ql-toolbar button {
   padding: 3px 5px !important;
 
+  .ql-stroke,
+  .ql-fill {
+    transition: light-dark(stroke 0.5s, stroke 0);
+  }
+
   .ql-stroke {
-    stroke: var(--black);
-    transition: stroke 0.5s;
+    stroke: light-dark(var(--grey), var(--white));
+  }
+
+  .ql-fill {
+    fill: light-dark(var(--grey), var(--white));
   }
 
   &:disabled {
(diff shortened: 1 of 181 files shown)


michaelchadwick:
Did a check on all the site sections and here are my notes:

* Global
    * Magnifying glass is now outside the global search input (Chrome/FF/Safari confirmed)
         * Screenshot 2026-08-12 at 9 57 45 AM
    * Locale/User dropdowns should probably only have a white outline when active/focused (like the global search acts)
    * User Guide link doesn’t need the white outline
    * Version numbers status in bottom-right more readable in dark mode!
    * Checkmark/Xmark when editing single line things should probably inherit the green of the star icon and the red (whatever it becomes) of the trashcan icon
    * Calendar popups can probably use the same coloring for header, and change the calendar part to light on dark
    * QuillJS textarea toolbar icons look bad; remove `fill`
    * Breadcrumbs should use white border and link color should be the same as other blue links
    * New [Model] forms: background color should be just a touch lighter than the page background color
    * Gray user “bubbles” (i.e. chosen users to be instructors) could probably use the same bg/color
    * Hamburger icon can just be white, like in light mode
* Dashboard
    * My ICS Link icon needs dark color mode
* Course
    * Session
        * Publication Review: Yes green in column should be brighter
        * All Session Publication Review: background-color should match the rest of the page, not be pink?
* Learner Groups
    * Manage Group Members: why background color of headers pink? It should be the same color as the rest of the page
* Program->Program Years
    * Details/overview should just use page background-color, not pink
* Reports
    * Switcher should use the same color scheme as ToggleButton/MultiButton
* Visualizations
    * Tooltips should invert their colors

packages/ilios-common/app/styles/ilios-common/components/html-editor.scss
```diff
@@ -71,8 +71,10 @@
 .ql-snow .ql-toolbar button {
   padding: 3px 5px !important;
 
-  .ql-stroke {
-    stroke: var(--black);
+  .ql-stroke,
+  .ql-fill {
+    stroke: light-dark(var(--black), var(--white));
+    fill: light-dark(var(--black), var(--dark-grey));
```
michaelchadwick: These styles don't quite work, but the following should work for both light and dark mode, and hovering on each:
```
.ql-stroke,
.ql-fill {
  transition: light-dark(stroke 0.5s, stroke 0);
}

.ql-stroke {
  stroke: light-dark(var(--editor-icon-black), var(--white));
}

.ql-fill {
  stroke: light-dark(var(--editor-icon-black), var(--white));
  fill: light-dark(var(--editor-icon-black), var(--white));
}
```

Also, `colors.scss` needs this: `--editor-icon-black: hsl(0, 0%, 27%);`

jrjohnson: ✅

michaelchadwick (approved):
All the issues I brought up were fixed and this looks pretty darn amazing! Great job.

dartajax (approved):

PR #9468, opened 2026-08-12
Add Darkmode Tests

jrjohnson:
We need the full a11y suite of tests for our darkmode colors and screenshots as well. This is complicated, and might be possible to simplify it later, but for now let's see.

Refs ilios/ilios#7300


jrjohnson:
Visual Diff Tests will fail for now and we don't have dark mode yet so this isn't doing anything, but it's the baseline to get something going and tested. To get good visual diff results we'll need to merge this in with the current failures.

stopfstedt (approved):
i accept this.

PR #9469, opened 2026-08-12
Fix Publish All Sessions bucket and status logic

michaelchadwick:
Fixes ilios/ilios#7348
Fixes ilios/ilios#7349

- [x] Published Sessions now only show sessions already published, regardless of publication issues
- [x] Unpublished Sessions only show unpublished sessions that can be published as-is or scheduled
- [x] Scheduled Sessions now shown in Unpublished Sessions bucket
- [x] Sessions that are ILMs show `n/a` under Offerings column
- [x] PublicationStatus now shown next to each session link, for even easier identification
- [x] `Publish (All) As-is` changed to just `Publish (All)`
- [x] Empty buckets have no toggle
- [x] Add Unpublished Sessions third choice "Leave as unpublished/scheduled" if session is not scheduled
- [x] Default choice should be Publish unless the session already has a status
- [x] Review status at bottom now reflects the new logic (ignore -> leave X as unchanged)
- [x] Action buttons changed from "Go" to "Publish/Schedule/Leave" or "Cancel"
- [x] Positive action button only displays if there is work to do
- [x] Overhaul acceptance/integration tests


dartajax:
initial thoughts ...
1. Maybe as we discussed at our standup, get rid of the "ignore (x)" part of the message since no action is being taken on those sessions. 
2. I hate to bring this up but for the "Unpublished Sessions: for review (x)" section there should be a way for user to decide "nah I'm cool with the current "Not Published" status". Currently user needs to set as scheduled or publish every session or abort the workflow which some users may do if they get confused on this one. We may need a "Take No Action" or "Keep Not Published" option along with a third option at the top if that's possible. I apologize that we hadn't discussed this before.

michaelchadwick:
@dartajax Having a "cancel" button seems like a good idea, and would make this function more consistent with other places where an action needs to be confirmed/canceled.

dartajax:
> @dartajax Having a "cancel" button seems like a good idea, and would make this function more consistent with other places where an action needs to be confirmed/canceled.

That's a good idea. Probably another ticket? That's for overall "Cancel" not individual sessions I assume.

Regarding individual sessions ... 

What if user for example in the ABC 26-27 course which has 28 sessions currently requiring review wants to publish 10 of them, schedule 5 of them, and leave the remaining 13 in not published state? We should have gone over this before and I apologize for that.

packages/ilios-common/addon/components/publish-all-sessions.gjs
```diff
@@ -36,35 +37,60 @@ export default class PublishAllSessionsComponent extends Component {
     return new TrackedAsyncData(this.args.course.courseObjectives);
   }
 
+  get courseObjectives() {
+    return this.courseObjectivesData.isResolved ? this.courseObjectivesData.value : null;
+  }
+
```
michaelchadwick, line 45: This got moved just to keep with the 1. TrackedAsyncData method 2. TrackedAsyncData value convention in most other places

michaelchadwick:
@dartajax I have the same course/sessions on my local as on demo, so let me know if the sessions are in the right bucket, thanks!

dartajax:
Generally, I like this. The only thing I noticed is something I mentioned earlier that I guess would result in a bit of scope creep or follow-up ticket possibly instead. If you do not wish to publish or schedule an offering in the bottom bucket, there is no way to simply leave it as-is in a "not published" state. Workflow-wise I think this might be important and could discourage people from using this functionality. Can we not add a third option? Leave as-is or Take no action - something along those lines. I understand that might make radio buttons no longer usable in this context or maybe we can use check boxes instead so users could simply de-select any action and essentially leave the session "as-is" - "not published".

dartajax:
Another scope creep fantasy thing would be to be able to sort the sessions in the "Unpublished Sessions: for review" bucket by publication status - either not published or scheduled are the two options. If you want to see all of the "Scheduled" offerings at the top of the list you could sort Z-A on status. Follow-up ticket? no problem - just an idea.

dartajax:
As far as the buckets are concerned, and as far as currently scoped out, it works great - accurate buckets.

michaelchadwick:
@dartajax There should be a cancel button now, and I figured that satisfied the "don't do anything" option. As for the other stuff, please ticket if not already, as this PR has gotten a bit out of hand now, and I want to keep scope creep a minimum. Thanks!

dartajax:
We are looking at Aardvark, a course with 9 sessions. Many courses of course have many times that number of sessions. I was thinking we need a session by session way of telling the system not to publish or schedule one or more sessions and leave them as is. I see the Cancel button but that is for the entire operation at a course level not at a session level. I am sorry to bring this up and I am certainly willing to address that as a follow-up ticket since this PR fixes a whole bunch of issues compared to the way it was before.

michaelchadwick:
@dartajax Oh, I see what you mean now. Yes, that would be useful, but please include as a followup.

dartajax:
Final comment then on the topic - and this is the starting point for the follow-up ticket. 

Regarding the final bucket `Unpublished Sessions: for review (x)`, two suggestions

Three options instead of two - controls may have to change or as long as one is selected radio buttons might still be okay. 

* Publish 
* Mark as Scheduled
* Ignore

If user selects "ignore" the "ignore" count would increment by the number of ignored sessions selected in this bucket. 

Or ...

Check boxes instead allowing the user to de-select "Publish" and de-select "Mark as Scheduled" leaving nothing selected - essentially "ignoring" the session selected and incrementing the "ignored" session count on the message at the bottom of the screen when processing these changes.

dartajax:
The problem here is when you set a session to "Scheduled", it will appear on learners' calendars just without the additional information of a published session. This is not something that would be wanted by course coordinators and administrators in many situations. They may want to leave sessions in a not published state until completely ready. That having been said, we could merge this one and put it into production and then later address my concern on this forcing of the hand issue I have been going on about.

dartajax (approved):
I have some ideas for follow-up tickets to be certain but the way this is set up seems to work quite well especially if you really do want to publish all sessions and leave a couple of them in "scheduled" mode.

packages/frontend/tests/acceptance/course/publish-all-sessions-test.js
```diff
@@ -55,79 +55,101 @@ module('Acceptance | Course - Publish All Sessions', function (hooks) {
...
       'Yes (1)',
     );
-    assert.ok(page.publishAllSessions.publishableSessions.sessions[2].objectives.isLinked);
+    assert.ok(page.publishAllSessions.publishedSessions.sessions[2].objectives.isLinked);
   });
 
   test('after publishing user is returned to the courses route #4099', async function (assert) {
     const terms = await this.server.createList('term', 1);
 
-    // const course = await this.server.create('course', {
```
jrjohnson, line 125: 💯

packages/ilios-common/addon/components/publish-all-sessions.gjs
```diff
@@ -17,138 +17,67 @@ import SaveButton from 'ilios-common/components/save-button';
...
-    }
-
-    return Boolean(
-      this.courseObjectives.find((objective) => {
-        return objective.programYearObjectives.length === 0;
-      }),
-    );
-  }
-
+  // sessions with no required, but possible optional, publication issues
```
jrjohnson: These would be a little easier to read as `/**` comment blocks. Not a blocker.

packages/ilios-common/addon/components/publish-all-sessions.gjs
```diff
@@ -222,46 +193,182 @@ export default class PublishAllSessionsComponent extends Component {
...
   get ignoreCount() {
-    return this.unPublishableSessions.length;
+    return this.publishedSessions.length + this.unPublishableSessions.length;
+  }
+
+  get actionCount() {
+    return this.publishCount + this.scheduleCount;
+  }
+
+  get saveButtonText() {
```
jrjohnson, line 301: Up for discussion, but I find this much easier to understand in the template. Here in the JS with all the returns and `intl`s it's tricky to figure out what's going to display.

packages/ilios-common/addon/components/publish-all-sessions.gjs
```diff
@@ -222,46 +193,182 @@ export default class PublishAllSessionsComponent extends Component {
...
-        (s) => s !== session,
-      );
-      this.userSelectedSessionsToSchedule = [...this.userSelectedSessionsToSchedule, session];
-    } else {
-      this.userSelectedSessionsToSchedule = this.userSelectedSessionsToSchedule.filter(
-        (s) => s !== session,
-      );
-      this.userSelectedSessionsToPublish = [...this.userSelectedSessionsToPublish, session];
+  selectSessionAction(session, action) {
+    switch (action) {
```
jrjohnson: Any reason not to break this up into actions for `publish`, `schedule`, and `leave`? The switch makes it pretty complicated.

packages/ilios-common/translations/en-us.yaml
```diff
@@ -148,7 +148,6 @@ general:
   firstOffering: First Offering
   forward: Forward
   friday: Friday
-  go: Go
```
jrjohnson, line 151: Bye bye `go`!

jrjohnson (approved):
I have some changes, but they're all really style not structural. Your call to fix them here of followup. Marking as approved. It's a beast!

(I didn't check behavior, I think @dartajax is all over that, just read the code)

michaelchadwick:
@jrjohnson thanks for looking into this! I'm going to implement your suggestions, so setting this back to draft for a bit more.

packages/ilios-common/addon/components/publish-all-sessions.gjs
michaelchadwick: This is a great idea, and you'd think after 2+ years of working on Ilios (and knowing how this was originally implemented) I would done that, but...I didn't. Updating.

packages/ilios-common/addon/components/publish-all-sessions.gjs
michaelchadwick: Chad Gippety made a case for a simpler `saveButtonText()` method, and I like it, so I went with that instead.

```
get saveButtonText() {
  return [
    this.publishCount && this.intl.t('general.publish'),
    this.scheduleCount && this.intl.t('general.schedule'),
    this.leaveCount && this.intl.t('general.leave'),
  ]
    .filter(Boolean)
    .join('/');
}
```

dartajax (approved):
this is slick

packages/ilios-common/addon/components/publish-all-sessions.gjs
jrjohnson: Pretty!

jrjohnson (approved):
Now we're cooking with steam!

PR #9470, opened 2026-08-13
Align session search and actions with loaded state

james-chua-1493:
Fixes #7351

## Problem
The loading skeleton shown while a course page is fetching data (`course/loading.gjs`) didn't match the layout of the loaded component (`course/sessions.gjs`). Specifically:
- The session search input rendered as a full-width block below the `course-sessions-header`, instead of inline within it
- The header's `.actions` container was empty during loading, with no placeholder for the "New Session" and "Publication Review" buttons

This caused a visible layout shift once the real data finished loading and the search box snapped into its correct position.

## Fix
- Moved the `.filter` (search input) block so it's nested inside `.actions`, matching the structure of `course/sessions.gjs`
- Added two disabled placeholder buttons inside `.actions` for "New Session" and "Publication Review", using the same translation keys as the real buttons, so the loading state's header height and layout match the loaded state

## Testing
- Ran the full test suite (`ember exam`) — 1774 tests, 0 failures
- Ran `lint:hbs` and `lint:js` on the changed file — no issues
- Manually verified the markup structure matches `course/sessions.gjs` by comparing both files directly

## Note
My local pre-commit hook (Husky) flagged pre-existing Prettier formatting issues across ~233 files in unrelated packages, not caused by this change. I bypassed the hook (`--no-verify`) for this commit since my modified file passes lint cleanly on its own.

### DIFF
--- a/packages/ilios-common/addon/components/course/loading.gjs
+++ b/packages/ilios-common/addon/components/course/loading.gjs
@@ -3,6 +3,8 @@ import animateLoading from 'ilios-common/modifiers/animate-loading';
 import t from 'ember-intl/helpers/t';
 import FaIcon from '@fortawesome/ember-fontawesome/components/fa-icon';
 import { faSquarePlus } from '@fortawesome/free-solid-svg-icons';
+import ExpandCollapseButton from 'ilios-common/components/expand-collapse-button';
+const noop = () => {};
 <template>
   <BackToCourses />
 
@@ -59,10 +61,13 @@ import { faSquarePlus } from '@fortawesome/free-solid-svg-icons';
             {{! template-lint-disable require-input-label }}
             <input disabled />
           </div>
-          <button type="button" disabled>
-            {{t "general.newSession"}}
-          </button>
-          <button type="button" disabled>
+          <ExpandCollapseButton
+            @value={{false}}
+            @action={{noop}}
+            @expandButtonLabel={{t "general.newSession"}}
+            @collapseButtonLabel={{t "general.close"}}
+          />
+          <button type="button" class="loading-text" disabled>
             {{t "general.publicationReview"}}
           </button>
         </div>
--- a/packages/ilios-common/addon/components/course/loading.gjs
+++ b/packages/ilios-common/addon/components/course/loading.gjs
@@ -2,9 +2,7 @@ import BackToCourses from 'ilios-common/components/course/back-to-courses';
 import animateLoading from 'ilios-common/modifiers/animate-loading';
 import t from 'ember-intl/helpers/t';
 import FaIcon from '@fortawesome/ember-fontawesome/components/fa-icon';
-import { faSquarePlus } from '@fortawesome/free-solid-svg-icons';
-import ExpandCollapseButton from 'ilios-common/components/expand-collapse-button';
-import noop from 'ilios-common/helpers/noop';
+import { faSquarePlus, faPlus } from '@fortawesome/free-solid-svg-icons';
 <template>
   <BackToCourses />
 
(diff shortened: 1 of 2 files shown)


packages/ilios-common/addon/components/course/loading.gjs
```diff
@@ -54,11 +54,18 @@ import { faSquarePlus } from '@fortawesome/free-solid-svg-icons';
...
-      <div class="filter">
-        {{! template-lint-disable require-input-label }}
-        <input disabled />
+        <div class="actions">
+          <div class="filter">
+            {{! template-lint-disable require-input-label }}
+            <input disabled />
+          </div>
+          <button type="button" disabled>
+            {{t "general.newSession"}}
```
jrjohnson: When finished loading this is a [plus sign](https://github.com/ilios/frontend/blob/master/packages/ilios-common/addon/components/course/sessions.gjs#L183). Please adjust the loading template to match here.

jrjohnson (changes requested):
Thanks for contributing! 
Need to adjust the new session button and add some style as the buttons are too clearly visible when loading, take a look at https://github.com/ilios/frontend/blob/master/packages/frontend/app/styles/components/courses/loading.scss for some inspiration.

packages/ilios-common/addon/components/course/loading.gjs
james-chua-1493: I have updated the template to use `<ExpandCollapseButton>` so it now renders the plus icon instead of text together with the muted styling. Hope this helps, happy to make any more changes if required.

packages/ilios-common/addon/components/course/loading.gjs
```diff
@@ -3,6 +3,8 @@ import animateLoading from 'ilios-common/modifiers/animate-loading';
 import t from 'ember-intl/helpers/t';
 import FaIcon from '@fortawesome/ember-fontawesome/components/fa-icon';
 import { faSquarePlus } from '@fortawesome/free-solid-svg-icons';
+import ExpandCollapseButton from 'ilios-common/components/expand-collapse-button';
+const noop = () => {};
```
jrjohnson:
```suggestion
import noop from 'ilios-common/helpers/noop';
```

jrjohnson (changes requested):
Closer!

The buttons need to be much more washed out so they don't feel clickable.

This is an example of a select list that is more what I'm wanting here. It's by adding `class="loading-text"`, use that instead of `color`, but make the background color `var(--grey)` as well for those buttons.

Image

There is also an alignment issue with the plus sign, not sure where that is coming from, it appears to be slightly too short.

Image

james-chua-1493:
Both the plus icon and "Publication Review" now use the shared loading-text class together with background-color: var(--grey).

Applied a flex layout to .course-sessions-header which mimics the loaded state styling from the detail-container-header mixin.

james-chua-1493:
Also — is there a recommended local dev setup doc you could refer me to for visually comparing frontend changes. I followed the CONTRIBUTING.md instructions and got the app running, but I'm stuck at the login page. Would love to get this working smoothly since I'm hoping to pick up more frontend issues going forward!

jrjohnson:
> Also — is there a recommended local dev setup doc you could refer me to

Great question @james-chua-1493, I thought we had this all sorted, but I needed to make an update. You can find this info in [CONTRIBUTING.md](https://github.com/ilios/frontend/blob/master/CONTRIBUTING.md). It's not super simple because you'll need to email us and get some credentials, but that's the best option for doing local development.

You can also open up the Netlify preview here and see your changes live, but it requires the same username / password combination.

Nothing special required, just shoot us an email and we'll get you that ASAP.


It does look like there is something wrong with calling `noop` like this - I think that is the cause of the tests failures, you can chase that down, but I think it might be easier just to create a plain button in the loader with a matching icon and then grey it out.

james-chua-1493:
Latest commit swaps out the <ExpandCollapseButton> for a plain button with a matching icon. 

Also sent an email requesting the demo credentials. Thanks for the feedback!

packages/ilios-common/app/styles/ilios-common/components/course/loading.scss
```diff
@@ -20,9 +20,21 @@
...
     @include m.for-phone-only {
       height: 3rem;
     }
+
+    .actions {
+      display: flex;
+      cursor: default;
+
+      .loading-text {
+        background-color: var(--grey);
```
jrjohnson:
```suggestion
        background-color: var(--lighter-grey);
```

jrjohnson (changes requested):
Structure looks great, just a little bit dark on the grey. Let's try this (or play around with something else now that you (hopefully) have access.

james-chua-1493:
I've got the local setup running now and verified this visually — var(--lighter-grey) lines up better with the rest of the app's muted styling. Applied, thanks!

jrjohnson (approved):
Really nice!

PR #9475, opened 2026-08-20
Add Persistent Storage for User Preferences

jrjohnson:
Add Persistent Storage for User Preferences

Using the new /application/preferences API we can now save user preferences to the database. A couple of design decisions worth calling out:

- Preferences are stored as semiPrivate properties to get tracking working, tracking on private properties doesn't work.
- Setting a preference uses the setter, which is async, attempting to set the preference directly results in an error.
- I versioned the preference data, we can upgrade it during setup if we determine a different structure is better later.
- Implemented the move of local preferences to here, I needed something to use for testing
- Moved localStorage over into preferences

Refs ilios/ilios#7362


jrjohnson:
Moving this back to drat as the getters aren't getting auto tracked because we're using an object.

jrjohnson:
My attempts to limit this change failed, I ended up needing to take on the migration of localStorage and the storing of locale preferences in this PR to get everything lined up correctly and tested. Best reviewed as individual commits.

michaelchadwick (approved):
Nice to have all this a way that works for both db and localStorage options. Woot!

dartajax (approved):
looks cool to me

PR #9477, opened 2026-08-21
Fix Calendar Event divide by zero bug

michaelchadwick:
sliver

Fixes ilios/ilios#3590

Calendar events can sometimes display as narrow slivers even when there aren't enough events at the same time to compress them. This feels like a bug, and this PR will fix that...eventually.

This PR fixes the divide by zero issue that would cause a span of `Infinity`.

Future PR will address the "two events happening at the same time, but taking up same space and causing everything else to sliver" bug


jrjohnson (approved):

jrjohnson:
Causes some really interesting visual diff changes. Looks like we may have a reproduction of something else in `Acceptance _ Dashboard Calendar_load day calendar` because both the baseline and candidate are wrong.

dartajax (approved):

PR #9484, opened 2026-08-25
Use focus Modifier on Login Form

jrjohnson:
Might be more accessible, but I'm primarily trying to resolve a race condition in our screenshots.


stopfstedt (approved):
fine by me.

PR #9485, opened 2026-08-25
ILM Session Publish shows "ILM" for offerings if none, otherwise shows count as normal

michaelchadwick:
Fixes ilios/ilios#7350

### DIFF
--- a/packages/ilios-common/addon/components/session-publicationcheck.gjs
+++ b/packages/ilios-common/addon/components/session-publicationcheck.gjs
@@ -124,15 +124,21 @@ export default class SessionPublicationCheckComponent extends Component {
                 <td data-test-session-title>
                   {{@session.title}}
                 </td>
-                {{#if @session.offerings.length}}
-                  <td class="yes" data-test-offerings>
-                    {{t "general.yes"}}
-                    ({{@session.offerings.length}})
+                {{#if @session.isIndependentLearning}}
+                  <td data-test-offerings>
+                    {{t "general.notApplicableAbbr"}}
                   </td>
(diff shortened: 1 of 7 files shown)


dartajax (approved):
I like this and approve this PR.

packages/ilios-common/addon/components/session-publicationcheck.gjs
```diff
@@ -124,15 +124,21 @@ export default class SessionPublicationCheckComponent extends Component {
                 <td data-test-session-title>
                   {{@session.title}}
                 </td>
-                {{#if @session.offerings.length}}
-                  <td class="yes" data-test-offerings>
-                    {{t "general.yes"}}
-                    ({{@session.offerings.length}})
+                {{#if @session.isIndependentLearning}}
```
stopfstedt: close, but not quite.

this fails when we have an ILM with offerings attached.

you'll need to add a check for offerings length to this conditional.

```suggestion
                {{#if (and @session.isIndependentLearning (not @session.offerings.length))}}
```

btw, we're doing this wrong in the same way on the bulk session publication screen as well.

stopfstedt (changes requested):
LGTM overall, but needs some work around the edges.

michaelchadwick:
ILMs can have offerings, so if they do, show a count. If they don't have any, then just put "ILM" instead of "No".

stopfstedt (approved):
LGTM.

dartajax (approved):
new changes seem fine too even though I had approved this PR in its previous state

PR #9488, opened 2026-08-26
Set Params to Defaults

jrjohnson:
Instead of importing this from Ember we can just set these to false, which is their default. This export is no longer provided and using it is SUPER deprecated.


michaelchadwick (approved):
/me blows dust off code, coughs

dartajax (approved):

PR #9489, opened 2026-08-27
Update Snapdom

jrjohnson:
Bumping up to the latest version to hopefully stabilize our screenshots, fixed an import of the wrong method in a component as the new snapdom fails when it doesn't get a real element.


jrjohnson:
Doesn't help. Or breaks in a different way. Not right.

PR #9491, opened 2026-08-27
Fix ICS Feed icon color when in dark mode

michaelchadwick:
Fixes ilios/ilios#7372

Changed ICS Feed icon color in dark mode to match darker orange header color. Added white background so that details of icon are white (they aren't explicitly colored, but are just transparent).

### DIFF
--- a/packages/ilios-common/app/styles/ilios-common/components/dashboard/navigation.scss
+++ b/packages/ilios-common/app/styles/ilios-common/components/dashboard/navigation.scss
@@ -23,15 +23,18 @@
       }
 
       .ilios-calendar-ics-feed {
-        padding: 0;
+        background-color: light-dark(transparent, var(--white));
         text-align: center;
         width: 90%;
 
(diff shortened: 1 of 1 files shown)


jrjohnson (changes requested):
Instead of guessing the dimensions I think you can use https://docs.fontawesome.com/web/style/icon-canvas#automatic-width or one of the other FA styling options to lock the icon in s better spot and then just make the button width/height 100%.

michaelchadwick:
@jrjohnson I tried a bunch of things, but it doesn't seem to work as well as the exact dimensions, even though avoiding that would be ideal. Open to you giving it a go :D

jrjohnson (approved):
I accept this

dartajax (approved):
for what it's worth I didn't have a problem with the way it was - it was yelling a bit I guess - now it matches the header color and that is what we discussed so I am going to approve and merge this beaut

PR #9493, opened 2026-08-28
Migrate Locale Chooser Into User Profile

jrjohnson:
Moved the locale chooser into a box on the profile page and rebuilt the header to take advantage of the new space. I think there may be more we can do with the header, but this works for now.


saschaben:
French translation update: `Choisissez votre langue préférée pour la présentation d’Ilios. Notez: cela ne modifie pas les Matériels d'étude ni le contenu sous-jacents.`

packages/frontend/translations/en-us.yaml
```diff
@@ -170,6 +170,8 @@ general:
     en-us: English (en)
     es: Español (es)
     fr: Français (fr)
+  languages: Languages
+  languagePickerExplanation: "Choose your prefered lanuage for Ilios presentation. Note: this doesn't modify the underlying materials and content."
```
michaelchadwick: "lanuage" -> "language"

dartajax (approved):
Spanish translation looks fine to me and I'm generally in favor of where this is headed. Looks good. Screen adjust slightly to the right a bit more when French or Spanish is selected. Maybe 'cuz the night mode text is longer in those languages? Just a guess.

dartajax:
> Spanish translation looks fine to me and I'm generally in favor of where this is headed. Looks good. Screen adjust slightly to the right a bit more when French or Spanish is selected. Maybe 'cuz the night mode text is longer in those languages? Just a guess.

This occurs because the other left menus adjust for longer French and Spanish translation strings. Students only would never notice this anyway since they wouldn't have those side menus.

michaelchadwick (approved):
Muy excelente, amigo

PR #9494, opened 2026-08-28
Remove Visual Diff Approval for New Code

jrjohnson:
When a PR gets updated this task will run to remove the `approve visual diff` label. This will then require a re-check of screenshots if there are any problems and avoid merging (or automerging) the PR if there are new issues that pop up.

I had to use the `pull_request_target` type here to support forks so this won't run until after it is merged.

For reference: https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request

### DIFF
--- a/.github/workflows/remove-labels.yml
+++ b/.github/workflows/remove-labels.yml
@@ -1,0 +1,21 @@
+name: Reset Visual Diff Approval
+
+on:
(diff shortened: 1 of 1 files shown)


stopfstedt (approved):
LGTM.

PR #9496, opened 2026-08-30
Fix empty My Materials search state

lKolabrodl:
Fixes ilios/ilios#7377.

## Problem

Filtering My Materials down to zero rows left the sortable table headers visible and displayed the generic `None` message.

## User impact

An empty search looked like a populated table with missing data, rather than a clear no-results state.

## Root cause

The table header was rendered whenever the user had any materials before filtering. It did not consider the filtered result count.

## Fix

- Hide the table header when the filtered total is zero.
- Reuse the localized `noResultsFound` message for an empty text search.
- Add an integration test covering the zero-result state.

## Verification

- `ember test --filter 'Integration | Component | dashboard/materials'` — 11 passed.
- ESLint passed for both changed files.
- Ember template lint passed for both changed files.
- Prettier and `git diff --check` passed.

AI assistance was used to help investigate and prepare this change; the final diff and test results were reviewed manually.

### DIFF
--- a/packages/frontend/tests/acceptance/dashboard/materials-test.js
+++ b/packages/frontend/tests/acceptance/dashboard/materials-test.js
@@ -510,10 +510,12 @@ module('Acceptance | Dashboard Materials', function (hooks) {
       };
     });
     await page.visit({ course: 4 });
-    assert.strictEqual(
-      page.materials.topPaginator.controls.pagerDetails.text,
-      'Showing 1 - 0 of 0',
-    );
+    assert.dom('[data-test-dashboard-materials] table').doesNotExist();
+    assert.dom('[data-test-dashboard-materials] [data-test-paginator-top]').doesNotExist();
+    assert.dom('[data-test-dashboard-materials] [data-test-paginator-bottom]').doesNotExist();
+    assert
+      .dom('[data-test-dashboard-materials] [data-test-none]')
+      .hasText('No results found. Please try again.');
     assert.strictEqual(page.materials.courseFilter.options.length, 6);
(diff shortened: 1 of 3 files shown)


jrjohnson (changes requested):
Thanks for contributing! Hiding the header doesn't match our other implementations of this feature. Instead an if block should be used to decide to paint the table at all and if not display `No results found`.

lKolabrodl:
Updated in 013f5348a. The materials table is now rendered only when `this.total` is non-zero; otherwise the component displays the localized `No results found` message outside the table. The integration test now verifies that the table does not exist for an empty filtered result.

Validated locally: 11/11 dashboard/materials integration tests pass, and ESLint plus ember-template-lint pass for both changed files.

jrjohnson (changes requested):
Needs one more change that wasn't in the original issue, the pagination should be hidden as well when there are no materials to display.

Image

lKolabrodl:
Fixed in b5071ba. Both the top and bottom pagination controls now use the same `this.total` condition as the table, so they are not rendered when there are no results. Clearing the search restores the table and pagination.

Updated the regression tests for empty results and course filters. All 21 relevant tests (11 integration and 10 acceptance) and the full `pnpm lint` pass locally.

jrjohnson:
@dartajax can you take a look at this? The orange line needs to go, but I also wonder if we should really use two translations for this? Maybe no results is right for all situations? Your call.

jrjohnson (changes requested):
Tests should use the page object pattern found in the other tests.

Waiting on some visual feedback as well.

PR #9505, opened 2026-08-31
control visibility of MeSH UI based on school configuration

stopfstedt:
refs ilios/ilios#3549

i added visibility switches for MeSH UI components in the following areas
- program-year/course/session objective management
- course/session learning materials management
- course/session MeSH management
- print course

the school config check happens at top level components, then the value gets passed down into the relevant sub-components.


jrjohnson (approved):
Sweet!
Doing the tests with a data provider also gets a good screenshot for both states!

I added https://github.com/ilios/ilios/issues/7414 to follow up on a better DX for accessing the config values.

dartajax (approved):

PR #9506, opened 2026-09-01
Profile Updates

jrjohnson:
Some cleanup on MyProfile.

Fixes ilios/ilios#3667 ilios/ilios#7385

I elected not to put an expand control on learner groups as it's now at the bottom of the page and not in the way of anything else. Happy to revisit if that's still needed.


dartajax (approved):
I like this.

michaelchadwick (approved):
Sweet. Nice job making the token gen stuff into its own component.

PR #9514, opened 2026-09-02
Keep + button for new things aligned right regardless of viewport

michaelchadwick:
Fixes ilios/ilios#7402

This fixes the reported issue in Program/Program Years, but the style is shared, so it affects Courses, LGs, IGs, and Schools, too.


stopfstedt (approved):
LGTM

dartajax (approved):
this is great

PR #9516, opened 2026-09-03
compile zod objects.

stopfstedt:
refs https://github.com/ilios/ilios/issues/7413

doesn't seem to do much to speed up performance, yet. but here it is anyway.


jrjohnson:
As you indicated, doesn't seem to impact our test speed at all. I think we should skip this complexity.

PR #9519, opened 2026-09-06
Consolidate School Config Work

jrjohnson:
We had a few different paths to getting at the configuration for a school and a lot of async work to get there. Instead I've moved everything into a schoolConfig service that is loaded at startup and the provides synchronous access to the config for each school.

Fixes ilios/ilios#7414


stopfstedt (commented):
whatever the LLM spat out here (i assume this was vibed), i'm not convinced. this does not look good to me.

utililty methods that remove some code redundancy around loading and accessing these school config values would have sufficed, in my opinion.

would like to discuss this a bit more when you're back.

jrjohnson:
Yes, let's discuss. I think my desire to limit this changes may have made my choices here a bit confusing. We already had a path to access school config by name and limit this redundancy using the school model, but it wasn't used in most places. What I'm trying to do here is two big things:
1) To improve the interface for accessing these config values by passing them through a method. This allows us to standardize how we would provide a default value and removes the necessity of knowing exactly the right string to access a value by making access pass though a getter avoiding things like capitalization issues with `MeSH` vs `Mesh`.
2) To eliminate the need to treat these values (which almost never change) as async. This simplifies the components a lot.

The next step is to either pass the config value or the school into these components, which will remove a lot more data loading and asynchronous work from several places. I didn't want to do that here as it's much more change and this gets the job done as a start.


Also no, I did generate some of the test cases, but none of the application code. Curious what gives it that feel.

jrjohnson:
Discussion:
1. Move the saving back out of the service, put it back in the school tree and have the components call it as an action
2. Move forward with consolidating lookups and passing the values down, don't wait for another PR
3. Get rid of tracked, we're not doing anything reactive with these values

stopfstedt:
Hereby dismissing my paranoid review from earlier this week.

Instead of making accusations, I should have waited until we talk. Further, I should have read the description more thoroughly, this would have avoided another misunderstanding on my part.

@jrjohnson - I was wrong, I'm sorry that I came at you like this, and I apologize for that.

jrjohnson:
@stopfstedt this is ready for review again. The only thing from our discussion I didn't implement was moving the saving out of the service. I think it belongs there as there is some complexity saving config values that new ones may need to be created, plus doing this in the school root component just doesn't feel like the right place. I think it belongs in the service where config is managed.

packages/ilios-common/addon-test-support/msw/setup.js
```diff
@@ -47,7 +47,7 @@ export function setupMSW(hooks) {
   });
 }
 
-function get(url, callback) {
+function get(url, callback, once = true) {
```
stopfstedt, line 50: ❤️

packages/test-app/tests/integration/services/school-config-test.js
```diff
@@ -0,0 +1,379 @@
...
+
+    await this.server.create('school-config', {
+      school,
+      name: 'showMeSH',
+      value: 'true',
+    });
+
+    await this.schoolConfig.load();
+
+    assert.true(this.schoolConfig.getShowMeSH(null));
```
stopfstedt, line 110: that's some really goofy logic here - get true on a non-existing school. that being said, i accept this, since this will only ever manifest itself in tests like this.

packages/test-app/tests/integration/services/school-config-test.js
```diff
@@ -0,0 +1,379 @@
...
+    assert.strictEqual(config.name, 'showSessionSupplemental');
+    assert.false(config.parsedValue);
+
+    await this.schoolConfig.save();
+
+    assert.false(this.schoolConfig.getShowSessionSupplemental(school.id));
+    assert.verifySteps(['POST called']);
+  });
+
+  test.each(
```
stopfstedt, line 169: data providers! nice.

packages/test-app/tests/integration/services/school-config-test.js
```diff
@@ -0,0 +1,379 @@
...
+      const school = await this.server.create('school');
+      const existingConfig = await this.server.create('school-config', {
+        school,
+        name: configName,
+        value: initialValue,
+      });
+
+      await this.schoolConfig.load();
+
+      const upperCase = configName[0].toUpperCase() + configName.slice(1);
```
stopfstedt, line 272: would have used https://api.emberjs.com/ember/3.3/classes/String/#capitalize here, but that's a nitpick and not a blocker.

stopfstedt (approved):
LGTM.

packages/test-app/tests/integration/services/school-config-test.js
jrjohnson: Damn! I knew there was a way to do this, I kept hitting up the JS MDN docs. I was so sure I didn't have to make it this annoying. I'll fix this in an followup for sure. It was bugging me. Thanks!

stopfstedt: nevermind, we're not extending string anymore. ecmascript 5555 will give us a native `capitalize` function, no doubt. :|

stopfstedt: found it after all.
`import { capitalize } from '@ember/string'`.

see our template helper for reference. https://github.com/ilios/frontend/blob/da51e658e5394758fe26fc4594000704079f4e62/packages/ilios-common/addon/helpers/capitalize.js#L2

dartajax (approved):

PR #9522, opened 2026-09-08
Fix LG bulk upload table header size

michaelchadwick:
Fixes ilios/ilios#6867


stopfstedt (approved):
looks better/good enough to me.

dartajax (approved):

PR #9524, opened 2026-09-08
Fix Single Event expand/collapse elements' `aria-label`s so they match text

michaelchadwick:
Fixes ilios/ilios#7389


stopfstedt (dismissed):
this misses the point. instead, please ditch the aria-label attribute altogether. aria-collapsed provides enough context for screen readers.

without the label attr, screen readers will output something like 

`Materials 3, button, expanded` and `Materials 3, button, collapsed`. that's on point.

stopfstedt (approved):
LGTM.

PR #9536, opened 2026-09-10
Objective Sorting style improvement

michaelchadwick:
Fixes ilios/ilios#7388

* Spacing between action buttons and [+] button is improved across course/session/program-year objectives
* Increased sortable button width so they maximize usable space
* Sortable buttons now have hover state

**Edit [by ST]**

Since the differ doesn't capture this screen, here is a side-by-side. Improvements are on the left.

image

**Edit [by MC]**

Added a screenshot to Program Year->Objectives to hopefully catch this change in the future. Also, once this goes in, the `.bigadd, .bigcancel` issue can be addressed: https://github.com/ilios/ilios/issues/7415

### DIFF
--- a/packages/frontend/tests/integration/components/program-year/objective-list-test.gjs
+++ b/packages/frontend/tests/integration/components/program-year/objective-list-test.gjs
@@ -1,5 +1,5 @@
 import { module, test } from 'qunit';
-import { setupRenderingTest } from 'frontend/tests/helpers';
+import { setupRenderingTest, takeScreenshot } from 'frontend/tests/helpers';
 import { render } from '@ember/test-helpers';
 import { component } from 'frontend/tests/pages/components/program-year/objective-list';
 import a11yAudit from 'ember-a11y-testing/test-support/audit';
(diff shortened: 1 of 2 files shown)


stopfstedt (approved):
looks very good to me.

dartajax:
get a little closer - don't be shy ...

image

... oh yeah - sorry

Are the "plus" buttons and the "undo" buttons the exact same size? The icon itself for "undo" is probably ever so slightly wider. This is better for sure and I'm sorry to be such a pain.

dartajax:
or put those buttons on the left where the "Sort Objectives" button was and now has changed into having the two buttons instead

dartajax (changes requested):
see my comments for the change request - if you put the buttons back on the left where the "Sort Objectives" button used to be, the minuscule spacing irregularities I pointed out would no longer apply

dartajax:
my review screen shot may be used for future tickets - same size buttons - possibly moving them - sizing of visualization - I plan on approving and merging this PR after a brief re-test

dartajax (approved):
I approve this.

PR #9538, opened 2026-09-11
Create BigSaveCancelButtons component out of disparate `.bigadd`/`.bigcancel` instances

michaelchadwick:
Fixes ilios/ilios#7415

- [x] Move combined styles from various places into main `buttons.scss` foundational style file
- [x] Remove individual styles so they all inherit from foundational style
- [x] Create BigAddCancel component
- [x] Replace instances of lone styles with component
- [x] Create BigSaveCancelButtons page-object
- [x] Replace `[data-test-save|data-test-cancel]` instances in tests with page-object
- [x] Fix borked tests

Note: `learner-group/instructor-manager` ended up being an edge case. To convert it to this component required passing an `ember-concurrency` Task into a component, which proved infeasible to fix for now. Will need to be converted in a followup.


michaelchadwick:
Ideas for API improvement, summarized from @jrjohnson:
* Split Add/Cancel/Combined into three components (probably all in the same file)
* Possible [block/yield](https://guides.emberjs.com/release/components/block-content/) contextual component tricks
* `@addClasses` and `@disableSave` arguments having side effects not great for long term maintenance

packages/ilios-common/addon/components/big-save-cancel-buttons.gjs
```diff
@@ -0,0 +1,45 @@
...
+        {{yield @save @cancel @disableSave @disableCancel}}
+      {{else}}
+        <button
+          aria-label={{t "general.save"}}
+          type="button"
+          class="bigsave"
+          disabled={{@disableSave}}
+          {{on "click" @save}}
+          data-test-save
+          ...attributes
```
stopfstedt, line 29: hmm, if i read this right, then the splattributes are used to pass the `title` attribute to the save button, and we do this only for the `UserProfileCohortComponent`. not ideal.

there's also an a11y aspect to that - screen readers will read out the `aria-label` here, and ignore the `title` value. this is not a a11y-violation per-se - title and aria-label are allowed to have different values.
but i think that it would be better to make aria-label overridable by component input - if `title` is provided then it should be used as aria-label as well.

not a blocker for this PR, but could be a good follow-up item.

stopfstedt (approved):
nice cleanup, code looks good to me (with the understanding that there's some need for follow-up items). spot-checked the UI with click-testing, and eyeballed the differ output. the slight horizontal spacing changes between the buttons that the differ points are acceptable to me, they're not noticeable to me when looking at it in a browser. 

this looks good to me.

packages/ilios-common/addon-test-support/ilios-common/page-objects/components/big-save-cancel-buttons.js
```diff
@@ -0,0 +1,20 @@
+import { attribute, create, property } from 'ember-cli-page-object';
+
+const definition = {
+  scope: '[data-test-big-save-cancel-buttons]',
+  addButton: {
```
jrjohnson: Annoying, sorry, but this should now be `saveButton`

packages/ilios-common/addon-test-support/ilios-common/page-objects/components/detail-learning-materials.js
```diff
@@ -26,7 +26,7 @@ const definition = {
   pickNew: clickOnText(
     '.detail-learningmaterials-actions [data-test-choose-material-type] [data-test-item]',
   ),
-  save: clickable('.actions button.bigadd'),
+  save: clickable('.actions button.bigsave'),
```
jrjohnson: I think it's worth it in this PR, since these are changing anyway. To replace them with `[data-test-save]` wherever they're switching to `bigsave`

packages/ilios-common/addon/components/big-save-cancel-buttons.gjs
jrjohnson: With splatributes you can override what's in the component as well. To in the case where we pass title, also pass `aria-label` then the passed value will override the default from the component.

packages/ilios-common/addon/components/big-save-cancel-buttons.gjs
```diff
@@ -0,0 +1,45 @@
...
+        >
+          <FaIcon
+            @icon={{if (or @save.isRunning @disableSave) faSpinner faCheck}}
```
jrjohnson: This relies on `@save` being a task. I don't object to that. It's nice. But I think we need an assertion, maybe in the constructor. or we could add a

pseudocode:
```
get save() {
assert.save(this.args.save.isTask, "save must be a task");
return this.args.save()
``` 

to ensure it's true.

packages/ilios-common/addon/components/big-save-cancel-buttons.gjs
```diff
@@ -0,0 +1,45 @@
...
+            @spin={{if (or @save.isRunning @disableSave) true false}}
```
jrjohnson: I'm not a fan of `disableSave` causing the spin. I think maybe both disabling and spinning should be tied to an `isSaving` property instead. This isn't a blocker. Could easily be followup.

packages/ilios-common/addon/components/big-save-cancel-buttons.gjs
```diff
@@ -0,0 +1,45 @@
...
+import FaIcon from '@fortawesome/ember-fontawesome/components/fa-icon';
+import { faSpinner, faCheck, faArrowRotateLeft } from '@fortawesome/free-solid-svg-icons';
+
+export default class BigSaveCancelButtonsComponent extends Component {
+  @service intl;
+
+  <template>
+    <div data-test-big-save-cancel-buttons>
+      {{#if (has-block)}}
+        {{yield @save @cancel @disableSave @disableCancel}}
```
jrjohnson: I think there is probably one more API improvement possible for this component. We could yield BigSave and BigCancel components instead of the actions. We're only using this in the block form in a few places so this isn't a blocker. Could easily be followup.

packages/ilios-common/addon/components/big-save-cancel-buttons.gjs
```diff
@@ -0,0 +1,45 @@
+import Component from '@glimmer/component';
+import { service } from '@ember/service';
+import { on } from '@ember/modifier';
+import t from 'ember-intl/helpers/t';
+import { or } from 'ember-truth-helpers';
+import FaIcon from '@fortawesome/ember-fontawesome/components/fa-icon';
+import { faSpinner, faCheck, faArrowRotateLeft } from '@fortawesome/free-solid-svg-icons';
+
+export default class BigSaveCancelButtonsComponent extends Component {
+  @service intl;
```
jrjohnson: I don't think we're using this service (just the `t` helper). So this can go away and then the component is template only.

packages/test-app/tests/integration/components/big-save-cancel-buttons-test.gjs
```diff
@@ -0,0 +1,21 @@
...
+import { setupRenderingTest } from 'test-app/tests/helpers';
+import { render } from '@ember/test-helpers';
+import { component } from 'ilios-common/page-objects/components/big-save-cancel-buttons';
+import BigSaveCancelButtons from 'ilios-common/components/big-save-cancel-buttons';
+import noop from 'ilios-common/helpers/noop';
+
+module('Integration | Component | big-save-cancel-buttons', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
```
jrjohnson: Needs tests for clicking handlers, for the spin change when saving, and for the block form.

jrjohnson (changes requested):
Awesome. Just a couple of things to change and a few more to either fixup here or ticket for later.

packages/ilios-common/addon/components/big-save-cancel-buttons.gjs
michaelchadwick: I think a followup would be good!

packages/test-app/tests/integration/components/big-save-cancel-buttons-test.gjs
michaelchadwick: Added block vs non-block layout tests, as well as save/cancel handlers. The spin testing is eluding me at the moment, but it's a start.

jrjohnson (approved):
Ginormous. Possibly needs another refactor. But looking mergable and good.

PR #9539, opened 2026-09-13
Bring Back Nice Errors

jrjohnson:
This version is now a babel plugin and works with our build. Provides much better error messages when tests fail.

We had some old config cruft from when we had this before, it's no longer relevant and I removed it.


michaelchadwick (approved):
This is much better than before, and worth the change, for sure.

michaelchadwick:
Screenshot 2026-09-14 at 10 07 40 AM
Screenshot 2026-09-14 at 10 07 19 AM

Some tests (the above is a unit test that has the same output regardless of this PR) still don't show anything better than before, and I'm not sure why, but that could be settled later.

dartajax (approved):

PR #9540, opened 2026-09-14
Fix for Cohort Scroll Position

jrjohnson:
This is my proposal to fix ilios/ilios#7305 in two parts.

The first is a test for the problem, it's useful even if this isn't the final solution.

The second commit is a cache on the getter. This allows us to manage the state a bit better for the list of learners
in a cohort. Because we need to call `getAllDescendantUsers` from the learner group tracking breaks down a bit and we get a re-render for the list. Introducing a cache (that isn't tracked) makes this getter return the last successful value and we don't re-render until we have a full list again and then Ember can compute only the changes.


michaelchadwick:
This seems like a good solution, but since it's draft I'll just note my thoughts:
1. It looks like Ilios uses a local cache idea on the Course->Sessions route, so not a completely novel trick: https://github.com/ilios/frontend/blob/master/packages/ilios-common/addon/components/course/sessions.gjs#L32
2. There's a service called `PreserveScroll` that's also used in the same area, but it doesn't do the thing we want here (which is why I noticed but ignored it when I was trying to fix this) since it handles route transition scroll saving, not within-route scroll saving. However, it's only used in one place right now, so maybe it could be leveraged?

jrjohnson:
My feeling is that isn't a scroll position bug, it's an accidental re-rendering bug. It would also cause a loss of focus and maybe some other a11y issues, we're only seeing the scroll right now. 

Good find on the other cache - still probably cheating. If this looks mostly ok though I'll try and clean it up a bit so I can be more proud and less ashamed than I am with it as is.

stopfstedt (approved):
works as intended in learner group management where the list of unassigned learners in the cohort is long.
however, if the target list of group members is long, then the user still gets scrolled up.

partial fix, looks good enough to me.

dartajax (approved):
fine with me

PR #9541, opened 2026-09-14
add section about contributor accountability to our contributing guideline.

stopfstedt:


jrjohnson (approved):
Agreed. And we can link directly to this if we close PRs.

PR #9543, opened 2026-09-15
Editinplace yes/no buttons now auto-wrap if space is tight

michaelchadwick:
Fixes ilios/ilios#7390

The Screenshot 2026-09-15 at 10 49 52 AM buttons that appear when you edit something in-place now will auto-wrap if the space they're in is too narrow. This fixes the issue in question (wrapping in the event space is too narrow), but also always wraps them in places like Objective and Session descriptions since the space is always narrow.


stopfstedt (approved):
code reviewed, and visually reviewed. looks good to me.

dartajax (approved):

PR #9544, opened 2026-09-15
Course->Rollover scroll to component correction

michaelchadwick:
Fixes ilios/ilios#7434

As has been done in several other parts of the app, the `scrollIntoView` modifier needs a little delay before activating so that the component it's scrolling to is fully loaded. Otherwise, it scrolls to what's been loaded so far since the route changed, which is not the whole enchilada (only like, maybe 1/2->2/3 of it).


stopfstedt (approved):

dartajax (approved):
this is great

PR #9545, opened 2026-09-16
Silence ESA Deprecations

jrjohnson:
Ember Simple Auth is changing the way it works internally to work better for newer ember apps. The documentation at
https://github.com/mainmatter/ember-simple-auth/blob/master/guides/upgrade-to-v9.md doesn't feel fully baked to me and we're not that modern yet. So for now, I'm silencing these and we'll deal with them later.


michaelchadwick (approved):
SILENCIO

PR #9551, opened 2026-09-17
Improve Test Target

jrjohnson:
After removing the LTI the BackToCourses component will always be visible and this test will no longer work pointing to just the div.


stopfstedt (approved):
i like.

PR #9552, opened 2026-09-17
Update Stylelint

jrjohnson:
We have this pinned to the ember-cli version, but we don't pin the other packages. This caused the scss packages to get ahead of the stylelint version and cause peer resolution issues. I've removed the exception from dependabot and grouped all these together. This is the first update to get everything back in sync.


jrjohnson:
Something about this is letting loose the hounds of dependency hell. Will need to wait for the monorepo consolidation effort.

PR #9553, opened 2026-09-17
Remove Unused Helpers

jrjohnson:
We deprecated and stopped using these a long time ago, get rid of them.


stopfstedt (approved):
awesome.

PR #9554, opened 2026-09-19
Use Script Runners in CI

jrjohnson:
By parameterizing the parallel flag for ember exam we can use these scripts directly. This makes our runs, particularly visual diff, better because we will always run the right script even if things change between master and the commit under test.


zorgbort:
Baseline failures are expected, changed the name of the run command in the second run.

